PR #4, opened 2018-04-04
RDM-1982 Timeout configuration applied to RestTemplate object

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-1982

Timeout for callbacks were not applied. Added a fix for this and added a timeout test

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
@@ -119,6 +119,7 @@ public class CallbackService {
             requestFactory.setConnectionRequestTimeout(secondsToMilliseconds(timeout));
             requestFactory.setReadTimeout(secondsToMilliseconds(timeout));
             requestFactory.setConnectTimeout(secondsToMilliseconds(timeout));
+            restTemplate.setRequestFactory(requestFactory);
             return Optional.ofNullable(
                 restTemplate.exchange(url, HttpMethod.POST, requestEntity, clazz));
         } catch (RestClientException e) {
--- a/src/test/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackServiceTest.java
+++ b/src/test/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackServiceTest.java
@@ -105,7 +105,8 @@ public class CallbackServiceTest {
 
         long duration = (endTime - startTime);
         final CallbackResponse response = result.orElseThrow(() -> new AssertionError("Missing result"));
-        assertTrue(duration/1000000 > 2500);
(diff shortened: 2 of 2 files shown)


src/test/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackServiceTest.java
```diff
@@ -82,6 +82,33 @@ public void happyPathWithNoErrorsOrWarnings() throws Exception {
...
+        mockCallbackServer.stubFor(post(urlMatching("/test-callback.*"))
+            .willReturn(okJson(mapper.writeValueAsString(callbackResponse)).withStatus(200).withFixedDelay(1500)));
+
+        long startTime = System.nanoTime();
+        final Optional<CallbackResponse> result = callbackService.send(testUrl, null, caseEvent, caseDetails);
+        long endTime = System.nanoTime();
+
+        long duration = (endTime - startTime);
+        final CallbackResponse response = result.orElseThrow(() -> new AssertionError("Missing result"));
+        assertTrue(duration/1000000 > 2500);
```
mario-paniccia: can you use 1__000__000 ? helps readability

mario-paniccia: if we are testing retry, can we do a kind of verify with wiremock to make sure the endpoint is called twice? not sure whether wiremock supports that

tzarsmango: both comments appreciated and applied 👍

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -119,6 +119,7 @@ private Integer secondsToMilliseconds(final Integer seconds) {
             requestFactory.setConnectionRequestTimeout(secondsToMilliseconds(timeout));
             requestFactory.setReadTimeout(secondsToMilliseconds(timeout));
             requestFactory.setConnectTimeout(secondsToMilliseconds(timeout));
+            restTemplate.setRequestFactory(requestFactory);
             return Optional.ofNullable(
```
adr1ancho: I am now confused.  A while back, restTemplate was created inside this function.  Then it was injected.  Now we are changing the properties of this restTemplate.  Should we create a restTemplate as per call back??  Just a thought

tzarsmango: Your point may be right, but I think we should discuss it out of the scope of this bug fix. Perhaps with another task to validate the need for that.
Just for the sake of completing this task.

PR #8, opened 2018-04-06
RDM-1448: Enforce user access at DB level

vlaurin:

https://tools.hmcts.net/jira/browse/RDM-1448

### Change description ###

Enforce user access constraints as part of the SQL queries. This allows pagination on search to stay accurate when cases are filtered out by this constraint.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
@@ -28,7 +28,7 @@ public class CachedCaseDetailsRepository implements CaseDetailsRepository {
     private static final String META_AND_FIELD_DATA_HASH_FORMAT = "%s%s";
 
     private final CaseDetailsRepository caseDetailsRepository;
-    private final Map<Long, CaseDetails> idToCaseDetails = newHashMap();
+    private final Map<Long, Optional<CaseDetails>> idToCaseDetails = newHashMap();
     private final Map<String, CaseDetails> referenceToCaseDetails = newHashMap();
     private final Map<String, CaseDetails> findHashToCaseDetails = newHashMap();
     private final Map<String, List<CaseDetails>> metaAndFieldDataHashToCaseDetails = newHashMap();
@@ -46,15 +46,13 @@ public class CachedCaseDetailsRepository implements CaseDetailsRepository {
 
     @Override
     public Optional<CaseDetails> findById(String jurisdiction, Long id) {
-        final Function<Long, CaseDetails> findCase = (key) -> caseDetailsRepository.findById(jurisdiction, id)
-                                                                                   .orElse(null);
-        return Optional.ofNullable(idToCaseDetails.computeIfAbsent(id, findCase))
-                       .filter(caseDetails -> jurisdiction.equals(caseDetails.getJurisdiction()));
+        return idToCaseDetails.computeIfAbsent(id, (key) -> caseDetailsRepository.findById(jurisdiction, id));
     }
 
     @Override
     public CaseDetails findById(final Long id) {
-        return idToCaseDetails.computeIfAbsent(id, caseDetailsRepository::findById);
+        return idToCaseDetails.computeIfAbsent(id, (key) -> Optional.ofNullable(caseDetailsRepository.findById(id)))
+                              .orElse(null);
     }
 
     @Override
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
@@ -29,7 +29,7 @@ public class CachedCaseDetailsRepository implements CaseDetailsRepository {
 
     private final CaseDetailsRepository caseDetailsRepository;
     private final Map<Long, Optional<CaseDetails>> idToCaseDetails = newHashMap();
-    private final Map<String, CaseDetails> referenceToCaseDetails = newHashMap();
+    private final Map<String, Optional<CaseDetails>> referenceToCaseDetails = newHashMap();
     private final Map<String, CaseDetails> findHashToCaseDetails = newHashMap();
     private final Map<String, List<CaseDetails>> metaAndFieldDataHashToCaseDetails = newHashMap();
     private final Map<String, PaginatedSearchMetadata> hashToPaginatedSearchMetadata = newHashMap();
@@ -57,7 +57,10 @@ public class CachedCaseDetailsRepository implements CaseDetailsRepository {
 
     @Override
     public CaseDetails findByReference(final Long caseReference) {
-        return referenceToCaseDetails.computeIfAbsent(caseReference.toString(), (key) -> caseDetailsRepository.findByReference(caseReference));
+        final Function<String, Optional<CaseDetails>> findFunction = (key) -> Optional.ofNullable(
+            caseDetailsRepository.findByReference(caseReference));
+        return referenceToCaseDetails.computeIfAbsent(caseReference.toString(), findFunction)
+                                     .orElse(null);
     }
 
     @Override
@@ -67,17 +70,9 @@ public class CachedCaseDetailsRepository implements CaseDetailsRepository {
 
     @Override
     public Optional<CaseDetails> findByReference(String jurisdiction, String reference) {
-        if (referenceToCaseDetails.containsKey(reference)) {
-            return Optional.ofNullable(referenceToCaseDetails.get(reference))
-                           .filter(caseDetails -> jurisdiction.equals(caseDetails.getJurisdiction()));
-        } else {
-            final Optional<CaseDetails> optionalCaseDetails = caseDetailsRepository.findByReference(jurisdiction,
-                                                                                                    reference);
-
-            optionalCaseDetails.ifPresent(caseDetails -> referenceToCaseDetails.put(reference, caseDetails));
-
-            return optionalCaseDetails;
-        }
+        return referenceToCaseDetails.computeIfAbsent(reference,
+                                                      (key) -> caseDetailsRepository.findByReference(jurisdiction,
+                                                                                                     reference));
     }
 
     @Override
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
@@ -11,6 +11,7 @@ import javax.inject.Inject;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static com.google.common.collect.Maps.newHashMap;
@@ -28,7 +29,7 @@ public class CachedCaseDetailsRepository implements CaseDetailsRepository {
 
     private final CaseDetailsRepository caseDetailsRepository;
     private final Map<Long, CaseDetails> idToCaseDetails = newHashMap();
-    private final Map<Long, CaseDetails> referenceToCaseDetails = newHashMap();
+    private final Map<String, CaseDetails> referenceToCaseDetails = newHashMap();
     private final Map<String, CaseDetails> findHashToCaseDetails = newHashMap();
     private final Map<String, List<CaseDetails>> metaAndFieldDataHashToCaseDetails = newHashMap();
     private final Map<String, PaginatedSearchMetadata> hashToPaginatedSearchMetadata = newHashMap();
@@ -44,26 +45,49 @@ public class CachedCaseDetailsRepository implements CaseDetailsRepository {
     }
 
     @Override
+    public Optional<CaseDetails> findById(String jurisdiction, Long id) {
+        final Function<Long, CaseDetails> findCase = (key) -> caseDetailsRepository.findById(jurisdiction, id)
+                                                                                   .orElse(null);
+        return Optional.ofNullable(idToCaseDetails.computeIfAbsent(id, findCase))
+                       .filter(caseDetails -> jurisdiction.equals(caseDetails.getJurisdiction()));
+    }
+
+    @Override
     public CaseDetails findById(final Long id) {
         return idToCaseDetails.computeIfAbsent(id, caseDetailsRepository::findById);
     }
 
     @Override
     public CaseDetails findByReference(final Long caseReference) {
-        return referenceToCaseDetails.computeIfAbsent(caseReference, caseDetailsRepository::findByReference);
+        return referenceToCaseDetails.computeIfAbsent(caseReference.toString(), (key) -> caseDetailsRepository.findByReference(caseReference));
     }
 
     @Override
-    public Optional<CaseDetails> findByReference(String jurisdictionId, Long caseReference) {
-        if (referenceToCaseDetails.containsKey(caseReference)) {
-            return Optional.ofNullable(referenceToCaseDetails.get(caseReference))
-                           .filter(caseDetails -> jurisdictionId.equals(caseDetails.getJurisdiction()));
+    public Optional<CaseDetails> findByReference(String jurisdiction, Long reference) {
+        return findByReference(jurisdiction, reference.toString());
+    }
+
+    @Override
+    public Optional<CaseDetails> findByReference(String jurisdiction, String reference) {
+        if (referenceToCaseDetails.containsKey(reference)) {
+            return Optional.ofNullable(referenceToCaseDetails.get(reference))
+                           .filter(caseDetails -> jurisdiction.equals(caseDetails.getJurisdiction()));
         } else {
-            final Optional<CaseDetails> optionalCaseDetails = caseDetailsRepository.findByReference(jurisdictionId,
-                                                                                                    caseReference);
-            referenceToCaseDetails.put(caseReference, optionalCaseDetails.orElse(null));
+            final Optional<CaseDetails> optionalCaseDetails = caseDetailsRepository.findByReference(jurisdiction,
+                                                                                                    reference);
+            referenceToCaseDetails.put(reference, optionalCaseDetails.orElse(null));
             return optionalCaseDetails;
         }
+    }
+
+    @Override
+    public Optional<CaseDetails> lockByReference(String jurisdiction, Long reference) {
+        return lockByReference(jurisdiction, reference.toString());
+    }
+
+    @Override
+    public Optional<CaseDetails> lockByReference(String jurisdiction, String reference) {
+        return caseDetailsRepository.lockByReference(jurisdiction, reference);
     }
 
     @Override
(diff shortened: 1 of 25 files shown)


src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+        metadata.getCreatedDate().map(this::whereCreatedDate);
+        metadata.getLastModified().map(this::whereLastModified);
+        metadata.getSecurityClassification().map(this::whereSecurityClassification);
+
+        return this;
+    }
+
+    public CaseDetailsQueryBuilder orderByCreatedDate() {
+        orders.add(cb.asc(root.get("createdDate")));
+
```
adr1ancho: Sonar usually said use Constant rather than "createdDate"

src/test/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepositoryTest.java
```diff
@@ -119,17 +125,15 @@ public void findByReferenceShouldReturnCorrectSingleRecord() {
...
+
         MetaData metadata = new MetaData("TestAddressBookCase", "PROBATE");
         HashMap<String, String> searchParams = new HashMap<>();
         searchParams.put("PersonFirstName", "Janet");
         final List<CaseDetails> byMetaDataAndFieldData = caseDetailsRepository.findByMetaDataAndFieldData(metadata,
                                                                                                           searchParams);
         assertAll(
-            () -> assertThat(byMetaDataAndFieldData.size(), is(1)),
+            () -> assertThat(byMetaDataAndFieldData.size(), is(2)),
             () -> assertThat(byMetaDataAndFieldData.get(0).getId(), is(1L)),
```
adr1ancho: Just wonder whether you want also to assertThat byMetaDataAndFieldData.get(1) attributes

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
vlaurin: I think it did say that because it was used multiple times before. Now that this string is only used once, Sonar is not complaining. However, we should consider using Metadata API to reference properties, instead of strings or constant.

src/test/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepositoryTest.java
vlaurin: I think that the tests are already very fragile as they are, and that would make them even more fragile. We should review our approach for test cases, it feels wrong that I had to update ~50 test cases just because I haded a new cases to the set of test cases.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
```diff
@@ -43,29 +44,52 @@ public CaseDetails set(final CaseDetails caseDetails) {
         return caseDetailsRepository.set(caseDetails);
     }
 
+    @Override
+    public Optional<CaseDetails> findById(String jurisdiction, Long id) {
+        final Function<Long, CaseDetails> findCase = (key) -> caseDetailsRepository.findById(jurisdiction, id)
+                                                                                   .orElse(null);
+        return Optional.ofNullable(idToCaseDetails.computeIfAbsent(id, findCase))
+                       .filter(caseDetails -> jurisdiction.equals(caseDetails.getJurisdiction()));
```
mario-paniccia: if we are doing a case detail find by id and jurisdiction why are we performing a filter on jurisdiction here?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
```diff
@@ -43,29 +44,52 @@ public CaseDetails set(final CaseDetails caseDetails) {
...
+        if (referenceToCaseDetails.containsKey(reference)) {
+            return Optional.ofNullable(referenceToCaseDetails.get(reference))
+                           .filter(caseDetails -> jurisdiction.equals(caseDetails.getJurisdiction()));
         } else {
-            final Optional<CaseDetails> optionalCaseDetails = caseDetailsRepository.findByReference(jurisdictionId,
-                                                                                                    caseReference);
-            referenceToCaseDetails.put(caseReference, optionalCaseDetails.orElse(null));
+            final Optional<CaseDetails> optionalCaseDetails = caseDetailsRepository.findByReference(jurisdiction,
+                                                                                                    reference);
+            referenceToCaseDetails.put(reference, optionalCaseDetails.orElse(null));
```
mario-paniccia: can we replace with 
optionalCaseDetails.ifPresent(referenceToCaseDetails.put..

src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
```diff
@@ -43,29 +44,52 @@ public CaseDetails set(final CaseDetails caseDetails) {
...
 
     @Override
-    public Optional<CaseDetails> findByReference(String jurisdictionId, Long caseReference) {
-        if (referenceToCaseDetails.containsKey(caseReference)) {
-            return Optional.ofNullable(referenceToCaseDetails.get(caseReference))
-                           .filter(caseDetails -> jurisdictionId.equals(caseDetails.getJurisdiction()));
+    public Optional<CaseDetails> findByReference(String jurisdiction, String reference) {
+        if (referenceToCaseDetails.containsKey(reference)) {
+            return Optional.ofNullable(referenceToCaseDetails.get(reference))
+                           .filter(caseDetails -> jurisdiction.equals(caseDetails.getJurisdiction()));
```
mario-paniccia: because in case of cache miss we are doing a get by jurisdiction and id, is this check necessary?
Also, do we have to check this at all given that case reference is unique across jurisdictions?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -157,6 +174,27 @@ public PaginatedSearchMetadata getPaginatedSearchMetadata(MetaData metaData, Map
         return sr;
     }
 
+    // TODO This accepts null values for backward compatibility. Once deprecated methods are removed, parameters should
+    // be annotated with @NotNull
+    private Optional<CaseDetailsEntity> find(String jurisdiction, Long id, String reference) {
+        final CaseDetailsQueryBuilder qb = queryBuilderFactory.select(em);
+
+        if (null != jurisdiction) {
+            qb.whereJurisdiction(jurisdiction);
```
mario-paniccia: not sure whether would be better, but we could have whereJurisdiction method do the null check itself. 
This way we avoid repeated null checks in the clients of query builder

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -157,6 +174,27 @@ public PaginatedSearchMetadata getPaginatedSearchMetadata(MetaData metaData, Map
...
+        }
+
+        if (null != reference) {
```
mario-paniccia: same as above, if this is a common pattern we could have the query builder itself handle this whole if-else block to avoid duplications

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -157,6 +174,27 @@ public PaginatedSearchMetadata getPaginatedSearchMetadata(MetaData metaData, Map
...
+            qb.whereReference(reference);
+        } else {
+            qb.whereId(id);
+        }
+
+        final TypedQuery<CaseDetailsEntity> query = qb.build();
```
mario-paniccia: can we remove final please

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -157,6 +174,27 @@ public PaginatedSearchMetadata getPaginatedSearchMetadata(MetaData metaData, Map
...
+        return query.getResultList()
```
mario-paniccia: can we move this into a getSingleResult method?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CountCaseDetailsQueryBuilder.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.hmcts.ccd.data.casedetails.query;
```
mario-paniccia: wonder whether it's better to have a buildCountQuery, buildSelectQuery, createCountQuery and createSelectQuery instead of introducing a hierarchy here

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/UserAuthorisationSecurity.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private final UserAuthorisation userAuthorisation;
+
+    @Autowired
+    UserAuthorisationSecurity(UserAuthorisation userAuthorisation) {
+        this.userAuthorisation = userAuthorisation;
+    }
+
+    <T> CaseDetailsQueryBuilder<T> secure(CaseDetailsQueryBuilder<T> builder) {
+        if (AccessLevel.GRANTED.equals(userAuthorisation.getAccessLevel())) {
```
mario-paniccia: could we have the query builder itself handling this instead of introducing  a new component?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java
```diff
@@ -28,21 +29,22 @@
...
         this.criteraFactory = criteraFactory;
         this.entityManager = entityManager;
         this.applicationParam = applicationParam;
+        this.userAuthorisation = userAuthorisation;
     }
 
     public Query build(MetaData metadata, Map<String, String> params, boolean isCountQuery) {
         final List<Criterion> criteria = criteraFactory.build(metadata, params);
-        String queryString = String.format(isCountQuery ? MAIN_COUNT_QUERY : MAINQUERY, toClauses(criteria));
+        String queryString = String.format(isCountQuery ? MAIN_COUNT_QUERY : MAINQUERY, secure(toClauses(criteria)));
```
mario-paniccia: mind renaming MAINQUERY to MAIN_QUERY?

src/main/java/uk/gov/hmcts/ccd/infrastructure/user/UserAuthorisationFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private final CaseAccessService caseAccessService;
+
+    @Autowired
+    public UserAuthorisationFactory(CaseAccessService caseAccessService) {
+        this.caseAccessService = caseAccessService;
+    }
+
+    @Bean
+    @RequestScope
```
mario-paniccia: does UserAuthorisation need to be a @Bean? could be a simple object to me

src/main/java/uk/gov/hmcts/ccd/infrastructure/user/UserAuthorisationFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.web.context.annotation.RequestScope;
+import uk.gov.hmcts.ccd.domain.service.common.CaseAccessService;
+import uk.gov.hmcts.ccd.infrastructure.user.UserAuthorisation.AccessLevel;
+import uk.gov.hmcts.reform.auth.checker.spring.serviceanduser.ServiceAndUserDetails;
+
+@Configuration
+public class UserAuthorisationFactory {
```
mario-paniccia: can we rename this to make it more evident that this is a spring configuration class?

src/test/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpointIT.java
```diff
@@ -114,7 +114,7 @@ public void validSearch() throws Exception {
         // Check that we have the expected test data set size, this is to ensure
```
mario-paniccia: CaseDetailsQueryBuilder is rather complex, I would add unit tests for it. We can't rely only on Integration Tests

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpointIT.java
```diff
@@ -1215,7 +1215,7 @@ public void shouldReturn201WhenPostCreateCaseEventWithNoDataForCaseworker() thro
             mapper.readTree(mvcResult.getResponse().getContentAsString()).get("case_data").toString());
 
         final List<CaseDetails> caseDetailsList = template.query("SELECT * FROM case_data", this::mapCaseData);
```
mario-paniccia: shall we add an explicit integration test for the access level granted?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
vlaurin: The reason for the filter is because the same cache is used by both `findById(id)` and `findById(jurisdiction, id)`. Because of that, the cache could be populated by `findById(id)` which doesn't enforce jurisdiction but `findById(jurisdiction, id)` is expected to enforce jurisdiction check.

I agree that this seems like a trivial case and adds complexity; so I'm removing that filter.

vlaurin: Changed cache to `Map<Long, Optional<CaseDetails>>`

src/main/java/uk/gov/hmcts/ccd/data/casedetails/CachedCaseDetailsRepository.java
vlaurin: 👍

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
vlaurin: The only instances where `jurisdiction` may be `null` is where deprecated methods are used. By keeping that check close to the deprecated code, it will be easier to clean. Moreover, `null` jurisdiction is not the expected behaviour so I wouldn't want to facilitate it.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
vlaurin: This part of the code will go away once the deprecated methods are removed. This is not the expected behaviour or a pattern that should be encouraged.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
vlaurin: `getSingleResult()` would throw an exception on empty result we would then need to be caught to return an empty optional. I think `.getResultList().stream().findFirst()` gives a more fluent approach than exception handling.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
vlaurin: Removed `query` variable and used `qb.build()` instead.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+
+public abstract class CaseDetailsQueryBuilder<T> {
+
+    protected final EntityManager em;
+    protected final CriteriaBuilder cb;
+    protected final CriteriaQuery<T> query;
+    protected final Root<CaseDetailsEntity> root;
+    protected final ArrayList<Predicate> predicates;
```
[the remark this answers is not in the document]
vlaurin: 👍

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CountCaseDetailsQueryBuilder.java
vlaurin: That's what I tried initially, but I couldn't find a way to make it work without loosing the inner type of the query. Having a generic abstract class with subclasses to capture the type-specific code is the only way I found to preserve the `TypedQuery`. Also, it keeps the interface clean and keep the type specific functions concentrated in the factory.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/UserAuthorisationSecurity.java
vlaurin: I think that would be giving too much responsibility to the QueryBuilder.
Also, the QueryBuilder is quite hard to unit tests given it's using the Criteria API. Keeping the security logic separated gives the abstraction necessary to easily unit test the security logic.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java
vlaurin: 👍

src/main/java/uk/gov/hmcts/ccd/infrastructure/user/UserAuthorisationFactory.java
vlaurin: 👍

src/test/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpointIT.java
vlaurin: I would as well, but I cannot think of a decent way of doing it because of the Criteria API. One way would be to create a rather complex hierarchy of mocks; the other way would be to use a real persistence context, build query, unwrap them into the JPA provider specific type and then output them as SQL and test the string. I'm not sold on either of those :s

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
mario-paniccia: sorry didn't mean using getSingleResult of jpa. I meant create  a method on the builder and move this whole code in there

vlaurin: oh sorry, I misunderstood. Can do, yes, good idea.

vlaurin: Done.

PR #19, opened 2018-04-16
Sandbox configuration change for probate team

adr1ancho:

Sandbox configuration change for probate team

```
[x] Yes
[ ] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -11,7 +11,7 @@ provider "vault" {
 locals {
   app_full_name = "${var.product}-${var.component}"
   env_ase_url = "${var.env}.service.${data.terraform_remote_state.core_apps_compute.ase_name[0]}.internal"
-  default_default_print_url = "https://return-case-doc-ccd.nonprod.platform.hmcts.net/jurisdictions/:jid/case-types/:ctid/cases/:cid"
+  default_default_print_url = "https://ccd-case-print-service-${env_ase_url}"
   default_print_url = "${var.default_print_url != "" ? var.default_print_url : local.default_default_print_url}"
 }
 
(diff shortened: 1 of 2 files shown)


infrastructure/variables.tf
```diff
@@ -57,9 +57,11 @@ variable "authorised-services" {
 }
 
 variable "document_management_valid_domain" {
-  default = "https://api-gateway.test.dm.reform.hmcts.net"
```
vlaurin: Changing this is not required as long as it's using tactical. This should be reverted and left to use tactical `test`.

infrastructure/main.tf
```diff
@@ -11,6 +11,10 @@ provider "vault" {
 locals {
   app_full_name = "${var.product}-${var.component}"
   env_ase_url = "${var.env}.service.${data.terraform_remote_state.core_apps_compute.ase_name[0]}.internal"
+  default_document_management_valid_domain = "https://api-gateway.preprod.dm.reform.hmcts.net"
+  default_default_print_url = "https://return-case-doc-ccd.nonprod.platform.hmcts.net/jurisdictions/:jid/case-types/:ctid/cases/:cid"
```
vlaurin: This needs to dynamically build the `.internal` URL

infrastructure/main.tf
```diff
@@ -11,6 +11,8 @@ provider "vault" {
 locals {
   app_full_name = "${var.product}-${var.component}"
   env_ase_url = "${var.env}.service.${data.terraform_remote_state.core_apps_compute.ase_name[0]}.internal"
+  default_default_print_url = "https://ccd-case-print-service-${env_ase_url}"
```
vlaurin: I think this needs to be `http` instead of `https`. Just discovered that `SSL` is failing when using `.internal` names because it's a self-signed.

PR #24, opened 2018-04-17
RDM-2013 Do not override case data's default classification on each e…

andrewfolga:
…vent

https://tools.hmcts.net/jira/browse/RDM-2013

### Change description ###
Actual: Every time event is triggered all of case's default data classification gets its fields reset from case type definition even though it has already been set during previously triggered events.

Expected: We should only set default security classification once so that already existing fields do not get their classification reset on each new event and only new fields do.

Notes:
- performance improvement in that we do not need to loop over all case fields once we found the searched for one,
- fixes some tests that were treating Document type classification as complex object whereas it should be simple type

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseDataService.java
```diff
@@ -49,13 +51,19 @@ private void deduceDefaultClassifications(final JsonNode dataNode,
                     final String caseFieldType = caseField.getFieldType().getType();
                     if (caseFieldType.equalsIgnoreCase(COMPLEX_TYPE)) {
                         found = true;
-                        deduceClassificationForComplexType(dataNode, fieldIdPrefix, fieldName, caseField);
+                        deduceClassificationForComplexType(dataNode,
+                                                           getExistingDataClassificationNodeOrEmpty(existingDataClassificationNode, fieldName), fieldIdPrefix, fieldName, caseField);
+                        break;
```
andrewfolga: Performance improvement in that we do not need to loop over all case fields once we found the searched for one.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/FieldType.java
```diff
@@ -85,4 +86,9 @@ public FieldType getCollectionFieldType() {
     public void setCollectionFieldType(FieldType collectionFieldType) {
         this.collectionFieldType = collectionFieldType;
     }
+
+    @Override
+    public String toString() {
```
mario-paniccia: do we invoke this toString often? if yes, reflection can hurt performance

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseDataService.java
```diff
@@ -31,13 +31,15 @@
     private static final String CLASSIFICATION = "classification";
 
     public Map<String, JsonNode> getDefaultSecurityClassifications(final CaseType caseType,
-                                                                   final Map<String, JsonNode> caseData) {
-        final JsonNode dataClassification = cloneAndConvertDataMap(caseData);
-        deduceDefaultClassifications(dataClassification, caseType.getCaseFields(), EMPTY_STRING);
-        return MAPPER.convertValue(dataClassification, STRING_JSON_MAP);
+                                                                   final Map<String, JsonNode> caseData,
+                                                                   final Map<String, JsonNode> currentDataClassification) {
+        final JsonNode newDataClassification = cloneAndConvertDataMap(caseData);
```
mario-paniccia: perhaps better name it clonedDataClassification?

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpointIT.java
```diff
@@ -1328,7 +1322,6 @@ public void shouldReturn201WhenPostCreateCaseEventWithNoDataForCitizen() throws
         assertEquals(savedCaseDetails.getData(), caseAuditEvent.getData());
         assertEquals(SUMMARY, caseAuditEvent.getSummary());
         assertEquals(DESCRIPTION, caseAuditEvent.getDescription());
-        assertEquals(mapper.convertValue(EXPECTED_CLASSIFICATION, new TypeReference<HashMap<String, JsonNode>>() {}), caseAuditEvent.getDataClassification());
```
andrewfolga: Readd it.

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpointIT.java
```diff
@@ -1245,7 +1239,6 @@ public void shouldReturn201WhenPostCreateCaseEventWithNoDataForCaseworker() thro
...
```
andrewfolga: Readd it.

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpointIT.java
```diff
@@ -3261,7 +3251,6 @@ private void shouldReturn201WhenPostCreateCaseEventWithValidData(String userRole
...
```
andrewfolga: Readd it.

PR #36, opened 2018-04-26
RDM-1831 Hide and show condition for a field on a case view tab and case view tab field.

andrewfolga:

https://tools.hmcts.net/jira/browse/RDM-1831
https://tools.hmcts.net/jira/browse/RDM-1591
https://tools.hmcts.net/jira/browse/RDM-1964

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewField.java
```diff
@@ -121,4 +122,10 @@ public Boolean getShowSummaryChangeOption() {
     public void setShowSummaryChangeOption(final Boolean showSummaryChangeOption) {
         this.showSummaryChangeOption = showSummaryChangeOption;
     }
+
+    @Override
+    public String toString() {
+        return ReflectionToStringBuilder.toString(this);
```
mario-paniccia: beware of using reflection. Can slow down performance. Perhaps an IDE generated toString() better?

src/test/resources/mappings/display-tab-structure.json
```diff
@@ -15,7 +15,9 @@
         "id" : "NameTab",
         "label" : "Name",
         "order" : 1,
+        "show_condition": "NameTabShowCondition",
         "tab_fields" : [{
+          "show_condition": "PersonFirstNameFieldShowCondition",
```
mario-paniccia: perhaps we can add a valid example show condition here?

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewField.java
andrewfolga: Yeah maybe we should agree on a standard here. I can see many different usages across the code base including:
- StringBuilder,
- ReflectionToStringBuilder,
- commons-lang ToStringBuilder,
- commons-lang3 ToStringBuilder

I would lean towards the last option. What are your thoughts?

andrewfolga: ok changed to commons-lang3 ToStringBuilder

PR #45, opened 2018-05-03
Rdm 2069_off_demo

mario-paniccia:
Fixed bug with Label field type now being displayed on case view tab sections

https://tools.hmcts.net/jira/browse/RDM-2069

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
@@ -180,7 +180,7 @@ public class CaseDetails implements Cloneable {
 
     public boolean existsInData(CaseTypeTabField caseTypeTabField) {
         // Fields of type "Label" are a special case and should be allowed through regardless
-        return caseTypeTabField.getCaseField().getFieldType().getId().equals(LABEL_FIELD_TYPE) ||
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
```diff
@@ -180,7 +180,7 @@ public void setIncompleteCallbackResponse() {
 
     public boolean existsInData(CaseTypeTabField caseTypeTabField) {
         // Fields of type "Label" are a special case and should be allowed through regardless
-        return caseTypeTabField.getCaseField().getId().equals(LABEL_FIELD_TYPE) ||
+        return caseTypeTabField.getCaseField().getFieldType().getType().equals(LABEL_FIELD_TYPE) ||
```
mario-paniccia: I'm 90% sure it's getFieldType().getType() here instead of 
getFieldType().getId()
What do you guys think?

andrewfolga: Confirming

mario-paniccia: thanks @andrewfolga

PR #63, opened 2018-05-16
ROC-3537 Make sort direction configurable

timja-kainos:
https://tools.hmcts.net/jira/browse/ROC-3537

### Change description ###
Added metadata parameter sortDirection, which accepts 'asc' or 'desc' (ignoring case)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/MetaData.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/MetaData.java
@@ -15,9 +15,9 @@ public class MetaData {
     public static final String LAST_MODIFIED_PARAM = "last_modified_date";
     public static final String SECURITY_CLASSIFICATION_PARAM = "security_classification";
     public static final String PAGE_PARAM = "page";
-    public static final String SORT_DIRECTION = "sortDirection";
+    public static final String SORT_DIRECTION_PARAM = "sortDirection";
     private static final List<String> ALL_METADATA = newArrayList(STATE_PARAM, CASE_REFERENCE_PARAM,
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/MetaData.java
```diff
@@ -15,8 +15,9 @@
     public static final String LAST_MODIFIED_PARAM = "last_modified_date";
     public static final String SECURITY_CLASSIFICATION_PARAM = "security_classification";
     public static final String PAGE_PARAM = "page";
+    public static final String SORT_DIRECTION = "sortDirection";
```
mario-paniccia: can you suffix with PARAM please, to be consistent with the other params

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
```diff
@@ -461,6 +472,12 @@ private void validateMetadataSearchParameters(Map<String, String> queryParameter
                 throw new BadRequestException(String.format("unknown security classification '%s'", sc));
             }
         });
+
+        param(queryParameters, MetaData.SORT_DIRECTION).ifPresent(sd -> {
+            if (Stream.of("ASC", "DESC").noneMatch(direction -> direction.equalsIgnoreCase(sd))) {
```
mario-paniccia: perhaps we could introduce a SortDirection enum as we have one for the SecurityClassification param?

timja-kainos: I looked at it but it didn't seem to add much value because of the limited places it would be used and because Metadata is passing strings around so you would have to convert to it again.

SecurityClassification is used in lots more places than this would be

PR #66, opened 2018-05-21
RDM-2230: Support event ID max length 70 chars

vlaurin:

https://tools.hmcts.net/jira/browse/RDM-2230

### Change description ###

Definition store dictates that event IDs have a max length of 70 chars. However, data store only support max length of 40 chars. This is updating the column to be `varchar(70)`.

```
[ ] Yes
[x] No
```


tzarsmango:
Do we need a script for migrating the existing data, or is it handled by liquibase?

vlaurin:
@tzarsmango No migration needed, it's only extending the size of varchar.

PR #82, opened 2018-06-01
Infr: switch last envs to use UK db

dwaynebailey:
### Change description ###

Moves the remaining envs to use UK db: demo, preview and saat

Not sure which represent sprod.  And likely we want to do these one at a time.


infrastructure/preview.tfvars
```diff
@@ -2,3 +2,4 @@ vault_section = "preprod"
 idam_api_url = "https://preprod-idamapi.reform.hmcts.net:3511"
```
mario-paniccia: @dwaynebailey what about sprod ?

dwaynebailey:
This was superseded by other PRs

PR #111, opened 2018-06-28
RDM-2361 New endpoint for saving draft to draft store.

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-2361
https://tools.hmcts.net/jira/browse/RDM-2362
https://tools.hmcts.net/jira/browse/RDM-2363

### Change description ###
New endpoint(s) for saving/updating draft to draft store. 
Also includes terraform changes to create the secret key for encoding/decoding drafts in the draft store.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/endpoint/std/DraftsEndpoint.java
```diff
@@ -0,0 +1,49 @@
...
+        this.saveDraftOperation = saveDraftOperation;
+    }
+
+    @RequestMapping(value = "/caseworkers/{uid}/jurisdictions/{jid}/case-types/{ctid}/event-trigger/{etid}/drafts", method = RequestMethod.POST)
+    @ResponseStatus(HttpStatus.CREATED)
+    @ApiOperation(
+        value = "Save draft as a caseworker."
+    )
+    @ApiResponses(value = {
+        @ApiResponse(code = 201, message = "Case created"),
```
andrewfolga: Message should have Draft not Case.

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftRepository.java
```diff
@@ -0,0 +1,68 @@
...
+            final RestTemplate restTemplate,
+            final SecurityUtils securityUtils,
+            final ApplicationParams applicationParams) {
+        this.restTemplate = restTemplate;
+        this.securityUtils = securityUtils;
+        this.applicationParams = applicationParams;
+    }
+
+    @Override
+    public Draft set(final CreateCaseDataContentDraft draft) {
```
vlaurin: The name of the method is too generic I believe. This implementation is exclusively to create new drafts, not updating existing as `set` may suggest.

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftRepository.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private Draft assembleDraft(CreateCaseDataContentDraft requestDraft, HttpHeaders responseHeaders) {
+        Draft responseDraft = new Draft();
+        responseDraft.setId(getId(responseHeaders));
+        return responseDraft;
+    }
+
+    private Long getId(HttpHeaders responseHeaders) {
+        return Long.valueOf(responseHeaders.getLocation().getRawPath().split("/")[2]);
```
vlaurin: This does feel a bit fragile as a way to retrieve the ID though I appreciate it may be the only way. Are we sure there's no other way?

src/main/java/uk/gov/hmcts/ccd/endpoint/std/DraftsEndpoint.java
```diff
@@ -0,0 +1,49 @@
...
+@Api(value = "/", description = "Drafts API")
+public class DraftsEndpoint {
+    private final SaveDraftOperation saveDraftOperation;
+
+    @Autowired
+    public DraftsEndpoint(@Qualifier("default") final SaveDraftOperation saveDraftOperation) {
+        this.saveDraftOperation = saveDraftOperation;
+    }
+
+    @RequestMapping(value = "/caseworkers/{uid}/jurisdictions/{jid}/case-types/{ctid}/event-trigger/{etid}/drafts", method = RequestMethod.POST)
```
vlaurin: Given the Drafts would be listed per case type, not per event, I wonder if it makes sense to create them against an event-specific resource.

```POST /caseworkers/{uid}/jurisdictions/{jid}/case-types/{ctid}/drafts``` would feel more appropriate, however it would require passing the event ID as part of the body.

File: src/main/java/uk/gov/hmcts/ccd/domain/service/createdraft/DefaultSaveDraftOperation.java [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public Draft saveDraft(final String uid,
+                           final String jurisdictionId,
+                           final String caseTypeId,
+                           final String eventTriggerId,
+                           final CaseDataContent caseDataContent) {
+        CaseDataContentDraft caseDataContentDraft = buildCaseDataContentDraft(uid, jurisdictionId, caseTypeId, eventTriggerId, caseDataContent);
+        CreateCaseDataContentDraft createCaseDataContentDraft = new CreateCaseDataContentDraft(caseDataContentDraft, CASE_DATA_CONTENT, applicationParams.getDraftMaxStaleDays());
+        return draftRepository.set(createCaseDataContentDraft);
```
vlaurin: The 3 lines above are a bit complicated to read. I believe this is due to the number of parameters passed in 2 function calls and the name of the variables assigned being quite similar.

Maybe a builder pattern could be used to create an instance of `CreateCaseDataContentDraft` without exposing the underlying `CreateCaseDataContentDraft`?

`CaseDataContentDraft` and `CreateCaseDataContentDraft` could be renamed to be more distinguished. At the moment, having `buildCaseDataContentDraft` followed by `new CreateCaseDataContentDraft` almost feel like the same thing is done twice.

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftRepository.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+
+@Named
+@Qualifier(DefaultDraftRepository.QUALIFIER)
+@Singleton
+public class DefaultDraftRepository implements DraftRepository {
```
vlaurin: Could we rename those to `DraftGateway` and `DefaultDraftGateway` instead to avoid confusion with DAL?
For reference, see:
- https://martinfowler.com/eaaCatalog/gateway.html
- https://martinfowler.com/eaaCatalog/repository.html

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftRepository.java
andrewfolga: This is indeed changed and called `save` in the follow up PR for 2362:
https://github.com/hmcts/ccd-data-store-api/pull/115

src/main/java/uk/gov/hmcts/ccd/endpoint/std/DraftsEndpoint.java
andrewfolga: The endpoint was built this way so we store the draft with all metadata required. This metadata is needed to build CaseView and populate the event trigger dropdown on case view.

infrastructure/main.tf
```diff
@@ -52,6 +52,17 @@ data "vault_generic_secret" "gateway_oauth2_client_secret" {
   path = "secret/${var.vault_section}/ccidam/idam-api/oauth2/client-secrets/ccd-gateway"
 }
 
+resource "random_string" "draft_encryption_key" {
+  length  = 16
+  special = true
```
vlaurin: Could we double-check with someone from RPE if special characters are allowed in the encryption key?

PR #122, opened 2018-07-09
RDM-882: Cleanup collection classification structure

vlaurin:

https://tools.hmcts.net/jira/browse/RDM-882

### Change description ###

Couple of tests were assuming that classification structure for simple collection was:
```json
{
    "Addresses": [
        {
            "id": "xxx-xxx-xxx",
            "value": {
                "Address1": "PRIVATE"
            }
        }
    ]
}
```

This seems to be a confusion between structure for simple and complex collection. Actual structure for simple collection is:

```json
{
    "Addresses": [
        {
            "id": "xxx-xxx-xxx",
            "classification": "PRIVATE"
        }
    ]
}
```
As supported by function `CaseDataService#deduceClassificationForCollectionType()` which does:

```java
final ObjectNode simpleCollectionItemNode = (ObjectNode) field;
// Add `classification` property
deduceClassificationForSimpleType(simpleCollectionItemNode, existingDataClassificationNode, CLASSIFICATION, caseField);
// Remove `value` property
simpleCollectionItemNode.remove(VALUE);
```

This assumption has been corrected and warnings have been added in the event of incorrect classification structure being used.
The warning would display as:

```
2018-07-09T14:52:08.391+0100 WARN  [main] uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationService:159: Invalid security classification structure for collection item {"Address1":"PUBLIC"}
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationService.java
@@ -156,9 +156,8 @@ public class SecurityClassificationService {
                                        relevantDataClassificationValue,
                                        userClassification);
                 } else {
-                    filterSimpleField(userClassification,
-                                      dataCollectionIterator,
-                                      relevantDataClassificationValue.get(collectionElementValue.textValue()));
+                    LOG.warn("Invalid security classification structure for collection item: {}", relevantDataClassificationValue.toString());
+                    dataCollectionIterator.remove();
                 }
             } else {
                 // For collection of simple field type, the classification is stored as `classification`, not `value`
--- a/src/test/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationServiceTest.java
+++ b/src/test/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationServiceTest.java
@@ -896,6 +896,124 @@ public class SecurityClassificationServiceTest {
         }
 
         @Test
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationService.java
```diff
@@ -156,12 +156,20 @@ private void filterCollection(SecurityClassification userClassification, Iterato
                                        relevantDataClassificationValue,
                                        userClassification);
                 } else {
+                    LOG.warn("Invalid security classification structure for collection item: {}", relevantDataClassificationValue.toString());
```
andrewfolga: can we have a collection of collection here?

andrewfolga: Do we need test for this one?

vlaurin: Collection of collections is not possible. It would need to be a collection of complex object containing a collection. So there would be a Complex in the middle. But that's a good point, we need to make sure we test with that scenario.

vlaurin: I'll add a test for it, just to be safe 👍

vlaurin:
@andrewfolga Test added

PR #126, opened 2018-07-12
RDM-626 - Display metadata in search results

gbenadikar:
https://tools.hmcts.net/jira/projects/RDM/issues/RDM-626

### Change description ###
Code changes to support metadata fields in search results. Also did a little refactoring from array to list for the search results and made case fields data type more generic rather than a map of String and JsonNode, it is now a map of String and Object. Please review

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/search/SearchResultViewItem.java
```diff
@@ -1,31 +1,34 @@
...
     public String getCaseId() {
         return caseId;
     }
 
-    public Map<String, JsonNode> getCaseFields() {
+    public Map<String, Object> getCaseFields() {
         return caseFields;
     }
+
+    public <T> SearchResultViewItem addCaseFields(Map<String, T> caseFields) {
```
mario-paniccia: another alternative which if we don't want to introduce generics would be:
```
    public SearchResultViewItem addCaseFields(Map<String, ? extends Object> caseFields) {

```
not sure which one is better though :)

gbenadikar: Its synonymous in this context, but wild card looks more readable. Changed now.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/MetaData.java
```diff
@@ -19,8 +21,8 @@
     private static final List<String> ALL_METADATA = newArrayList(STATE_PARAM, CASE_REFERENCE_PARAM,
```
mario-paniccia: can you update this with the new params?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/MetaData.java
```diff
@@ -9,6 +9,8 @@
 
 public class MetaData {
 
+    public static final String JURISDICTION_PARAM = "jurisdiction";
```
mario-paniccia: there's a bit of an issue here we have to address. 
This list of *_PARAM is used to keep track of metadata that is passed to the api as a query parameter. They are added to the ALL_METADATA map which is used to validate the request. 
Jurisdiction and case type were not there because they are not passed as query parameters. 

I think we should rename ALL_METADATA to METADATA_QUERY_PARAMETERS
and rename all the constants from *_PARAM to *_METADATA

src/main/java/uk/gov/hmcts/ccd/domain/model/search/SearchResultViewItem.java
```diff
@@ -1,31 +1,34 @@
...
 
+import java.util.HashMap;
 import java.util.Map;
 
 public class SearchResultViewItem {
     @JsonProperty("case_id")
     private String caseId;
     @JsonProperty("case_fields")
-    private Map<String, JsonNode> caseFields;
+    private final Map<String, Object> caseFields = new HashMap<>();
```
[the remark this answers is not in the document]
gbenadikar: Originally I was thinking on the same lines, metadata fields and case data are different. Then post arch discussion, we came to conclusion that metadata and case data both are actually case fields from domain perspective, as all these fields belong to a case. Hence metadata and case data fields are now stored in case_fields table of definition store DB. To make distinction between case data and metadata fields, there is boolean flag 'metadata' for UI. So thought both metadata and casedata can be stored in caseFields

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/MetaData.java
gbenadikar: sounds good, will rename it.

src/main/java/uk/gov/hmcts/ccd/domain/model/search/SearchResultViewItem.java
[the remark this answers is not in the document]
mario-paniccia: I see. Ok, if we need to think of metadata as being fields from now on, then the implementation is correct 👍 
If in the future we are asked to differentiate the behaviour between metadata and real fields we can always change this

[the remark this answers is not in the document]
gbenadikar: 👍

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/MetaData.java
gbenadikar: Done

PR #128, opened 2018-07-13
Rdm 2394 View case draft

andrewfolga:
All subtasks covered:
https://tools.hmcts.net/jira/browse/RDM-2465

### Change description ###
Build CaseView from draft - add DefaultGetDraftViewOperation to build case view from draft.
Introduce draftStoreTemplate that reduces the standard 30s timeout to 100ms for all draft retrieval operations - post, put, get, get all.
Add app insights support for draft store gateway.
Update SearchResult to add resultError containing any problems related to slow or unresponsive draft store.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
```diff
@@ -77,7 +77,7 @@ public Long save(final CreateCaseDraft draft) {
             return getDraftId(responseHeaders);
         } catch (Exception e) {
             LOG.warn("Error while saving draft=" + draft, e);
-            throw new ServiceException("Problem saving draft because of " + e.getMessage());
+            throw new ServiceException(DRAFT_STORE_DOWN_ERR_MESSAGE);
```
mario-paniccia: Even if we are logging the Exception e as a WARN, 
I think we should add the whole Exception e as a a caused by when we create the ServiceException. This way when later we log the error with ERROR level we get the 'caused by' and we know what the real issue is. 
ServiceException at the moment doesn't support passing a caused by, but we can add that.

mario-paniccia: we could do this in a tech debt story, as I see we are doing this everywhere

PR #150, opened 2018-08-03
RDM-2223 As a user creating a case I want to be able to save the progress of my form

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-2223:
- https://tools.hmcts.net/jira/browse/RDM-2361
- https://tools.hmcts.net/jira/browse/RDM-2362
- https://tools.hmcts.net/jira/browse/RDM-2363

https://tools.hmcts.net/jira/browse/RDM-2394:
- https://tools.hmcts.net/jira/browse/RDM-2430
- https://tools.hmcts.net/jira/browse/RDM-2431
- https://tools.hmcts.net/jira/browse/RDM-2440
- https://tools.hmcts.net/jira/browse/RDM-2465

### Change description ###

New endpoint(s) for saving/updating/fetching a single draft and fetching all drafts to/from draft store.
Also includes terraform changes to create the secret key for encoding/decoding drafts in the draft store.
Change SearchQueryOperation to build and add drafts to SearchResult.
Change GetCaseViewOperation to build case view from draft.

```
[ ] Yes
[x] No
```


README.md
```diff
@@ -42,6 +42,10 @@ The following environment variables are required:
 | HTTP_CLIENT_MAX_CLIENT_PER_ROUTE | 20 | Used for Pooling connection manager; for further information, see https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html |
 | HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY | 0 | Used for Pooling connection manager; for further information, see https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html |
 | HTTP_CLIENT_CONNECTION_TIMEOUT | 30000 | 30 seconds, Used for Pooling connection manager; for further information, see https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html |
+| HTTP_CLIENT_CONNECTION_DRAFTS_TIMEOUT | 200 | 200 milliseconds, Used for Pooling connection manager for draft store; for further information, see https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html |
+| DRAFT_STORE_URL | - | Base URL for Draft Store API service. `http://localhost:8800` for the dockerised local instance. |
+| DRAFT_ENCRYPTION_KEY | - | Draft encryption key. The encryption key used by draft store to encrypt documents with. |
+| DRAFT_MAX_STALE_DAYS | - | Number of days after which the saved draft will be deleted if unmodified. |
```
mario-paniccia: can we call this DRAFT_TTL?

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
```diff
@@ -0,0 +1,179 @@
...
+            HttpHeaders responseHeaders = restTemplate.exchange(applicationParams.draftBaseURL(),
+                                                                HttpMethod.POST,
+                                                                requestEntity,
+                                                                HttpEntity.class).getHeaders();
+            final Duration duration = Duration.between(start, Instant.now());
+            appInsights.trackDependency(DRAFT_STORE, "Create", duration.toMillis(), true);
+            return getDraftId(responseHeaders);
+        } catch (Exception e) {
+            LOG.warn("Error while saving draft", e);
+            throw new ServiceException(DRAFT_STORE_DOWN_ERR_MESSAGE, e);
```
mario-paniccia: this call could fail for many reasons, not only draft store down. Shall we make this error message more generic?

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
```diff
@@ -0,0 +1,179 @@
...
+            HttpHeaders headers = securityUtils.authorizationHeaders();
+            headers.add(DRAFT_ENCRYPTION_KEY_HEADER, applicationParams.getDraftEncryptionKey());
+            final HttpEntity requestEntity = new HttpEntity(headers);
+            final Instant start = Instant.now();
+            Draft draft = restTemplate.exchange(applicationParams.draftURL(draftId), HttpMethod.GET, requestEntity, Draft.class).getBody();
+            final Duration duration = Duration.between(start, Instant.now());
+            appInsights.trackDependency(DRAFT_STORE, "Get", duration.toMillis(), true);
+            return assembleDraft(draft);
+        } catch (HttpClientErrorException e) {
+            LOG.warn("Error while getting draftId=" + draftId, e);
```
mario-paniccia: can we use placeholders rather than string concatenation?

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
```diff
@@ -0,0 +1,179 @@
...
+            DraftList getDrafts = restTemplate.exchange(getUriWithQueryParams(), HttpMethod.GET, requestEntity, DraftList.class).getBody();
+            final Duration duration = Duration.between(start, Instant.now());
+            appInsights.trackDependency(DRAFT_STORE, "GetAll", duration.toMillis(), true);
+            return getDrafts.getData()
+                .stream()
+                .map(this::assembleDraft)
+                .collect(Collectors.toList());
+        } catch (Exception e) {
+            LOG.warn("Error while getting drafts", e);
+            throw new DraftAccessException(DRAFT_STORE_DOWN_ERR_MESSAGE, e);
```
mario-paniccia: same as above, this could be a bit misleading.

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+        } catch (Exception e) {
+            LOG.warn("Error while updating draftId={}", draftId, e);
+            throw new ServiceException(DRAFT_STORE_DOWN_ERR_MESSAGE, e);
+        }
+        return null;
+    }
+
+    @Override
+    public DraftResponse get(final String draftId) {
```
mario-paniccia: I think get methods are better return Optional in the case the entity is not found. We are using Optional in our new CaseDetailsRepository API

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseEventTrigger.java
```diff
@@ -23,6 +23,8 @@
     private Boolean showEventNotes;
     @JsonProperty("end_button_label")
     private String endButtonLabel;
+    @JsonProperty("can_save_draft")
```
mario-paniccia: why not more simply `draftEnabled`?

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseEventTrigger.java
```diff
@@ -103,4 +105,12 @@ public String getEndButtonLabel() {
     public void setEndButtonLabel(String endButtonLabel) {
         this.endButtonLabel = endButtonLabel;
     }
+
+    public Boolean getCanSaveDraft() {
```
mario-paniccia: the java convention for bool methods is to start with `is`. we should rename this to `isCanSaveDraft`?
which doesn't sound that nice. I think we should rename this to `isDraftEnabled`

src/main/java/uk/gov/hmcts/ccd/domain/model/draft/DraftResponse.java
```diff
@@ -0,0 +1,60 @@
...
+@ToString
+public class DraftResponse {
+
+    private String id;
+
+    private CaseDraft document;
+
+    private String type;
+
+    private LocalDateTime created;
```
mario-paniccia: in the Draft class we are using ZonedDateTime for these props. Here a LocalDataTime. Is this ok?

src/main/java/uk/gov/hmcts/ccd/domain/model/draft/UpdateCaseDraftBuilder.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.hmcts.ccd.domain.model.draft;
+
+public class UpdateCaseDraftBuilder {
+    private final UpdateCaseDraftRequest updateCaseDraftRequest;
```
mario-paniccia: since we are writing many builders in this PR, perhaps is worth mentioning that the way we are implementing Builders here is not the recommended/standard way of doing it. 
The recommended way would be to have all the fields of the object we want to build as fields of the builder. And then to instantiate the object in the build method. This allows the creation of immutable objects.
Also, builders are usually worth the effort when the object to build is a complex object. Not sure i would provide a builder for a class with 3/4 fields only

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
```diff
@@ -0,0 +1,102 @@
...
+        caseView.setCaseType(CaseViewType.createFrom(caseType));
+        caseView.setTabs(getTabs(caseDetails, caseDetails.getData(), caseTabCollection));
+
+        caseView.setTriggers(new CaseViewTrigger[] {resumeTrigger, DELETE_TRIGGER});
+        caseView.setEvents(new CaseViewEvent[]{});
+
+        return caseView;
+    }
+
+    private CaseDetails buildCaseDetailsFromDraft(DraftResponse draftResponse) {
```
mario-paniccia: I think this deserves it's own mapper class

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public CaseView execute(String jurisdictionId, String caseTypeId, String draftId) {
+
+        final CaseType caseType = getCaseType(jurisdictionId, caseTypeId);
+        final DraftResponse draftResponse = draftGateway.get(draftId);
+
+        final CaseDetails caseDetails = buildCaseDetailsFromDraft(draftResponse);
+        final CaseViewTrigger resumeTrigger = buildResumeTriggerFromDraft(draftResponse);
```
mario-paniccia: could we explicitly model a resume trigger with a subclass of CaseViewTrigger called ResumeTrigger? It could provide a constructor which accepts id and description. 
This way we could also remove the builder

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
```diff
@@ -0,0 +1,102 @@
...
+    private CaseViewTrigger buildResumeTriggerFromDraft(DraftResponse draftResponse) {
+        return aCaseViewTrigger()
+            .withId(draftResponse.getDocument().getEventTriggerId())
+            .withName(RESUME)
+            .withDescription(draftResponse.getDocument().getCaseDataContent().getEvent().getDescription())
+            .withOrder(1)
+            .build();
+    }
+
+    private String getCaseTypeIdFromDraft(DraftResponse draftResponse) {
```
mario-paniccia: this method could be moved to DraftResponse itself

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
```diff
@@ -0,0 +1,102 @@
...
+        return draftResponse.getDocument().getCaseTypeId();
+    }
+
+    private CaseView merge(CaseDetails caseDetails, CaseViewTrigger resumeTrigger, CaseType caseType, CaseTabCollection caseTabCollection) {
+        CaseView caseView = new CaseView();
```
mario-paniccia: this logic it's similar to what we have in DefaultGetCaseViewOperation. 
It could be possible to introduce  a CaseViewBuilder and use that one in both DefaultGetCaseViewOperation and here?

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
```diff
@@ -0,0 +1,102 @@
...
+        .withName("Delete")
+        .withDescription("Delete draft")
+        .withOrder(2)
+        .build();
+    private static final String RESUME = "Resume";
+
+    private final DraftGateway draftGateway;
+
+    @Autowired
+    public DefaultGetDraftViewOperation(@Qualifier(CreatorGetCaseOperation.QUALIFIER) final GetCaseOperation getCaseOperation,
```
mario-paniccia: wouldn't DefaultGetCaseViewFromDraftOperation be better?

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
andrewfolga: This is as expected. Per design a generic message is supposed to be returned regardless of the reason.

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
andrewfolga: Done.

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
andrewfolga: This is as expected. Per design a generic message is supposed to be returned regardless of the reason.

README.md
andrewfolga: Let's keep it same as draft store. Also it has the DAYS in it so more descriptive.

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseEventTrigger.java
andrewfolga: This is consistent with excel file hence the naming.

src/main/java/uk/gov/hmcts/ccd/data/draft/DefaultDraftGateway.java
andrewfolga: It is consistent with general pattern which is handled by error handler. Both possible exceptions are converted into a generic HttpError and sent back to UI.

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseEventTrigger.java
andrewfolga: I think this is consistent with all other Boolean type variables in this class. Also not sure jackson accepts is... vs get... methods.

src/main/java/uk/gov/hmcts/ccd/domain/model/draft/CreateCaseDraftRequest.java
```diff
@@ -0,0 +1,49 @@
+package uk.gov.hmcts.ccd.domain.model.draft;
```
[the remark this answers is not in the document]
andrewfolga: It is consistent with CallbackRequest. If we need to move it let's have a tech debt ticket.

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
andrewfolga: done

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
andrewfolga: done

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
```diff
@@ -24,32 +27,49 @@ public MergeDataToSearchResultOperation(final UIDefinitionRepository uiDefinitio
...
+        return new SearchResultView(viewColumns, viewItems, resultError);
+    }
+
+    private SearchResultViewItem buildSearchResultViewItem(CaseDetails caseData) {
+        return new SearchResultViewItem(hasCaseReference(caseData) ? getCaseReference(caseData) : getDraftReference(caseData),
+                                        caseData.getCaseDataAndMetadata());
+    }
+
+    private String getCaseReference(CaseDetails caseData) {
+        return caseData.getReference().toString();
```
mario-paniccia: can we move this into CaseDetails?

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
```diff
@@ -24,32 +27,49 @@ public MergeDataToSearchResultOperation(final UIDefinitionRepository uiDefinitio
...
+    }
+
+    private String getDraftReference(CaseDetails caseData) {
+        return String.format(DRAFT_ID, caseData.getId());
+    }
+
+    private boolean hasCaseReference(CaseDetails caseData) {
+        return caseData.getReference() != null;
```
mario-paniccia: The way to do this in Java 8 is to have the methods return Optional. We definitely should change getReference() to return an optional now. This is very important. If we don't do this, we risk many subtle bugs and null pointer exceptions in the future whenever we'll retrieve the reference

src/main/java/uk/gov/hmcts/ccd/domain/model/draft/UpdateCaseDraftBuilder.java
andrewfolga: done

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
```diff
@@ -24,32 +27,49 @@ public MergeDataToSearchResultOperation(final UIDefinitionRepository uiDefinitio
...
         final List<SearchResultViewItem> viewItems = caseDetails.stream()
-            .map(caseData -> new SearchResultViewItem(caseData.getReference().toString(), caseData.getCaseDataAndMetadata()))
+            .map(caseData -> buildSearchResultViewItem(caseData))
             .collect(Collectors.toList());
-        return new SearchResultView(viewColumns, viewItems);
+        return new SearchResultView(viewColumns, viewItems, resultError);
+    }
+
+    private SearchResultViewItem buildSearchResultViewItem(CaseDetails caseData) {
+        return new SearchResultViewItem(hasCaseReference(caseData) ? getCaseReference(caseData) : getDraftReference(caseData),
```
mario-paniccia: to make the code more self explanatory, instead of checking hasCaseReference, we could introduce a method on CaseDetails `isDraft` which does the check. 
And here: 
```caseData.isDraft() ? caseData.getCaseReference().get() : caseData.getDraftReference()

File: src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java [hunk 19]
```diff
@@ -24,32 +27,49 @@ public MergeDataToSearchResultOperation(final UIDefinitionRepository uiDefinitio
...
+                                        caseData.getCaseDataAndMetadata());
+    }
+
+    private String getCaseReference(CaseDetails caseData) {
+        return caseData.getReference().toString();
+    }
+
+    private String getDraftReference(CaseDetails caseData) {
+        return String.format(DRAFT_ID, caseData.getId());
```
mario-paniccia: can we move this into CaseDetails?

src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
```diff
@@ -30,15 +30,15 @@ public void grantAccess(final String jurisdictionId, final String caseReference,
                                                                                       Long.valueOf(caseReference));
 
         final CaseDetails caseDetails = maybeCase.orElseThrow(() -> new CaseNotFoundException(caseReference));
-        caseUserRepository.grantAccess(caseDetails.getId(), userId);
+        caseUserRepository.grantAccess(Long.valueOf(caseDetails.getId()), userId);
```
mario-paniccia: Now that we changed caseId to String, I think we should raise a tech debt story convert all our methods that still expect caseId to be a Long. Otherwise we'll have to add these conversions everywhere

mario-paniccia: https://tools.hmcts.net/jira/browse/RDM-2695

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
andrewfolga: That is a bit of a stretch for few lines.

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
andrewfolga: I would avoid extracting this as a new class as there is no difference in content. I think the method name `buildResumeTriggerFromDraft` as well as the variable name `resumeTrigger` should be enough to understand it here.

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
mario-paniccia: in fact we have a duplicated block of code in `DefaultGetDraftsOperation` that is doing the same mapping. We could introduce a mapper class and use it here and `DefaultGetDraftsOperation`

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultGetDraftViewOperation.java
andrewfolga: Let's keep with the no builders solution here:) We could add a tech debt story if needs be.

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
```diff
@@ -20,37 +25,54 @@
...
         if (!caseType.isPresent()) {
-            return new SearchResultView(Collections.emptyList(), Collections.emptyList());
+            return new SearchResultView(Collections.emptyList(), Collections.emptyList(), NO_ERROR);
         }
-        final List<CaseDetails> caseData = searchOperation.execute(metadata, queryParameters);
-        return mergeDataToSearchResultOperation.execute(caseType.get(), caseData, view);
+
+        final List<CaseDetails> cases = searchOperation.execute(metadata, queryParameters);
+
+        String draftResultError = NO_ERROR;
```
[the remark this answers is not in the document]
andrewfolga: No because we have multiple output values namely: `draftResultError` and `draftsAndCases`

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
andrewfolga: done

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
andrewfolga: But then do you want to change Long reference to Optional<Long>? How would jackson ser/deser that. Also should we have it as a separate tech debt story? Btw, I do not think we should change it.

src/main/java/uk/gov/hmcts/ccd/domain/service/getdraft/DefaultGetDraftsOperation.java
```diff
@@ -0,0 +1,63 @@
...
+        CaseDraft document = draftResponse.getDocument();
+        return document.getCaseTypeId().equalsIgnoreCase(metadata.getCaseTypeId())
+            && document.getJurisdictionId().equalsIgnoreCase(metadata.getJurisdiction());
+    }
+
+    private List<CaseDetails> buildCaseDataFromDrafts(List<DraftResponse> drafts) {
+        return drafts.stream()
+            .map(d -> {
+                CaseDraft document = d.getDocument();
+                return aCaseDetails()
```
mario-paniccia: this is similar to what we have in DefaultGetDraftViewOperation. Can we introduce a mapper to remove this duplication?

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
andrewfolga: See above.

src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
andrewfolga: good point

src/main/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpoint.java
```diff
@@ -198,6 +189,23 @@ public CaseView findCase(@PathVariable("jid") final String jurisdictionId,
...
     }
 
+    @Transactional
+    @RequestMapping(value = "/caseworkers/{uid}/jurisdictions/{jid}/case-types/{ctid}/drafts/{did}",
+        method = RequestMethod.GET)
+    @ApiOperation(value = "Fetch a draft for display")
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "A displayable draft")
+    })
+    public CaseView findDraft(@PathVariable("jid") final String jurisdictionId,
```
mario-paniccia: shouldn't this be on the `DraftsEndpoint`?

src/main/resources/application.properties
```diff
@@ -5,7 +5,7 @@ server.compression.enabled=true
 server.compression.mime-types=application/json,application/xml,text/html,text/xml,text/plain
 
 spring.jpa.show-sql=false
-spring.datasource.url=jdbc:postgresql://${DATA_STORE_DB_HOST:localhost}:${DATA_STORE_DB_PORT:5452}/${DATA_STORE_DB_NAME:ccd_data}?stringtype=unspecified&ssl=${DATA_STORE_DB_USE_SSL:true}
+spring.datasource.url=jdbc:postgresql://${DATA_STORE_DB_HOST:localhost}:${DATA_STORE_DB_PORT:5000}/${DATA_STORE_DB_NAME:ccd_data}?stringtype=unspecified&ssl=${DATA_STORE_DB_USE_SSL:true}
```
mario-paniccia: how come are we changing this?

README.md
mario-paniccia: if you want to have days indication we could call it DRAFT_TTL_DAYS. I'm insisting because these type of properties are universally called TTL. We should really align to that

src/main/java/uk/gov/hmcts/ccd/domain/model/draft/CreateCaseDraftRequest.java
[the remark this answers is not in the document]
mario-paniccia: ok guys. But if the answer for continuing doing things is the wrong way is because is consistent then there will never be improvements.

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
mario-paniccia: if we don't change it, and in the future we call getReference() forgetting to do the null check, how can we prevent null pointer exceptions?

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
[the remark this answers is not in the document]
mario-paniccia: you could do 
`retrieveDrafts(draftAndCases, draftResultError)`

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseEventTrigger.java
mario-paniccia: Jackson can handle bool methods starting with is

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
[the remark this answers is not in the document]
andrewfolga: What I meant earlier is we have multiple return values so would have to build a grouping object here which is bloating it too much.

src/main/java/uk/gov/hmcts/ccd/domain/service/getdraft/DefaultGetDraftsOperation.java
andrewfolga: done

src/main/java/uk/gov/hmcts/ccd/domain/service/upsertdraft/DefaultUpsertDraftOperation.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.inject.Inject;
+
+import static uk.gov.hmcts.ccd.domain.model.draft.CaseDraftBuilder.aCaseDraft;
+import static uk.gov.hmcts.ccd.domain.model.draft.CreateCaseDraftBuilder.aCreateCaseDraft;
+import static uk.gov.hmcts.ccd.domain.model.draft.DraftResponseBuilder.aDraftResponse;
+import static uk.gov.hmcts.ccd.domain.model.draft.UpdateCaseDraftBuilder.anUpdateCaseDraft;
+
+@Service
+@Qualifier("default")
+public class DefaultUpsertDraftOperation implements UpsertDraftOperation {
```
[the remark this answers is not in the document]
andrewfolga: Yeah so we could extract CreateDraftOperation and UpdateDraftOperation here but I was inclined into having as few of these Operation classes as possible. I am against explosion of Operations that have just one method in them and would prefer to actually have DefaultDraftOperations class that would group: create, update, get, getAll and delete operations. What do you think?

src/main/resources/application.properties
andrewfolga: good spot

PR #175, opened 2018-09-05
RDM-2776 - search results pagination bug fix

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-2776

### Change description ###
Code changes to include authorised case states clause when retrieving case search results and count. Originally the filtering was applied post retrieval of search results. Since states can have CRUD authority, some of the cases were filtered after fetching the count and hence the discrepancy in search result pagination counts.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilderFactory.java
```diff
@@ -3,24 +3,39 @@
...
+import uk.gov.hmcts.ccd.data.casedetails.search.MetaData;
 
 import javax.persistence.EntityManager;
+import java.util.List;
 
 @Component
 public class CaseDetailsQueryBuilderFactory {
 
     private final UserAuthorisationSecurity userAuthorisationSecurity;
+    private final List<CaseDetailsAuthorisationSecurity> caseDetailsAuthorisationSecurities;
```
mario-paniccia: UserAuthorisationSecurity and CaseDetailsAuthorisationSecurity at the moment have slightly different apis (one returns void and accepts a metadata, the other accepts no metadata and returns the builder), but we could unify them and have both extend a common interface, e.g. `AuthorisationSecurity`. Or we could have UserAuthorisationSecurity be a CaseDetailsAuthorisationSecurity.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -207,13 +207,13 @@ private Query getQueryByParameters(MetaData metadata, final Map<String, String>
     }
 
     private Query getCountQueryByMetaData(MetaData metadata) {
-        return queryBuilderFactory.count(em)
+        return queryBuilderFactory.count(em, metadata)
                                   .whereMetadata(metadata)
```
mario-paniccia: now that the factory gets a metadata, we could have the factory itself calling whereMetadata(metadata) on the builder, rather than doing it explicitly here

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilderFactory.java
```diff
@@ -3,24 +3,39 @@
...
     @Autowired
-    public CaseDetailsQueryBuilderFactory(UserAuthorisationSecurity userAuthorisationSecurity) {
+    public CaseDetailsQueryBuilderFactory(UserAuthorisationSecurity userAuthorisationSecurity,
+                                          List<CaseDetailsAuthorisationSecurity> caseDetailsAuthorisationSecurities) {
         this.userAuthorisationSecurity = userAuthorisationSecurity;
+        this.caseDetailsAuthorisationSecurities = caseDetailsAuthorisationSecurities;
+    }
+
+    public CaseDetailsQueryBuilder<CaseDetailsEntity> select(EntityManager em, MetaData metaData) {
+        return secure(select(em), metaData);
```
mario-paniccia: select and count methods are implemented in a slightly different way, but do that same thing. 
The both apply all security checks to the builder. 
It would make things simpler if both were implemented using the same approach

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
```diff
@@ -81,6 +82,13 @@ public CaseDetailsQueryBuilder whereState(String state) {
         return this;
     }
 
+    public CaseDetailsQueryBuilder whereStates(List<String> states) {
+        if (CollectionUtils.isNotEmpty(states)) {
```
mario-paniccia: can we use
```if(! states.isEmpty())```
?

File: src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java [hunk 5]
```diff
@@ -60,14 +68,24 @@ public Query build(MetaData metadata, Map<String, String> params, boolean isCoun
...
             );
         }
-        return clauses;
+
+        // restrict cases to the case states the user has access to
+        List<CaseState> caseStates = authorisedCaseDefinitionDataService.getUserAuthorisedCaseStates(metadata.getJurisdiction(),
+                                                                                                     metadata.getCaseTypeId(),
+                                                                                                     CAN_READ);
+        if (CollectionUtils.isNotEmpty(caseStates)) {
+            newClauses += String.format(" AND state IN ('%s')", String.join("','", caseStates.stream().map(CaseState::getId).collect(Collectors.toList())));
```
mario-paniccia: shall we also add a method to AuthorisedCaseDefinitionDataService to retrieve only the state ids rather than having to 
```caseStates.stream().map(CaseState::getId).collect(Collectors.toList()))```
We are doing this in 2 places already, so maybe it's worth considering

File: src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java [hunk 6]
```diff
@@ -60,14 +68,24 @@ public Query build(MetaData metadata, Map<String, String> params, boolean isCoun
...
-            clauses += String.format(
+            newClauses += String.format(
                 " AND id IN (SELECT cu.case_data_id FROM case_users AS cu WHERE user_id = '%s')",
                 userAuthorisation.getUserId()
             );
         }
-        return clauses;
+
+        // restrict cases to the case states the user has access to
+        List<CaseState> caseStates = authorisedCaseDefinitionDataService.getUserAuthorisedCaseStates(metadata.getJurisdiction(),
```
mario-paniccia: can we break down `private String secure(String clauses)` body  into further private methods? the code you added can be extracted into a `addInStatesCondition` method for example

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java
gbenadikar: good point. Done that

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
gbenadikar: Any reason for not using apache commons CollectionUtils ?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java
gbenadikar: Done

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilderFactory.java
gbenadikar: @mario-paniccia , I originally made UserAuthorisationSecurity to implement CaseDetailsAuthorisationSecurity but then realised there is a method in DefaultCaseDetailsRepository that calls CaseDetailsQueryBuilderFactory.secure(em) without metadata. Its used for individual case clause. If I unify, I will have to pass metadata as null for this call and in the implementation for CaseStateAuthorisationSecurity and I will have to explicitly check if metadata is not null before adding the clause (which doesn't look great). 

It will be good to unify it, but couldn't find a common ground.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java
```diff
@@ -23,32 +28,35 @@
...
+        this.authorisedCaseDefinitionDataService = authorisedCaseDefinitionDataService;
     }
 
     public Query build(MetaData metadata, Map<String, String> params, boolean isCountQuery) {
-        final List<Criterion> criteria = criteraFactory.build(metadata, params);
+        final List<Criterion> criteria = criterionFactory.build(metadata, params);
         String queryString = String.format(isCountQuery ? MAIN_COUNT_QUERY : MAIN_QUERY,
-            secure(toClauses(criteria)),
-            metadata.getSortDirection().orElse(SORT_ASCENDING).toUpperCase()
+                                           secure(toClauses(criteria), metadata),
```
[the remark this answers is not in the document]
gbenadikar: will check on the extra tab

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
gbenadikar: Actually the factory should just return the builder and not do any building within. Do I have leverage to refactor this?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilderFactory.java
gbenadikar: reorganised code, count and select are consistent now

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
gbenadikar: reorganised code now

gbenadikar:
@mario-paniccia addressed review comments. Please re-review

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchQueryFactoryOperation.java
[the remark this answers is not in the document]
gbenadikar: seems to be fine

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
mario-paniccia: just because this can be done using no library method and is even more readable(for me of course)  :)
leave it if you want no big deal

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilderFactory.java
```diff
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.ccd.data.casedetails.query;
```
mario-paniccia: the advantage of the Factory before was that it had injected the UserAuthorisationSecurity. 
if now we expose a secure method on the builder itself, which accepts the list of AuthorisationSecurity, I don't see what's the benefit anymore of the factory.
Shall we remove it? 
Up to you

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseStateAuthorisationSecurity.java
```diff
@@ -0,0 +1,32 @@
...
+    private final AuthorisedCaseDefinitionDataService authorisedCaseDefinitionDataService;
+
+    @Autowired
+    public CaseStateAuthorisationSecurity(AuthorisedCaseDefinitionDataService authorisedCaseDefinitionDataService) {
+        this.authorisedCaseDefinitionDataService = authorisedCaseDefinitionDataService;
+    }
+
+    @Override
+    public <T> void secure(CaseDetailsQueryBuilder<T> builder, MetaData metadata) {
+        if (metadata != null) {
```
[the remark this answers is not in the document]
gbenadikar: that's the disadvantage of unifying UserAuthorisationSecurity with CaseStateAuthorisationSecurity. Some calls do not pass metadata for e.g. for an individual case it passes case reference. So had to do this.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
gbenadikar: just that it checks for null and empty collection in one call, I tend to use it :-)

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilderFactory.java
gbenadikar: yes removed now

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilderFactory.java
```diff
@@ -1,26 +0,0 @@
...
-
-    private final UserAuthorisationSecurity userAuthorisationSecurity;
-
-    @Autowired
-    public CaseDetailsQueryBuilderFactory(UserAuthorisationSecurity userAuthorisationSecurity) {
-        this.userAuthorisationSecurity = userAuthorisationSecurity;
-    }
-
-    public CaseDetailsQueryBuilder<CaseDetailsEntity> select(EntityManager em) {
-        return userAuthorisationSecurity.secure(new SelectCaseDetailsQueryBuilder(em));
```
vlaurin: @gbenadikar The whole point of this class, with the package-scoped `SelectCaseDetailsQueryBuilder` was that the `secure` step could not programmatically be bypassed outside of this package. By removing that, we now don't have a guaranty that security has been applied. Worse, it's now the responsibility of each repository function to enable the security which makes it optional at best. I don't think this is a change in the right direction, but happy to be proved wrong.

gbenadikar: @vlaurin thats alright, I have reverted this change now. Please review

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -207,16 +207,16 @@ private Query getQueryByParameters(MetaData metadata, final Map<String, String>
     }
 
     private Query getCountQueryByMetaData(MetaData metadata) {
-        return queryBuilderFactory.count(em)
-                                  .whereMetadata(metadata)
-                                  .build();
+        return queryBuilderFactory.count(em, metadata)
+            .whereMetadata(metadata)
+            .build();
```
vlaurin: The 2 lines above should be indented and aligned with `.count`

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -207,16 +207,16 @@ private Query getQueryByParameters(MetaData metadata, final Map<String, String>
...
 
     private Query getQueryByMetaData(MetaData metadata) {
-        return queryBuilderFactory.select(em)
-                                  .whereMetadata(metadata)
-                                  .orderByCreatedDate(metadata.getSortDirection().orElse("asc"))
-                                  .build();
+        return queryBuilderFactory.select(em, metadata)
+            .whereMetadata(metadata)
+            .orderByCreatedDate(metadata.getSortDirection().orElse("asc"))
+            .build();
```
vlaurin: The 2 lines above should be indented and aligned with `.select`

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
gbenadikar: indented

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
gbenadikar: indented

PR #180, opened 2018-09-10
Rdm 1626

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-1626

Populated existing return value of user Profile -> Jurisdictions to avoid a second call on building the Case List Filters

```
[ ] Yes
[X] No
```


src/test/java/uk/gov/hmcts/ccd/data/casedetails/JurisdictionMapperTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    Jurisdiction jurisdiction = newJurisdiction()
+        .withJurisdictionId("jid")
+        .withName("name")
+        .withDescription("description")
+        .withCaseType(caseType1)
+        .withCaseType(caseType2)
+        .build();
+
+    @BeforeEach
```
andrewfolga: Do we need it here?

PR #213, opened 2018-09-27
[PREVIEW] RDM-2811 - Add security to search API (Elasticsearch)

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-2811

### Change description ###
Security filters added
1. Case type security for ACL and security classification as validation step before querying ES
2. User authorised case states as ES query filter
3. When a user is solicitor/citizen/letter-holder, granted case ids as ES query filter
4. Filtering case fields based on ACLs and security classification as postfilter, after retrieval of results.

#### **Includes changes from AIP-2392, initial search API endpoint** *

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/user/CachedUserRepository.java
```diff
@@ -46,7 +47,13 @@ public UserDefault getUserDefaultSettings(String userId) {
...
     }
 
+    @Override
     public Set<SecurityClassification> getUserClassifications(String jurisdictionId) {
         return jurisdictionToUserClassifications.computeIfAbsent(jurisdictionId, userRepository::getUserClassifications);
     }
+
+    @Override
+    public SecurityClassification getHighestUserClassification() {
+        return userHighestSecurityClassification.computeIfAbsent("userHighestSecurityClassification", s -> userRepository.getHighestUserClassification());
```
mario-paniccia: do we really need to cache this? can't we just forward directly to the user repo?

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -107,12 +104,29 @@ public UserDefault getUserDefaultSettings(final String userId) {
...
-            if (message != null)
+            if (message != null) {
                 throw new BadRequestException(message);
+            }
             throw new ServiceException("Problem getting user default settings for " + userId);
         }
     }
 
+    @Override
+    public SecurityClassification getHighestUserClassification() {
```
mario-paniccia: minor, safe to ignore. 
I would have in repo methods only logic to deal with db access. A method like this, which adds further elaboration, would be better in the UserService. But I've seen we are doing this in other methods in this class already

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AuthorisedCaseDataFilter.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.hmcts.ccd.domain.service.common;
```
mario-paniccia: can we move all the filters and post processors to a `security` or similar subpackage?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
```diff
@@ -0,0 +1,51 @@
...
+        AuthorisedCaseDefinitionDataService authorisedCaseDefinitionDataService,
+        AuthorisedCaseDataFilter authorisedCaseDataFilter) {
+
+        this.caseDetailsSearchOperation = caseDetailsSearchOperation;
+        this.authorisedCaseDefinitionDataService = authorisedCaseDefinitionDataService;
+        this.authorisedCaseDataFilter = authorisedCaseDataFilter;
+    }
+
+    @Override
+    public CaseDetailsSearchResult execute(String caseTypeId, String query) {
```
mario-paniccia: my mistake. You mind renaming this parameter from query to jsonSearchRequest or searchRequest?  we pass through the whole search request not only the query part
Can you rename also in the interface please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
```diff
@@ -0,0 +1,51 @@
...
+        return authorisedCaseDefinitionDataService
+            .getAuthorisedCaseType(caseTypeId, CAN_READ)
+            .map(caseType -> search(caseType, query))
+            .orElseGet(() -> new CaseDetailsSearchResult(emptyList(), 0L));
```
mario-paniccia: instead of creating an empty result every time, can we define it once inside the CaseDetailsSearchResult and then always use that? i.e. 
.orElseGet(() -> CaseDetailsSearchResult.EMPTY)
(hope it's possible to jsonIgnore it inside CaseDetailsSearchResult. Otherwise let's define it somewhere else)

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private CaseDetailsSearchResult search(CaseType caseType, String query) {
+        CaseDetailsSearchResult result = caseDetailsSearchOperation.execute(caseType.getId(), query);
+        // filter case fields based on user authority
+        if (isNotEmpty(result.getCases())) {
```
mario-paniccia: this if check can be removed because the foreach won't iterate on an empty list

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParser.java
```diff
@@ -0,0 +1,51 @@
...
+
+    ElasticsearchQueryParser(ObjectMapperService objectMapperService, String query) {
+        this.objectMapperService = objectMapperService;
+        this.searchQueryObjectNode = this.objectMapperService.convertStringToObject(query, ObjectNode.class);
+    }
+
+    String extractQueryClause() {
+        return ofNullable(searchQueryObjectNode.get(QUERY_NAME))
+            .map(JsonNode::toString)
+            .orElseThrow(() -> new ValidationException("invalid search query"));
```
mario-paniccia: this is redundant. The search request has already been validated on the controller, so it must have a query entry

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParser.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private final ObjectMapperService objectMapperService;
+    private final ObjectNode searchQueryObjectNode;
+
+    ElasticsearchQueryParser(ObjectMapperService objectMapperService, String query) {
+        this.objectMapperService = objectMapperService;
+        this.searchQueryObjectNode = this.objectMapperService.convertStringToObject(query, ObjectNode.class);
+    }
+
+    String extractQueryClause() {
```
mario-paniccia: If you look at the controller `getQuery` method you'll see that we already parse the search request. But then we pass it down to the service as string. I wonder whether now it would be better to pass down the parsed query (Map) . 
If we do that then probably we no longer need this class so we could simplify things a bit. 
Not 100% sure whether that would complicate other areas thought, I leave it to you to decide.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/filter/CaseFilterFactory.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.hmcts.ccd.domain.service.search.filter;
+
+import java.util.Optional;
+
+public interface CaseFilterFactory<T> {
```
mario-paniccia: does this need to be a generic? I can't see this returning other types than a `Optional<QueryBuilder>`

src/main/java/uk/gov/hmcts/ccd/domain/service/search/filter/CaseSearchFilter.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.ccd.domain.service.search.filter;
+
+public enum CaseSearchFilter {
```
mario-paniccia: could we remove this class and use instead
CaseDetailsEntity.STATE_FIELD_COL etc...

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchUserCaseAccessFilterFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import org.elasticsearch.index.query.QueryBuilder;
+import org.elasticsearch.index.query.QueryBuilders;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.ccd.domain.service.common.CaseAccessService;
+import uk.gov.hmcts.ccd.domain.service.search.filter.CaseFilterFactory;
+import uk.gov.hmcts.ccd.domain.service.search.filter.UserAccessFilter;
+
+@Component
+public class ElasticsearchUserCaseAccessFilterFactory extends UserAccessFilter implements CaseFilterFactory<QueryBuilder> {
```
mario-paniccia: I can't see the benefit of having a UserAccessFilter superclass. I think we could merge it into ElasticsearchUserCaseAccessFilterFactory.

mario-paniccia: what about changing to
```public class UserAccessQueryFilter implements ElasticsearchQueryFilter {

File: src/main/java/uk/gov/hmcts/ccd/domain/service/search/filter/CaseStateFilter.java [hunk 12]
```diff
@@ -0,0 +1,19 @@
+package uk.gov.hmcts.ccd.domain.service.search.filter;
+
+import java.util.List;
+
+import static uk.gov.hmcts.ccd.domain.service.common.AccessControlService.CAN_READ;
+
+import uk.gov.hmcts.ccd.domain.service.common.AuthorisedCaseDefinitionDataService;
+
+public class CaseStateFilter {
```
mario-paniccia: same as the other comment, I'd merge this into `ElasticsearchCaseStateFilterFactory`

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchUserCaseAccessFilterFactory.java
mario-paniccia: same for the other filter

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AuthorisedCaseDataFilter.java
mario-paniccia: I'm a bit concerned about the number of small security related classes we are introducing recently. Classes that just group operations. 
I'd be tempted to remove this class and move this code and it's dependencies into AuthorisedCaseDetailsSearchOperation, even thought I agree AuthorisedCaseDetailsSearchOperation would become more complex

gbenadikar: moved only couple of classes, if I move all access related classes, it affects 120 files. So this could be done as part of a separate tech debt.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
gbenadikar: renamed to jsonQuery

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
gbenadikar: Done

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
gbenadikar: fine, done

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
```diff
@@ -0,0 +1,51 @@
...
+        AuthorisedCaseDataFilter authorisedCaseDataFilter) {
+
+        this.caseDetailsSearchOperation = caseDetailsSearchOperation;
+        this.authorisedCaseDefinitionDataService = authorisedCaseDefinitionDataService;
+        this.authorisedCaseDataFilter = authorisedCaseDataFilter;
+    }
+
+    @Override
+    public CaseDetailsSearchResult execute(String caseTypeId, String query) {
+        return authorisedCaseDefinitionDataService
```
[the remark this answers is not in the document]
gbenadikar: simplified a little to make it more readable

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParser.java
gbenadikar: its harmless though, not a great deal to retain I guess

src/main/java/uk/gov/hmcts/ccd/domain/service/search/filter/UserAccessFilter.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.hmcts.ccd.domain.service.search.filter;
+
+import java.util.List;
+import java.util.Optional;
+
+import uk.gov.hmcts.ccd.domain.service.common.CaseAccessService;
+
+public class UserAccessFilter {
```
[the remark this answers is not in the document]
gbenadikar: I was organising code to separate domain logic from implementations out of my habit. Not applicable to this code base though, deleted.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AuthorisedCaseDataFilter.java
mario-paniccia: also see comment below on the implementation class

src/main/java/uk/gov/hmcts/ccd/domain/service/search/filter/CaseFilterFactory.java
gbenadikar: changed to non generic.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/DefaultAuthorisedCaseDataFilter.java
```diff
@@ -0,0 +1,48 @@
+package uk.gov.hmcts.ccd.domain.service.common;
```
mario-paniccia: I'm a bit concerned about the number of small security related classes we are introducing recently. Classes that just group operations for convenience and end up to be used only by one client (in this case AuthorisedCaseDetailsSearchOperation).
I'd be tempted to remove this class and move its code and dependencies into AuthorisedCaseDetailsSearchOperation, even thought I agree AuthorisedCaseDetailsSearchOperation would become more complex.
What's your take on this?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/filter/CaseSearchFilter.java
gbenadikar: Good point, done

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchUserCaseAccessFilterFactory.java
gbenadikar: it creates a filter that's why chose factory. Create looked better than applyFilter etc. Happy to change though.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/filter/CaseStateFilter.java
gbenadikar: merged

src/main/java/uk/gov/hmcts/ccd/domain/service/common/DefaultAuthorisedCaseDataFilter.java
gbenadikar: Well, this class should have been created in the first place and reused everywhere as the grouped code is repeated. That was the intention of creating it. Although, ok to move the logic to AuthorisedCaseDetailsSearchOperation and delete this class.

src/main/java/uk/gov/hmcts/ccd/CoreCaseDataConfiguration.java
```diff
@@ -14,4 +15,10 @@ public ObjectMapper objectMapper() {
         mapper.registerModule(new JavaTimeModule());
         return mapper;
     }
+
+    @Bean
+    @Qualifier("commonObjectMapper")
+    public ObjectMapper commonObjectMapper() {
```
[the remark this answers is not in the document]
gbenadikar: done

src/main/java/uk/gov/hmcts/ccd/CoreCaseDataConfiguration.java
```diff
@@ -2,16 +2,24 @@
...
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
 public class CoreCaseDataConfiguration {
 
     @Bean
+    @Qualifier("caseDetailsObjectMapper")
     public ObjectMapper objectMapper() {
```
mario-paniccia: can you rename the method as well please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseDetailsSearchOperation.java
```diff
@@ -2,12 +2,9 @@
...
 import uk.gov.hmcts.ccd.domain.model.search.CaseDetailsSearchResult;
 
-import java.io.IOException;
-import java.util.List;
-
 
 public interface CaseDetailsSearchOperation {
 
-    CaseDetailsSearchResult execute(List<String> caseTypeIds, String query) throws IOException;
+    CaseDetailsSearchResult execute(String caseTypeId, String jsonQuery);
```
mario-paniccia: shall we rename `jsonQuery` to `searchRequest`? because from the controller we pass down the whole request not only the query

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
```diff
@@ -0,0 +1,55 @@
...
+        AuthorisedCaseDefinitionDataService authorisedCaseDefinitionDataService,
+        AuthorisedCaseDataFilter authorisedCaseDataFilter) {
+
+        this.caseDetailsSearchOperation = caseDetailsSearchOperation;
+        this.authorisedCaseDefinitionDataService = authorisedCaseDefinitionDataService;
+        this.authorisedCaseDataFilter = authorisedCaseDataFilter;
+    }
+
+    @Override
+    public CaseDetailsSearchResult execute(String caseTypeId, String jsonQuery) {
```
mario-paniccia: shall we rename `jsonQuery` to `searchRequest` or `jsonSearchRequest`

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchQuerySecurity.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.ccd.domain.service.search.elasticsearch;
+
+public interface CaseSearchQuerySecurity {
+
+    String secureQuery(String caseTypeId, String query);
```
mario-paniccia: shall we rename this to `jsonQuery`?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchRequestFactory.java
```diff
@@ -0,0 +1,32 @@
+package uk.gov.hmcts.ccd.domain.service.search.elasticsearch;
+
+import static java.lang.String.format;
+
+import uk.gov.hmcts.ccd.ApplicationParams;
+
+public abstract class CaseSearchRequestFactory<T> {
```
mario-paniccia: are we operating on the overall search request in this class or only on the query? In the latter case, maybe better to rename this to `CaseSearchQueryFactory`
and its method to `createSearchQuery`

mario-paniccia: also, I'm thinking that the names factory and create are a bit misleading. Because we are not creating the query. The query comes from the client. 
We are transforming the query. So perhaps we should rename these to `CaseSearchQueryTransformer` and `transformSearchQuery`?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseDetailsSearchOperation.java
```diff
@@ -0,0 +1,77 @@
...
+                                                   CaseDetailsMapper caseDetailsMapper,
+                                                   CaseSearchRequestFactory<Search> caseSearchRequestFactory) {
+        this.jestClient = jestClient;
+        this.objectMapper = objectMapper;
+        this.caseDetailsMapper = caseDetailsMapper;
+        this.caseSearchRequestFactory = caseSearchRequestFactory;
+    }
+
+    @Override
+    public CaseDetailsSearchResult execute(String caseTypeId, String jsonQuery) {
```
mario-paniccia: can we rename jsonQuery to jsonSearchRequest please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchQuerySecurity.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.ccd.domain.service.search.elasticsearch;
```
mario-paniccia: I think we should organise better the package structure. I would move any class that deals with security under a .security subpackage

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseDetailsSearchOperation.java
```diff
@@ -0,0 +1,77 @@
...
+        SearchResult result = search(caseTypeId, jsonQuery);
+        if (result.isSucceeded()) {
+            return toCaseDetailsSearchResult(result);
+        } else {
+            throw new BadSearchRequest(result.getErrorMessage());
+        }
+    }
+
+    private SearchResult search(String caseTypeId, String jsonQuery) {
```
mario-paniccia: `jsonQuery` to `jsonSearchRequest` please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseSearchRequestFactory.java
```diff
@@ -0,0 +1,24 @@
...
+@Component
+public class ElasticsearchCaseSearchRequestFactory extends CaseSearchRequestFactory<Search> {
+
+    @Autowired
+    public ElasticsearchCaseSearchRequestFactory(ApplicationParams applicationParams, CaseSearchQuerySecurity caseSearchQuerySecurity) {
+        super(applicationParams, caseSearchQuerySecurity);
+    }
+
+    @Override
+    protected Search createSearchRequest(String caseTypeId, String query) {
```
mario-paniccia: any reason why this method has been declared as abstract and implemented in a subclass?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchRequestFactory.java
mario-paniccia: any reason why this is an abstract class with createSearchRequest not implemented directly here?

mario-paniccia: how come this is  a generic class?

mario-paniccia: the above seems a bit an uneeded complication to me. I would remove them and go for a simple concrete non generic class. If in the future the need arise, we can make it generic and create a hierarchy.
Unless I'm missing something

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQuerySecurity.java
```diff
@@ -0,0 +1,54 @@
...
+    private final List<CaseSearchFilterFactory> caseFilterFactories;
+
+    @Autowired
+    public ElasticsearchQuerySecurity(ElasticsearchQueryParserFactory queryParserFactory, List<CaseSearchFilterFactory> caseFilterFactories) {
+        this.queryParserFactory = queryParserFactory;
+        this.caseFilterFactories = caseFilterFactories;
+    }
+
+    @Override
+    public String secureQuery(String caseTypeId, String query) {
```
mario-paniccia: `query` -> `jsonSearchRequest` please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParserFactory.java
```diff
@@ -0,0 +1,20 @@
...
+public class ElasticsearchQueryParserFactory {
+
+    private final ObjectMapperService objectMapperService;
+
+    @Autowired
+    public ElasticsearchQueryParserFactory(ObjectMapperService objectMapperService) {
+        this.objectMapperService = objectMapperService;
+    }
+
+    ElasticsearchQueryParser createParser(String query) {
```
mario-paniccia: query to jsonSearchRequest please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParser.java
```diff
@@ -0,0 +1,51 @@
...
+ * }
+ */
+class ElasticsearchQueryParser {
+
+    private static final String QUERY_NAME = "query";
+
+    private final ObjectMapperService objectMapperService;
+    private final ObjectNode searchQueryObjectNode;
+
+    ElasticsearchQueryParser(ObjectMapperService objectMapperService, String query) {
```
mario-paniccia: query to jsonSearchRequest please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParser.java
mario-paniccia: the problem with this, is that it can create confusion. Imagine we add other validations to the search request in the controller. Then we might be tempted to add them here as well, or be undecided. That's how duplication builds up. 
Better to have all the validation of the SearchRequest only in the controller layer and fail fast

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AuthorisedCaseDataFilter.java
mario-paniccia: I meant to move only the new security classes we are creating in the new search api. Not the whole security classes

src/main/java/uk/gov/hmcts/ccd/domain/service/common/DefaultAuthorisedCaseDataFilter.java
mario-paniccia: if this code is repeated elsewhere then ok to leave this class, and perhaps in the future we can move some of that code to use this class as well

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
mario-paniccia: the point is that jsonSearchRequest and jsonQuery are 2 different things. From the controller we receive a jsonSearchRequest, which contains a query, size, from.... properties. 
This parameter is the whole request, not only the query

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AuthorisedCaseDataFilter.java
gbenadikar: moved new classes to security package

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
gbenadikar: Introduced new object CaseSearchRequest that converts Json string to JsonNode and validates existence of query field.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParser.java
gbenadikar: Moved logic to CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/CoreCaseDataConfiguration.java
gbenadikar: renamed

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseDetailsSearchOperation.java
gbenadikar: Passing the new request object everywhere CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/AuthorisedCaseDetailsSearchOperation.java
gbenadikar: Using CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchQuerySecurity.java
gbenadikar: using CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchRequestFactory.java
gbenadikar: Was just trying to enforce the security method is called. Although I agree, its unncessary complication. I have removed it now and simplified the code.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseDetailsSearchOperation.java
gbenadikar: Using CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchQuerySecurity.java
gbenadikar: done

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseDetailsSearchOperation.java
gbenadikar: Using CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseSearchRequestFactory.java
gbenadikar: Removed this class

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQuerySecurity.java
gbenadikar: Using CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParserFactory.java
gbenadikar: Using CaseSearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryParser.java
gbenadikar: Using CaseSearchRequest

gbenadikar:
Review comments addressed

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchRequest.java
```diff
@@ -0,0 +1,65 @@
...
+    private final ObjectMapperService objectMapperService;
+    private final String caseTypeId;
+    private final ObjectNode searchRequestJsonNode;
+
+
+    public CaseSearchRequest(ObjectMapperService objectMapperService, String caseTypeId, String jsonSearchRequest) {
+        this.objectMapperService = objectMapperService;
+        this.caseTypeId = caseTypeId;
+        this.searchRequestJsonNode = this.objectMapperService.convertStringToObject(jsonSearchRequest, ObjectNode.class);
+        validateJsonSearchRequest();
```
mario-paniccia: shall we move this line to be the first line?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchRequest.java
```diff
@@ -0,0 +1,65 @@
...
+public class CaseSearchRequest {
+
+    private static final String QUERY_NAME = "query";
+
+    private final ObjectMapperService objectMapperService;
+    private final String caseTypeId;
+    private final ObjectNode searchRequestJsonNode;
+
+
+    public CaseSearchRequest(ObjectMapperService objectMapperService, String caseTypeId, String jsonSearchRequest) {
```
mario-paniccia: 1) Can we make this class immutable? i.e. no setter. It'll contain the original request only.
When we create the secured one, we'll create a new instance of CaseSearchRequest which contains the secured query

mario-paniccia: 2. Shall we leave the objectMapperService outside of this class? i.e. the client has to build a ObjectNode and pass it in.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/security/ElasticsearchCaseSearchRequestSecurity.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Autowired
+    public ElasticsearchCaseSearchRequestSecurity(List<CaseSearchFilter> caseSearchFilters) {
+        this.caseSearchFilters = caseSearchFilters;
+    }
+
+    @Override
+    public void secureRequest(CaseSearchRequest caseSearchRequest) {
+        String queryClauseWithSecurityFilters = addFiltersToQuery(caseSearchRequest);
+        caseSearchRequest.replaceQuery(queryClauseWithSecurityFilters);
```
mario-paniccia: I'd create a totally new SearchRequest

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/CaseSearchRequest.java
gbenadikar: changes done. CaseSearchRequest is now immutable

PR #217, opened 2018-10-04
[PREVIEW] Optional elastic url data

mario-paniccia:
Ignoring retrieval of es url from key vault when ES is not enabled, to prevent missing value error from the vault

```
[ ] Yes
[ x] No
```


infrastructure/main.tf
```diff
@@ -102,7 +103,7 @@ module "ccd-data-store-api" {
 
     CCD_DEFAULTPRINTURL                 = "${local.default_print_url}"
 
-    ELASTIC_SEARCH_HOSTS                = "${format("http://%s:9200", data.azurerm_key_vault_secret.ccd_elastic_search_url.value)}"
+    ELASTIC_SEARCH_HOSTS                = "${var.elastic_search_enabled == "false" ? "" : "${format("http://%s:9200", join("", data.azurerm_key_vault_secret.ccd_elastic_search_url.*.value))}"}"
```
mario-paniccia: horrible hack with the join. But terraform ternary operator is broken and otherwise would always evaluate both of the sides

dwaynebailey: You need the `*`?

infrastructure/sprod.tfvars
```diff
@@ -1,3 +1,3 @@
 capacity = "2"
 asp_name = "ccd-data-store-api-sprod"
-asp_rg = "ccd-data-store-api-sprod"
+asp_rg = "ccd-data-store-api-sprod"
```
dwaynebailey: Can we drop this change

infrastructure/main.tf
mario-paniccia: yes

PR #219, opened 2018-10-04
[PREVIEW] Remove conflicting ObjectMapper config

vlaurin:

N/A

### Change description ###

Remove ObjectMapper config clashing with JacksonObjectMapperConfig

```
[ ] Yes
[x] No
```


vlaurin:
@mario-paniccia This PR does resolve the conflict as proved by the green functional tests in Jenkins pipeline. What was the motivation of the ObjectMapper with a different config that was introduced?

mario-paniccia:
@vlaurin 
for the deserialisation of ES cases I needed a custom object mapper. I thought of declaring it in a config class and injecting it rather than doing manual instantiation. 
Also, the goal was to have this config class as a central place where to configure all the ObjectMapper used in data store. At the moment OM creation is scattered all over the place, so it's difficult to understand the serialisation behaviour of Data Strore. Was planning to slowly refactor this.

PR #233, opened 2018-10-18
RDM-1202 - Case event fields label and hint overrides

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-1202

### Change description ###
Code changes to return case fields for events with label and hint text overrides if available

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventField.java
```diff
@@ -66,4 +68,30 @@ public Integer getShowSummaryContentOption() {
     public void setShowSummaryContentOption(Integer showSummaryContentOption) {
         this.showSummaryContentOption = showSummaryContentOption;
     }
+
+    /**
+     * event case field label
+     **/
+    @ApiModelProperty(required = true, value = "")
```
tzarsmango: should this be required?

gbenadikar: changed to optional, thanks

PR #235, opened 2018-10-19
RDM-2840 Get Case by API

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-2840
Added CaseRoles Repository for getting the case roles from ccd-case-definition-store and injecting them to the list of user roles.

```
[ ] Yes
[X] No
```


src/test/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpointIT.java
```diff
@@ -6,11 +6,7 @@
 
 import static com.google.common.collect.Lists.newArrayList;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.hasItem;
-import static org.hamcrest.Matchers.hasProperty;
-import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.*;
 import static org.mockito.Matchers.anyObject;
```
adr1ancho: Not sure checkstyle would complain about the use of `*`

src/test/java/uk/gov/hmcts/ccd/domain/service/common/DefaultAuthorisedCaseDefinitionDataServiceTest.java
```diff
@@ -1,10 +1,6 @@
 package uk.gov.hmcts.ccd.domain.service.common;
 
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
```
adr1ancho: Not sure whether checkstyle would complain about the use of `*`

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpointIT.java
tzarsmango: annoying but somehow Codecov does. Actually, I was expecting your return for this :D

src/main/java/uk/gov/hmcts/ccd/domain/service/getcase/AuthorisedGetCaseOperation.java
```diff
@@ -31,43 +34,43 @@
...
 
     private CaseType getCaseType(String caseTypeId) {
         return caseDefinitionRepository.getCaseType(caseTypeId);
     }
 
-    private Set<String> getUserRoles() {
-        return userRepository.getUserRoles();
+
+    private Set<String> getUserRoles(String caseTypeId) {
+        return Sets.union(userRepository.getUserRoles(), caseRoleRepository.getCaseRoles(caseTypeId));
```
vlaurin: I believe this is wrong, the case roles defined as part of the case type are only the list of valid case roles, but they do not represent the case role owned by the current user. Instead, the user case role (if any) must be read from the table `case_users` for the current case ID and user ID.

src/main/java/uk/gov/hmcts/ccd/data/user/UserRepository.java
```diff
@@ -17,4 +17,6 @@
     Set<SecurityClassification> getUserClassifications(String jurisdictionId);
 
     SecurityClassification getHighestUserClassification(String jurisdictionId);
+
+    String getUserName();
```
vlaurin: This user name is in fact the user ID, could the function be renamed to `getUserID()`?

src/main/java/uk/gov/hmcts/ccd/data/caseaccess/CaseUserRepository.java
```diff
@@ -43,4 +43,12 @@ public void revokeAccess(final Long caseId, final String userId) {
 
         return namedQuery.getResultList();
     }
+
+    public List<String> findCaseRolesUserHasForACase(final Long caseId, final String userId) {
```
vlaurin: The method name would be more readable if shorter. Could it be changed to `findCaseRoles()`? I think the parameters provide enough indications as to the filters used by the query.

src/main/java/uk/gov/hmcts/ccd/data/user/UserRepository.java
tzarsmango: 👍

src/main/java/uk/gov/hmcts/ccd/data/caseaccess/CaseUserRepository.java
tzarsmango: 👍

PR #258, opened 2018-11-16
Attempt to enable ES functional test on AAT

gbenadikar:
N/A

### Change description ###
Attempt to enable ES functional test on AAT

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/outputs.tf
+++ b/infrastructure/outputs.tf
@@ -29,3 +29,19 @@ output "OAUTH2_CLIENT_ID" {
 output "OAUTH2_REDIRECT_URI" {
   value = "${local.oauth2_redirect_uri}"
 }
+
+output "ELASTIC_SEARCH_ENABLED" {
+  value = "${var.elastic_search_enabled}"
+}
+
+output "ELASTIC_SEARCH_HOST" {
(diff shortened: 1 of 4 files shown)


infrastructure/outputs.tf
```diff
@@ -29,3 +29,19 @@ output "OAUTH2_CLIENT_ID" {
 output "OAUTH2_REDIRECT_URI" {
   value = "${local.oauth2_redirect_uri}"
 }
+
+output "ELASTIC_SEARCH_ENABLED" {
+  value = "${var.elastic_search_enabled}"
+}
+
+output "ELASTIC_SEARCH_HOST" {
```
mario-paniccia: do we need the port and the scheme in data store? At the moment they are passed here
`http://%s:9200"`
Unless we change the line above, having a var for scheme and port might be misleading

infrastructure/variables.tf
```diff
@@ -121,3 +121,11 @@ variable "elastic_search_case_index_type" {
   type = "string"
   default = "_doc"
 }
+
+variable "elastic_search_port" {
+  default = "9200"
+}
```
mario-paniccia: same as above

PR #283, opened 2018-12-12
RDM-3268 Authorisation definitions must be honored

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-3268

Added filtering for CaseFields displayed in tabs. The fields with no READ access are removed from the display. So if the current user doesn't have read access for a Payment History field it won't be displayed in case view.

Additionally if a tab has no fields to display, then it is removed and not displayed in the view case page.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewField.java
```diff
@@ -32,6 +34,8 @@
     private Boolean showSummaryChangeOption;
     @JsonProperty("show_summary_content_option")
     private Integer showSummaryContentOption;
+    @JsonProperty("acls")
```
vlaurin: Shouldn't this field be ignored from the JSON payload? Otherwise it would go against the principle that users are not meant to know they don't have access to something. In that case, given the filtering is already performed server-side, I guess we don't need those ACLs client side anyway do we?

```suggestion
    @JsonIgnore
```

PR #305, opened 2019-01-04
Rdm 3562 Implement internal Get WorkbasketInputDetails endpoint in API v2

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3562

### Change description ###

v1 endpoint: https://github.com/hmcts/ccd-data-store-api/blob/master/src/main/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpoint.java#L172

v2: GET /internal/case-types/{caseTypeId}/work-basket-inputs

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UIDefinitionController.java
```diff
@@ -0,0 +1,68 @@
...
+    )
+    @ApiResponses({
+        @ApiResponse(
+            code = 200,
+            message = "Success",
+            response = UIWorkbasketInputsResource.class
+        ),
+        @ApiResponse(
+            code = 404,
+            message = "Definition could not be found in Definition Store"
```
vlaurin:
```suggestion
            message = "Case type not found"
```

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UIDefinitionController.java
```diff
@@ -0,0 +1,68 @@
...
+        ),
+        @ApiResponse(
+            code = 500,
+            message = "Definition Store is down"
```
vlaurin: That seems a bit definitive, lots of things could cause a 500. I think keeping it without a message is okay, 500 conveys the meaning "something unexpected happened"

src/main/java/uk/gov/hmcts/ccd/v2/internal/resource/UIWorkbasketInputsResource.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.methodOn;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@NoArgsConstructor
+public class UIWorkbasketInputsResource extends ResourceSupport {
+
+    @JsonUnwrapped
+    private WorkbasketInput[] workbasketInputs;
```
vlaurin: Instead of unwrapping it, I think it would be better to duplicate the contract and copy each field across. While it does seem wrong, it achieves the primary goal of that class, which is to act as a definitive contract for the API. I'm concerned that by relying on `WorkBasketInput` for the contract, it could easily be broken by mistake because it may not be clear that `WorkbasketInput` is the contract

andrewfolga: I think the unwrapping is not needed here and I will remove. Are you suggesting I should copy the `WorkbasketInput` class exactly as is and call it `UIWorkbasketInput` and then have an array of `UIWorkbasketInput` inputs isntead?

PR #307, opened 2019-01-04
RDM-2836 2/3: Set case roles: Update user access

vlaurin:

RDM-2836

### Change description ###

* Update user accesses by granting and/or revoking roles to match target state

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
@@ -65,7 +65,7 @@ public class CaseAccessOperation {
         final Set<String> targetCaseRoles = caseUser.getCaseRoles();
 
         targetCaseRoles.stream()
-                       .filter((role) -> !validCaseRoles.contains(role))
+                       .filter(role -> !validCaseRoles.contains(role))
                        .findFirst()
                        .ifPresent(role -> {
                            throw new InvalidCaseRoleException(role);
@@ -76,12 +76,12 @@ public class CaseAccessOperation {
 
         // Grant added case roles
         targetCaseRoles.stream()
-                       .filter((targetRole) -> !currentCaseRoles.contains(targetRole))
+                       .filter(targetRole -> !currentCaseRoles.contains(targetRole))
                        .forEach(targetRole -> caseUserRepository.grantAccess(caseId, caseUser.getUserId(), targetRole));
 
         // Revoke removed case roles
         currentCaseRoles.stream()
-                        .filter((currentRole) -> !targetCaseRoles.contains(currentRole))
+                        .filter(currentRole -> !targetCaseRoles.contains(currentRole))
                         .forEach(currentRole -> caseUserRepository.revokeAccess(caseId,
                                                                                 caseUser.getUserId(),
                                                                                 currentRole));
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
@@ -64,26 +64,42 @@ public class CaseAccessOperation {
         final Set<String> validCaseRoles = caseRoleRepository.getCaseRoles(caseDetails.getCaseTypeId());
         final Set<String> targetCaseRoles = caseUser.getCaseRoles();
 
+        validateCaseRoles(validCaseRoles, targetCaseRoles);
+
+        final Long caseId = new Long(caseDetails.getId());
+        final String userId = caseUser.getUserId();
+        final List<String> currentCaseRoles = caseUserRepository.findCaseRoles(caseId, userId);
+
+        grantAddedCaseRoles(userId, caseId, currentCaseRoles, targetCaseRoles);
+        revokeRemovedCaseRoles(userId, caseId, currentCaseRoles, targetCaseRoles);
+    }
+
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
```diff
@@ -50,4 +59,31 @@ public void revokeAccess(final String jurisdictionId, final String caseReference
                                  .collect(Collectors.toList());
     }
 
+    @Transactional
+    public void updateUserAccess(CaseDetails caseDetails, CaseUser caseUser) {
+        final Set<String> validCaseRoles = caseRoleRepository.getCaseRoles(caseDetails.getCaseTypeId());
+        final Set<String> targetCaseRoles = caseUser.getCaseRoles();
+
+        targetCaseRoles.stream()
```
andrewfolga: Maybe extraction to a self describing validation method would improve readability here?

src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
```diff
@@ -50,4 +59,31 @@ public void revokeAccess(final String jurisdictionId, final String caseReference
...
+                       .filter(role -> !validCaseRoles.contains(role))
+                       .findFirst()
+                       .ifPresent(role -> {
+                           throw new InvalidCaseRoleException(role);
+                       });
+
+        final Long caseId = new Long(caseDetails.getId());
+        final List<String> currentCaseRoles = caseUserRepository.findCaseRoles(caseId, caseUser.getUserId());
+
+        // Grant added case roles
```
andrewfolga: Same here we could convert this block of code to a method with a signature name reflecting the comment.

src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
```diff
@@ -50,4 +59,31 @@ public void revokeAccess(final String jurisdictionId, final String caseReference
...
+        targetCaseRoles.stream()
+                       .filter(targetRole -> !currentCaseRoles.contains(targetRole))
+                       .forEach(targetRole -> caseUserRepository.grantAccess(caseId, caseUser.getUserId(), targetRole));
+
+        // Revoke removed case roles
```
andrewfolga: Same as above.

src/test/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperationTest.java
```diff
@@ -85,6 +101,59 @@ void shouldHandleWrongJurisdiction() {
...
+
+        @Test
+        @DisplayName("should ignore case roles already granted")
+        void shouldIgnoreGrantedCaseRoles() {
+            caseAccessOperation.updateUserAccess(caseDetails, caseUser(CASE_ROLE_GRANTED));
+
+            verify(caseUserRepository, never()).grantAccess(CASE_ID, USER_ID, CASE_ROLE_GRANTED);
+        }
+    }
+
```
andrewfolga: Are we missing a case where there is an update without case roles?
```
Given a case and a user
When granting access to the case for the user ID WITHOUT case role
Then 400 Case Role mandatory 
And user is not granted access
```

andrewfolga: Same for a case where there is invalid case id?

```
Given an invalid case ID and a user
When granting access to the case for the user ID with a valid case role
Then 400 invalid case ID
And user is not granted access
```

andrewfolga: I guess the above + few other ACs are checked on another level before this logic.

src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
vlaurin: Indeed, very good point, done

src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
vlaurin: That makes perfect sense, done

src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
vlaurin: Same as above :)

src/test/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperationTest.java
vlaurin: High-level input validation, like ID checksum, are checked one level above in subsequent PR #308 . However, I've actually changed the implementation to accept update without case roles, as I felt this was needed for `PUT` to keep its true meaning: PUT without roles => remove all accesses. Otherwise, this would require `POST` + `DELETE` and it doesn't feel as nice.
But I did forget to update the ticket, sorry for that 🙇

andrewfolga: Ok makes sense, thanks.

PR #308, opened 2019-01-04
RDM-2836 3/3: Set case roles: Endpoint

vlaurin:

RDM-2836

### Change description ###

- Add `PUT /cases/{caseId}/users/{userId}` endpoint
- Add swagger doc for endpoint
- Extract API errors into centralised class

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/infrastructure/user/UserAuthorisation.java
+++ b/src/main/java/uk/gov/hmcts/ccd/infrastructure/user/UserAuthorisation.java
@@ -1,13 +1,17 @@
 package uk.gov.hmcts.ccd.infrastructure.user;
 
+import java.util.Set;
+
 public class UserAuthorisation {
 
     private final String id;
     private final AccessLevel accessLevel;
+    private final Set<String> roles;
 
-    public UserAuthorisation(String id, AccessLevel accessLevel) {
+    public UserAuthorisation(String id, AccessLevel accessLevel, Set<String> roles) {
         this.id = id;
         this.accessLevel = accessLevel;
+        this.roles = roles;
     }
 
     public String getUserId() {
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/ccd/infrastructure/user/UserAuthorisation.java
```diff
@@ -18,6 +22,15 @@ public AccessLevel getAccessLevel() {
         return accessLevel;
     }
 
+    public Boolean hasJurisdictionRole(String jurisdictionId) {
+        final String jurisdictionRole = "caseworker-" + jurisdictionId;
```
andrewfolga: Is this always going to be for caseworker? Is it worth to parametrize the method further?

vlaurin: Good question. Yes, it is always to be caseworker.

My thinking is that at this point the format of base jurisdiction role (`caseworker-{jid}`) has become a core concept of this project, and is also hard-coded in other places of this project (see `authorizedRolesExtractor()`) , and other projects as well (Document Store, API Gateway). 

Because of that, a different pattern could not be used without refactoring of at least 3 projects and a migration of roles for our IDAM users.

So, should it be parameterised? I wouldn't think so, that would give the expectation it can easily be changed when in fact it can't.

src/main/java/uk/gov/hmcts/ccd/v2/external/controller/CaseUserController.java
```diff
@@ -0,0 +1,111 @@
...
+            message = V2.Error.CASE_ROLE_INVALID
+        ),
+        @ApiResponse(
+            code = 403,
+            message = V2.Error.GRANT_FORBIDDEN
+        ),
+        @ApiResponse(
+            code = 404,
+            message = V2.Error.CASE_NOT_FOUND
+        )
```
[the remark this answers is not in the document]
vlaurin: Yes, because in the end I went for multiple case roles support, which was mutually exclusive.
Once again, forgot to update Jira, no excuse 🙇

src/main/java/uk/gov/hmcts/ccd/infrastructure/user/UserAuthorisation.java
andrewfolga: Ok fair enough.

PR #327, opened 2019-01-23
change smoke test to a quicker one rather than search to avoid failin…

mario-paniccia:
change smoke test to a quicker one rather than search to avoid failing build

```
[ ] Yes
[ x] No
```


mario-paniccia:
@vlaurin I thought in case of deployment failure it would be much more likely a 50X type of error. I was not expecting a 40X error. But it's a good point, let's eliminate the risk. Will pick a new smoke test then

PR #331, opened 2019-01-25
RDM-3695 CRUD on Collection - Fix to propagate ACLs to the children case fields

tzarsmango:
The front end was failing since the nested fields does not have any ACLs. After a chat with architect, the ACLS are good to propagate to the embedded case fields.

```
[ ] Yes
[X] No
```


dependency-check-suppressions.xml
```diff
@@ -66,6 +66,38 @@
...
+        <cve>CVE-2018-19360</cve>
+    </suppress>
+    <suppress>
+        <notes>Temporary suppression</notes>
+        <cve>CVE-2018-19361</cve>
+    </suppress>
+    <suppress>
+        <notes>Temporary suppression</notes>
+        <cve>CVE-2018-19362</cve>
+    </suppress>
```
vlaurin: Could we exclude the suppressions from this PR so that they are properly addressed independently?

PR #342, opened 2019-01-31
[PREVIEW] change cache manager from hazelcast to ehcache

mario-paniccia:
We are having problems with definitions caching on AAT. Looks like the cache entries get removed prematurely when they shouldn't. 
After many attempts, I'd like to change the caching provider from Hazelcast to EhCache and see if that performs better

```
[ ] Yes
[x ] No
```


mario-paniccia:
Will test this on Demo for a while before issuing PR

mario-paniccia:
closing this as it looks like it's causing issues

PR #371, opened 2019-02-14
RDM-3989 - Enabled Elasticsearch nodes discovery for Jest client

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-3989

### Change description ###
Enabled ES nodes discovery or Jest client to prevent Jest from sending any requests to dead ES nodes

```
[ ] Yes
[x] No
```


infrastructure/variables.tf
```diff
@@ -137,6 +137,24 @@ variable "elastic_search_case_index_type" {
...
+variable "elastic_search_nodes_discovery_enabled" {
+  description = "Enable Elasticsearch node discovery by Jest client"
+  type = "string"
+  default = "true"
+}
+
+variable "elastic_search_nodes_discovery_frequency_millis" {
+  description = "Elasticsearch node discovery frequency in milliseconds"
+  type = "string"
+  default = "1000"
```
gbenadikar: Defaulting discovery check to 1 second. Let me know if this is low or high

infrastructure/variables.tf
```diff
@@ -137,6 +137,24 @@ variable "elastic_search_case_index_type" {
...
+}
+
+variable "elastic_search_nodes_discovery_filter" {
+  description = "Elasticsearch node discovery filter"
+  type = "string"
+  default = "_all"
```
gbenadikar: defaulting discovery of nodes to all ES nodes. This can be filtered to specific node patterns e.g. master, data etc. As all our ES nodes are data nodes, just using the all for now

infrastructure/variables.tf
mario-paniccia: seems ok to me

infrastructure/variables.tf
mario-paniccia: ok

src/main/java/uk/gov/hmcts/ccd/ElasticSearchConfiguration.java
```diff
@@ -15,20 +15,28 @@
...
         factory.setHttpClientConfig(new HttpClientConfig.Builder(applicationParams.getElasticSearchDataHosts())
-            .multiThreaded(true)
-            .maxConnectionIdleTime(15, TimeUnit.SECONDS)
-            .connTimeout(4000)
-            .readTimeout(4000)
-            .gson(gson).build());
+                                        .multiThreaded(true)
+                                        .maxConnectionIdleTime(15, TimeUnit.SECONDS)
+                                        .connTimeout(4000)
+                                        .readTimeout(4000)
```
mario-paniccia: since we're here, can you increase a bit the read timeout? let's do 6 secs?

src/main/resources/application.properties
```diff
@@ -135,6 +135,9 @@ search.elastic.data.hosts=${ELASTIC_SEARCH_DATA_NODES_HOSTS:"http://localhost:92
 search.blacklist=${ELASTIC_SEARCH_BLACKLIST:query_string}
 search.cases.index.name.format=${ELASTIC_SEARCH_CASE_INDEX_NAME_FORMAT:%s_cases}
 search.cases.index.name.type=${ELASTIC_SEARCH_CASE_INDEX_TYPE:_doc}
+search.elastic.nodes.discovery.enabled=${ELASTIC_SEARCH_NODES_DISCOVERY_ENABLED:true}
```
mario-paniccia: do we need this enabled by default for local development?

infrastructure/variables.tf
mario-paniccia: actually, perhaps a bit too aggressive? how about 30 secs?

infrastructure/variables.tf
gbenadikar: we can easily restrict the discovery to data nodes only in that case

src/main/java/uk/gov/hmcts/ccd/ElasticSearchConfiguration.java
gbenadikar: done

src/main/resources/application.properties
gbenadikar: right, setting to false for local

infrastructure/variables.tf
gbenadikar: set to 5 secs

PR #388, opened 2019-03-04
RDM-4060 - Data store - get case V2 api improvement

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-4060

### Change description ###
1. Improved case access service to retrieve user roles from the spring security context and not make expensive IDAM call. 
2. Introduced caching on user roles - security classification retrieval from definition store

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -127,10 +131,11 @@ public UserRole getUserRoleClassifications(String userRole) {
     }
 
     @Override
+    @Cacheable("userRolesCache")
```
gbenadikar: This method makes a call to the definition store to retrieve security classification associated with the roles. I think the security classification mapping on a role does not change often and this can be cached.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
```diff
@@ -45,11 +43,7 @@ public CaseAccessService(@Qualifier(CachedUserRepository.QUALIFIER) UserReposito
     }
 
     public Boolean canUserAccess(CaseDetails caseDetails) {
-
-        IDAMProperties currentUser = userRepository.getUserDetails();
```
gbenadikar: This is an expensive call to IDAM that returns user id and roles. User id and roles are already available in the spring security context and has been implemented in the UserRepository.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
dwaynebailey: How long is this cached?  I assume a change on def-store i.e. new definition would be the only way for this to change.  Do we have anything from def-store that trigger a cache invalidation on data-store?

gbenadikar: Its cached for 72 hours by default. Currently there is no mechanism for invalidating cache. What the endpoint does is retrieve security classifications for the roles passed. I assume the security classification on roles don't change. When we add a role, we assign security classification. I am not aware if security classification on a role can be changed. @mario-paniccia may know.

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
```diff
@@ -49,13 +49,17 @@ public UserRole getUserRoleClassifications(String userRole) {
...
+        List<String> missingRoles = userRoles
+            .stream()
             .filter(role -> !userRoleClassifications.containsKey(role))
+            .sorted()
             .collect(Collectors.toList());
+
         if (!missingRoles.isEmpty()) {
-            final List<UserRole> missingClassifications = caseDefinitionRepository.getClassificationsForUserRoleList(missingRoles);
-            missingClassifications.forEach(userClassification -> userRoleClassifications.putIfAbsent(userClassification.getRole(), userClassification));
+            caseDefinitionRepository.getClassificationsForUserRoleList(missingRoles)
```
mario-paniccia: I think the previous version with the explicit variable `missingClassifications` makes the code more clear. And has not impact on performance

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
mario-paniccia: I think we can refactor this to make sure caching works to its full potential. Let me explain. Imagine we have 10 user roles, each with its own classification. Because we try to cache the result of requesting 
a list of user roles, this means having to cache a huge number of potential combinations of those 10 roles. This mean lower probability the cached value is used in the future (cache hit).
Instead we could refactor the method to accept one single user role. Once we do that, at most the cache will cache 10 entries. This will increase the cache hit percentage.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
mario-paniccia: But @gbenadikar given that the UserRepository injected is a CachedUserRepository, which is caching the getUserDetails() method, aren't we retrieving that from the cache already?

gbenadikar: CaseViewType is request scope, doesn't survive beyond the request.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
gbenadikar: good point, done.

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
gbenadikar: I found the later more readable 😄 . Anyway reverted to previous version

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
mario-paniccia: yes correct but same is the spring security context right which we are switching to now

gbenadikar: spring security context holds the authentication token passed with the request and derives the user roles from it without making explicit call to IDAM. User.getUserDetails makes explicit IDAM API call to retrieve user roles and I found it to be expensive.

PR #394, opened 2019-03-06
Rdm 2523 sql injection fix

MSancaktutar:
Jira Link:
https://tools.hmcts.net/jira/browse/RDM-2523
```
[ ] Yes
[X] No
```


src/test/java/uk/gov/hmcts/ccd/data/casedetails/search/SearchDirectionTest.java
```diff
@@ -0,0 +1,30 @@
+package uk.gov.hmcts.ccd.data.casedetails.search;
+
+import java.util.Optional;
+
+import org.junit.Test;
+
+import org.junit.Assert;
+
+public class SearchDirectionTest {
```
rafalkalita: please update to SortDirectionTest

MSancaktutar: My silly semantic dyslexia.. :)
Corrected. Thank you.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortDirection.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.ccd.data.casedetails.search;
+
+import java.util.Optional;
+
+public enum SortDirection {
+    ASC, DESC;
+
+    public static SortDirection fromOptionalString(Optional<String> direction) {
```
mario-paniccia: do we really need this method? if you pass a String rather than an Optional you cold leverage the java provided enum method valueOf.
you could deal with the Optional on the caller. Something like: 
metadata.getSortDirection().map(direction -> SortDirection.valueOf(direction)).orElse(SortDirection. DESC)

mario-paniccia: feel free to ignore

MSancaktutar: As far as I understood, the relevant endpoint is accepting the sort direction in a case insensitive manner and moreover, anything other than asc or desc is treated as asc, as per the current functionality.
 
The reason I didn't go with implicit valueOf(String) is because it throws an IllegalArgumentException when the parameter is not a case-sensitively correct element name. If I was to use it, the rest endpoint would probably return an http-500. Now, the custom mapping method provides the case-insensitive mapping functionality as localised and centralised in the most directly relevant code unit, to enable reusability. Having the case-insensivity or any dummy value to ASC mapping inside the query factory or elsewhere would replicate the same business rule in different locations.

veeresha-hmcts:
Verified locally and ready for merge.

PR #399, opened 2019-03-12
RDM-4030 - cache IDAM user full name

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-4030

### Change description ###
Introduced caching for retrieving IDAM user. There is an IDAM call made for every event to retrieve user's first name and last name. This can be cached for 8 hours as you wouldn't anticipate name changes every hour. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
@@ -1,6 +1,10 @@
 package uk.gov.hmcts.ccd.data.user;
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -61,6 +67,13 @@ public IDAMProperties getUserDetails() {
         return restTemplate.exchange(applicationParams.idamUserProfileURL(), HttpMethod.GET, requestEntity, IDAMProperties.class).getBody();
     }
 
+    @Override
```
gbenadikar: Caching this method for 8 hours, should lend benefits on case events. The call to IDAM was being made just to retrieve the user's first name and last name which then is stored in case event. I don't think user details like first name and last name will change often. What do you think?

PR #400, opened 2019-03-12
RDM-4030 - cache IDAM user full name

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-4030

### Change description ###
Introduced caching for retrieving IDAM user. There is an IDAM call made for every event to retrieve user's first name and last name. This can be cached for 8 hours as you wouldn't anticipate name changes every hour. 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -61,6 +67,13 @@ public IDAMProperties getUserDetails() {
         return restTemplate.exchange(applicationParams.idamUserProfileURL(), HttpMethod.GET, requestEntity, IDAMProperties.class).getBody();
     }
 
+    @Override
```
gbenadikar: Caching this method for 8 hours, should lend benefits on case events. The call to IDAM was being made just to retrieve the user's first name and last name which then is stored in case event. I don't think user details like first name and last name will change often. What do you think?

dwaynebailey: I think the problem is not so much that it doesn't change often its that when it changes we'll be wrong for a period of time.  Would it make sense to reduce the period to say 1 hour? Or does that lose all the benefits?

gbenadikar: Yes we could lower it down. Although, lower the time, lesser the benefits. The cache key is the user token, so as long as the user logs out and logs back in (I assume any user detail changes on IDAM will be required to log off and log back in to take effect).

src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
```diff
@@ -67,6 +67,9 @@
     @Value("${definition.cache.ttl.secs}")
     private Integer definitionCacheTTLSecs;
 
+    @Value("${daily.cache.ttl.secs}")
```
mario-paniccia: would change this to `user.cache.ttl.secs`. Much more descriptive. Same for the property name

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
mario-paniccia: Ok for me for caching this info, but I agree we should lower the caching length down a bit.
But we have to be careful here. I don't recall exactly now please @gbenadikar check the documentation. But a TTL of say 1 hour means that **if the cache entry it's not used** within 1 hour, then the entry is evicted. But what happens if that entry is used again and again instead? i.e. if we never reach the TTL? It means we could be returning the outdated IDAM info for days, weeks...To be confirmed

src/main/resources/application.properties
```diff
@@ -24,6 +24,7 @@ liquibase.change-log=classpath:/db/changelog/db.changelog-master.xml
 #definitions cache parameters
 #ttl of cache map entry
 definition.cache.ttl.secs=${DEFINITION_CACHE_TTL_SEC:259200}
+daily.cache.ttl.secs=28800
```
mario-paniccia: can you add a description please

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
```diff
@@ -35,6 +38,7 @@ private void configCaches(int definitionCacheTTL, Config config) {
         config.addMapConfig(newMapConfig("caseTabCollectionCache", definitionCacheTTL));
         config.addMapConfig(newMapConfig("wizardPageCollectionCache", definitionCacheTTL));
         config.addMapConfig(newMapConfig("userRolesCache", definitionCacheTTL));
+        config.addMapConfig(newMapConfig("userCache", applicationParams.getDailyCacheTTLSecs()));
```
mario-paniccia: can you add a test to the DefinitionsCachingIT please

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
mario-paniccia: Furthermore, if the user was deleted in IDAM, the current code would somehow throw exception I think. Because it wouldn't find the requested user id. 
The new code would allow the case creation to happen anyway. Not a blocker but something to keep in mind

mario-paniccia: @gbenadikar I think my comment above was right, check https://docs.hazelcast.org/docs/latest-dev/manual/html-single/index.html#map
more precisely `max-idle-seconds` which is what we set the ttl config to

src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
gbenadikar: done

src/main/resources/application.properties
gbenadikar: done

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
gbenadikar: done

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
gbenadikar: @mario-paniccia  it is misleading to have TTL set to max idle time on hazelcast cache config. Do you know why we do that? In place of `setMaxIdleSeconds(definitionCacheTTL)`, we should ideally be using `setTimeToLiveSeconds(definitionCacheTTL)`
or atleast renaming the variable definitionCacheTTL

gbenadikar: @mario-paniccia @dwaynebailey I have lowered down the max idle time for user cache to 4 hours. Bear in mind, the key for this cache is the user token and user id.

gbenadikar: @mario-paniccia as the cache key is the user token, when a user is deleted/suspended, we should expect the user token to be invalidated by IDAM. The CCD gateway then will block the api call anyway.

PR #401, opened 2019-03-12
RDM-876: Hide Print button when not configured

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-876

### Change description ###
As a court staff I do not want to see the Print button if the print callback url is not configured so that I do not have any confusion.

ccd-case-ui-toolkit PR (related):
hmcts/ccd-case-ui-toolkit#270

ccd-case-management-web PR (related):
hmcts/ccd-case-management-web#590

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewType.java
```diff
@@ -51,10 +53,15 @@ public void setJurisdiction(CaseViewJurisdiction jurisdiction) {
         this.jurisdiction = jurisdiction;
     }
 
+    public String getPrintableDocumentUrl() { return printableDocumentUrl; }
+
+    public void setPrintableDocumentUrl(String printableDocumentUrl) { this.printableDocumentUrl = printableDocumentUrl; }
```
MSancaktutar: I would normally expect the indentation to be preserved, but if it is accepted as per the team's / reform's standard, than it's ok to me.

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewType.java
```diff
@@ -7,16 +7,18 @@
     private String name;
     private String description;
     private CaseViewJurisdiction jurisdiction;
+    private String printableDocumentUrl;
```
MSancaktutar: Could you please fix (or confirm the correctness of) the plurality of this attribute?

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewType.java
MKaramath: Agreed, but since it is maintained this standard in the rest of the places, I followed the same.

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewType.java
MKaramath: Done.

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseViewType.java
```diff
@@ -7,16 +7,18 @@
     private String name;
     private String description;
     private CaseViewJurisdiction jurisdiction;
+    private String printableDocumenstUrl;
```
gbenadikar: spell typo?

MKaramath:
> this implementation is ok. But considering that the print url is not used by the UI, it would perhaps have been better the backend not passing the url. Instead the backend could just have passed a boolean property called 'isPrintEnabled'.

It is good to have boolean property such as isPrintEnabled, but the property printableDocumentsUrl already existed in definitions API, so I just propagated the same missing property into data store API and hence into the UI, which could be useful if it is required to use in the UI in future

mario-paniccia:
> > this implementation is ok. But considering that the print url is not used by the UI, it would perhaps have been better the backend not passing the url. Instead the backend could just have passed a boolean property called 'isPrintEnabled'.
> 
> It is good to have boolean property such as isPrintEnabled, but the property printableDocumentsUrl already existed in definitions API, so I just propagated the same missing property into data store API and hence into the UI, which could be useful if it is required to use in the UI in future

yes it exists on DefinitionStore and DataStore because it's used there. But UI does not need to know about the url and will never use the url. Hence, let's just pass a boolean property rather than the url.

MKaramath:
> this implementation is ok. But considering that the print url is not used by the UI, it would perhaps have been better the backend not passing the url. Instead the backend could just have passed a boolean property called 'isPrintEnabled'.

Done.

veeresha-hmcts:
Verified and ready for merge.

PR #406, opened 2019-03-14
More appinsight cleanup

dwaynebailey:
Remove last of the trackDependency markers

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/lib/AI-Agent.xml
+++ b/lib/AI-Agent.xml
@@ -54,5 +54,12 @@
             <Method name="getWizardPageCollection"/>
             <Method name="getCaseTabCollection"/>
         </Class>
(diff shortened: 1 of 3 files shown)


lib/AI-Agent.xml
```diff
@@ -54,5 +54,15 @@
             <Method name="getWizardPageCollection"/>
             <Method name="getCaseTabCollection"/>
         </Class>
+        <Class name="uk.gov.hmcts.ccd.data.draft.DefaultDraftGateway">
```
mario-paniccia: Dwayne I've noticed we miss a very important class in this file: HttpUIDefinitionGateway
Can we add that please? I'm not sure whether that has already been automatically instrumented since it's http or we must add it explicitly in this file

PR #407, opened 2019-03-14
Treat empty string as null for security classifications

dwaynebailey:
### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationUtils.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationUtils.java
@@ -27,7 +27,7 @@ public class SecurityClassificationUtils {
     }
 
     public static Optional<SecurityClassification> getSecurityClassification(JsonNode dataNode) {
-        if (dataNode == null || dataNode.isNull()) {
+        if (dataNode == null || dataNode.isNull() || dataNode == "") {
             return Optional.empty();
         }
         SecurityClassification securityClassification;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationUtils.java
```diff
@@ -27,7 +27,7 @@ private SecurityClassificationUtils() {
     }
 
     public static Optional<SecurityClassification> getSecurityClassification(JsonNode dataNode) {
-        if (dataNode == null || dataNode.isNull()) {
+        if (dataNode == null || dataNode.isNull() || dataNode == "") {
             return Optional.empty();
```
mario-paniccia: try this instead:
```
public static Optional<SecurityClassification> getSecurityClassification(JsonNode dataNode) {
        if (dataNode == null || dataNode.isNull()) {
            return Optional.empty();
        }
        if (dataNode.textValue().isEmpty()) {
            return Optional.empty();
        }
        SecurityClassification securityClassification;
        try {
            securityClassification = valueOf(dataNode.textValue());
        }
        catch (IllegalArgumentException e) {
            LOG.error("Unable to parse security classification for {}", dataNode, e);
            return Optional.empty();
        }
        return Optional.of(securityClassification);
    }
```

src/main/java/uk/gov/hmcts/ccd/domain/service/common/SecurityClassificationUtils.java
```diff
@@ -27,7 +27,7 @@ private SecurityClassificationUtils() {
     }
```
mario-paniccia: can you add a test please?

PR #419, opened 2019-03-18
added caching for getVersion and fixed name of the TTL app param

MSancaktutar:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/ccd-data-store-api/values.template.yaml
+++ b/charts/ccd-data-store-api/values.template.yaml
@@ -35,7 +35,8 @@ java:
 
     DATA_STORE_S2S_AUTHORISED_SERVICES: ccd_data,ccd_gw,ccd_ps,probate_backend,divorce_ccd_submission,sscs,sscs_bulkscan,cmc,cmc_claim_store,jui_webapp,pui_webapp,bulk_scan_orchestrator,fpl_case_service,iac,finrem_ccd_data_migrator
 
-    DEFINITION_CACHE_TTL_SEC: 259200
+    DEFINITION_CACHE_MAX_IDLE_SEC: 259200
+    DEFINITION_CACHE_TTL_SEC: 30
     DEFINITION_CACHE_MAX_SIZE: 5000
     DEFINITION_CACHE_EVICTION_POLICY: NONE
 
--- a/charts/ccd-data-store-api/values.template.yaml
+++ b/charts/ccd-data-store-api/values.template.yaml
@@ -36,7 +36,7 @@ java:
     DATA_STORE_S2S_AUTHORISED_SERVICES: ccd_data,ccd_gw,ccd_ps,probate_backend,divorce_ccd_submission,sscs,sscs_bulkscan,cmc,cmc_claim_store,jui_webapp,pui_webapp,bulk_scan_orchestrator,fpl_case_service,iac,finrem_ccd_data_migrator
 
     DEFINITION_CACHE_MAX_IDLE_SEC: 259200
-    DEFINITION_CACHE_TTL_SEC: 30
+    DEFINITION_CACHE_LATEST_VERSION_TTL_SEC: 30
     DEFINITION_CACHE_MAX_SIZE: 5000
     DEFINITION_CACHE_EVICTION_POLICY: NONE
 
--- a/charts/ccd-data-store-api/values.template.yaml
+++ b/charts/ccd-data-store-api/values.template.yaml
@@ -36,7 +36,7 @@ java:
     DATA_STORE_S2S_AUTHORISED_SERVICES: ccd_data,ccd_gw,ccd_ps,probate_backend,divorce_ccd_submission,sscs,sscs_bulkscan,cmc,cmc_claim_store,jui_webapp,pui_webapp,bulk_scan_orchestrator,fpl_case_service,iac,finrem_ccd_data_migrator
 
     DEFINITION_CACHE_MAX_IDLE_SEC: 259200
-    DEFINITION_CACHE_LATEST_VERSION_TTL_SEC: 30
+    DEFINITION_CACHE_LATEST_VERSION_TTL_SEC: 1
     DEFINITION_CACHE_MAX_SIZE: 5000
     DEFINITION_CACHE_EVICTION_POLICY: NONE
 
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -65,7 +65,7 @@ variable "definition_cache_max_idle_sec" {
 
 variable "definition_latest_version_ttl_sec" {
   type = "string"
-  default = "0"
+  default = "30"
 }
 
 variable "definition_cache_max_size" {
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -63,7 +63,7 @@ variable "definition_cache_max_idle_sec" {
   default = "259200"
 }
 
-variable "definition_latest_version_ttl_sec" {
+variable "definition_cache_latest_version_ttl_sec" {
   type = "string"
   default = "1"
 }
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -58,9 +58,14 @@ variable "jenkins_AAD_objectId" {
   description                 = "(Required) The Azure AD object ID of a user, service principal or security group in the Azure Active Directory tenant for the vault. The object ID must be unique for the list of access policies."
 }
 
-variable "definition_cache_ttl_sec" {
+variable "definition_cache_max_idle_sec" {
   type = "string"
   default = "259200"
+}
+
+variable "definition_ttl_sec" {
+  type = "string"
+  default = "30"
 }
 
 variable "definition_cache_max_size" {
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -65,7 +65,7 @@ variable "definition_cache_max_idle_sec" {
 
 variable "definition_latest_version_ttl_sec" {
   type = "string"
-  default = "30"
+  default = "1"
 }
 
 variable "definition_cache_max_size" {
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -65,7 +65,7 @@ variable "definition_cache_max_idle_sec" {
 
 variable "definition_latest_version_ttl_sec" {
   type = "string"
-  default = "30"
+  default = "0"
 }
 
 variable "definition_cache_max_size" {
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -63,7 +63,7 @@ variable "definition_cache_max_idle_sec" {
   default = "259200"
 }
 
-variable "definition_ttl_sec" {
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
```diff
@@ -22,26 +25,34 @@ public Config hazelCastConfig() {
...
     }
 
-    private MapConfig newMapConfig(final String name, int definitionCacheTTL) {
-        return new MapConfig().setName(name)
+    private MapConfig newMapConfig(final String name, Integer definitionCacheMaxIdle, Integer definitionCacheTTL) {
+        MapConfig mapConfig = new MapConfig().setName(name)
                 .setMaxSizeConfig(new MaxSizeConfig(applicationParams.getDefinitionCacheMaxSize(), MaxSizeConfig.MaxSizePolicy.PER_NODE))
                 .setEvictionPolicy(applicationParams.getDefinitionCacheEvictionPolicy())
-                .setMaxIdleSeconds(definitionCacheTTL);
+                .setMaxIdleSeconds(definitionCacheMaxIdle);
```
mario-paniccia: I think if we use setTimeToLiveSeconds we can/should omit setMaxIdleSeconds, because the 2 serve the same purpose

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
```diff
@@ -22,26 +25,34 @@ public Config hazelCastConfig() {
...
+        Optional.ofNullable(definitionCacheTTL).ifPresent(value -> mapConfig.setTimeToLiveSeconds(value));
```
mario-paniccia: would it be more self descriptive to have the method argument be Optional<Integer> ?

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
```diff
@@ -22,26 +25,34 @@ public Config hazelCastConfig() {
...
+    private void configCaches(int definitionCacheMaxIdle, int definitionCacheTTL, Config config) {
+        config.addMapConfig(newMapConfig("caseTypeDefinitionsCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("workBasketResultCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("searchResultCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("searchInputDefinitionCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("workbasketInputDefinitionCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("caseTabCollectionCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("wizardPageCollectionCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("userRolesCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfig("caseTypeDefinitionLatestVersionCache", definitionCacheMaxIdle, definitionCacheTTL));
```
mario-paniccia: perhaps would be better to have in this case the property called `getVersionCacheTtl`? We can leave also `definitionCacheTTL`. We could have both but definitely we need to be able to control the ttl for the getVersion separately from all the other TTLs

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -148,7 +148,12 @@ public UserRole getUserRoleClassifications(String userRole) {
     }
 
     @Override
+    @Cacheable("caseTypeDefinitionLatestVersionCache")
```
mario-paniccia: not 100% sure about it, but if you put the Cacheable directly on the original method without adding this level of indirection it should still work. Needs to be tested though

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
MSancaktutar: Could this rather stay as it is now? This is because the logic around the 2 parameters are already managed internally in the cache implementation, and if there are 2 facts that apply to the cache, better to provide them and expect the consistent behaviour from the cache design and impl.

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
MSancaktutar: Makes full sense. done.

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
MSancaktutar: Makes sense. Refactored as agreed that the latest version ttl is intended to apply to a few items. Please check.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
MSancaktutar: It does work, indeed. The new integration test introduced verifies the cache behaviour.

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
mario-paniccia: I think it'll just make things confusing if we keep it both. I don't see the reason. We want the behaviour provided by the ttl and we want to make sure instead we don't get the behaviour of the max idle. So why setting it in the first place? Even if the cache is able to deal with it anyway.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
mario-paniccia: in that case can we remove 
```
public CaseTypeDefinitionVersion getLatestVersion(String caseTypeId) {
        return doGetLatestVersion(caseTypeId);
    }
```
and put the annotation directly on `getLatestVersion`?

src/main/resources/application.properties
```diff
@@ -26,8 +26,11 @@ spring.liquibase.change-log=classpath:/db/changelog/db.changelog-master.xml
 
 
 #definitions cache parameters
+#max idle time of cache map entry
+definition.cache.max-idle.secs=${DEFINITION_CACHE_MAX_IDLE_SEC:259200}
+
 #ttl of cache map entry
```
mario-paniccia: can you update this comment please

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
MSancaktutar: Well, I was thinking to design map configs like this: All our definition cache maps have at least  the max-idle defined. And they have an optional value of ttl defined as well. So, max-idle would apply to everything for obvious reason to evict items of no use for a long time, and a particular ttl, when defined, to evict when an items ages enough. This would mean a consistent use of all available parameters for all map configs, and we won't need to remember what we fed into which. Basically, if something is available as a param, it becomes part of the map config it applies to. In this case, I was suggesting max idle to apply to everything, even if the internal cache design/impl makes it logically obsolete.

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
MSancaktutar: Applied. Could you please check?

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
```diff
@@ -22,26 +23,35 @@ public Config hazelCastConfig() {
...
+        config.addMapConfig(newMapConfigWithMaxIdle("wizardPageCollectionCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfigWithMaxIdle("userRolesCache", definitionCacheMaxIdle));
+        config.addMapConfig(newMapConfigWithTtl("caseTypeDefinitionLatestVersionCache", latestVersionTTL));  
+    }
+
+    private MapConfig newMapConfigWithMaxIdle(final String name, final Integer maxIdle) {
+        return newMapConfig(name).setMaxIdleSeconds(maxIdle);
+    }
+
+    private MapConfig newMapConfigWithTtl(final String name, final Integer ttl) {
```
gbenadikar: 👍

infrastructure/variables.tf
```diff
@@ -58,11 +58,16 @@ variable "jenkins_AAD_objectId" {
...
 
-variable "definition_cache_ttl_sec" {
+variable "definition_cache_max_idle_sec" {
   type = "string"
   default = "259200"
 }
 
+variable "definition_latest_version_ttl_sec" {
+  type = "string"
+  default = "0"
```
mario-paniccia: Looking at the documentation it says: 
```Its default value is 0, which means infinite (no expiration and eviction)```
This means by default now we are caching forever, which is not what we want

File: charts/ccd-data-store-api/values.template.yaml [hunk 8]
```diff
@@ -35,7 +35,8 @@ java:
 
     DATA_STORE_S2S_AUTHORISED_SERVICES: ccd_data,ccd_gw,ccd_ps,probate_backend,divorce_ccd_submission,sscs,sscs_bulkscan,cmc,cmc_claim_store,jui_webapp,pui_webapp,bulk_scan_orchestrator,fpl_case_service,iac,finrem_ccd_data_migrator
 
-    DEFINITION_CACHE_TTL_SEC: 259200
+    DEFINITION_CACHE_MAX_IDLE_SEC: 259200
+    DEFINITION_CACHE_LATEST_VERSION_TTL_SEC: 30
```
mario-paniccia: this file is used by AKS, so I think at the moment is Preview and in the end will be applied to AAT as well. So I think this should be 1 sec

infrastructure/variables.tf
```diff
@@ -58,11 +58,16 @@ variable "jenkins_AAD_objectId" {
...
 
-variable "definition_cache_ttl_sec" {
+variable "definition_cache_max_idle_sec" {
   type = "string"
   default = "259200"
 }
 
+variable "definition_latest_version_ttl_sec" {
+  type = "string"
+  default = "30"
```
mario-paniccia: let's use 1 as default, to be as close as possible to the current behaviour by default (no caching)

infrastructure/variables.tf
MSancaktutar: Changed to default 30.

charts/ccd-data-store-api/values.template.yaml
MSancaktutar: Changed to 1 for AAT. Thank you!

infrastructure/variables.tf
MSancaktutar: Done so.

src/main/resources/application.properties
MSancaktutar: Done.

PR #420, opened 2019-03-18
RDM-4369: Support reserved characters in userId email address

Daniel-Lam:
https://tools.hmcts.net/jira/browse/RDM-4369.

### Change description ###
Ensure `GET` request to `/user-profile/users` endpoint uses a fully encoded URL, to allow reserved characters in the `uid` query parameter.

```
[x] Yes
[ ] No
```

**Note:** PR https://github.com/hmcts/ccd-user-profile-api/pull/173 must be merged to `master` *before* this change; User Profile API needs to be able to decode the encoded email address passed in the `uid` query parameter, as a result of this PR.


src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -93,11 +95,16 @@ public UserDefault getUserDefaultSettings(final String userId) {
             final HttpEntity requestEntity = new HttpEntity(securityUtils.authorizationHeaders());
             final Map<String, String> queryParams = new HashMap<>();
             queryParams.put("uid", userId);
+            final DefaultUriBuilderFactory builderFactory = new DefaultUriBuilderFactory();
+            // The EncodingMode.VALUES_ONLY mode ensures the query string variables are encoded
+            builderFactory.setEncodingMode(EncodingMode.VALUES_ONLY);
+            restTemplate.setUriTemplateHandler(builderFactory);
```
gbenadikar: May be the right place to set any properties on restTemplate is during the bean creation in RestTemplateConfiguration as restTemplate is an autowired bean here.  

Alternatively, to encode the user id, you could use the static method ApplicationParams.encodeBase64 - safer and isolated change. Setting properties on restTemplate may need regression testing on all endpoints where restTemplate is used.

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -93,11 +95,16 @@ public UserDefault getUserDefaultSettings(final String userId) {
...
             return restTemplate.exchange(applicationParams.userDefaultSettingsURL(),
                 HttpMethod.GET, requestEntity, UserDefault.class, queryParams).getBody();
-        } catch (HttpStatusCodeException e) {
+        } catch (RestClientResponseException e) {
```
gbenadikar: 👍

src/test/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepositoryTest.java
```diff
@@ -160,6 +177,69 @@ void shouldExcludeOtherRoles() {
         }
     }
 
+    @Nested
+    @DisplayName("getUserDefaultSettings()")
+    class GetUserDefaultSettings {
+
+        @Test
+        @DisplayName("should return the User Profile defaults for the user")
+        void shouldReturnUserProfileDefaultsForUser() {
```
gbenadikar: 👍

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
Daniel-Lam: Ideally, the query parameters just need `UTF-8` encoding, so `Base64` might be a bit overkill. Setting the encoding mode on `restTemplate` here - wouldn't this affect only the instance local to this class? I think it's quite safe, given `EncodingMode.VALUES_ONLY` targets only URI variables and not the whole URI template.

gbenadikar: @Daniel-Lam I thought restTemplate is an autowired singleton bean and not a local variable, I may be wrong. There is another static method ApplicationParams.encode (just need to make it public), that encodes to UTF-8. Whatever you think is right, go for it 🙂

Daniel-Lam: @gbenadikar - I think you're correct, actually... it is an autowired singleton. There will be a future ticket to ensure _all_ requests using `RestTemplate` encode their URI variables, but for now I think I'll make `ApplicationParams.encode()` public and use that instead.

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -93,8 +96,10 @@ public UserDefault getUserDefaultSettings(final String userId) {
             final HttpEntity requestEntity = new HttpEntity(securityUtils.authorizationHeaders());
             final Map<String, String> queryParams = new HashMap<>();
             queryParams.put("uid", ApplicationParams.encode(userId));
-            return restTemplate.exchange(applicationParams.userDefaultSettingsURL(),
-                HttpMethod.GET, requestEntity, UserDefault.class, queryParams).getBody();
+            final String encodedUrl = UriComponentsBuilder.fromHttpUrl(applicationParams.userDefaultSettingsURL())
+                .buildAndExpand(queryParams).toUriString();
+            return restTemplate.exchange(new URI(encodedUrl), HttpMethod.GET, requestEntity, UserDefault.class)
```
gbenadikar: 👍

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -92,17 +95,22 @@ public UserDefault getUserDefaultSettings(final String userId) {
...
-        } catch (HttpStatusCodeException e) {
+            queryParams.put("uid", ApplicationParams.encode(userId));
+            final String encodedUrl = UriComponentsBuilder.fromHttpUrl(applicationParams.userDefaultSettingsURL())
+                .buildAndExpand(queryParams).toUriString();
+            return restTemplate.exchange(new URI(encodedUrl), HttpMethod.GET, requestEntity, UserDefault.class)
+                .getBody();
+        } catch (RestClientResponseException e) {
             LOG.error("Failed to retrieve user profile", e);
-            final List<String> headerMessages = e.getResponseHeaders().get("Message");
+            final List<String> headerMessages = Optional.ofNullable(e.getResponseHeaders())
```
mario-paniccia: @Daniel-Lam can we please verify that in case a user does not exist, we don't throw an error? We should instead just return 404 non found
Can we add a test for this case if not there yet please?

Daniel-Lam: @mario-paniccia - current behaviour is that User Profile API returns a `400 Bad Request` when the user does not exist; Data Store takes this response and throws a `BadRequestException` with that `400` error. Do we want to change this behaviour, then?

Daniel-Lam: If so, which side should change: Data Store or User Profile API?

Daniel-Lam: Response from @mario-paniccia on Slack earlier today:
> Mario 7:51 AM
> morning Daniel. I'm thinking better leave it as it is now. Because I've realised it might somehow impact the UI as well if we change this now. So it's quite a big work. Not worth it. Cheers though

PR #424, opened 2019-03-20
Rdm 4394 cache find jurisdictions

MSancaktutar:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -178,7 +178,12 @@ public CaseType getCaseType(int version, String caseTypeId) {
     }
 
     @Override
+    @Cacheable("jurisdictionListCache")
     public List<Jurisdiction> getJurisdictions(List<String> ids) {
```
gbenadikar: may be worth sorting the ids before passing to this method, just to make sure it generates the same cache key for the same ids that are unordered e.g. {"TEST1", "TEST2"} and  {"TEST2", "TEST1"}

MSancaktutar: good one worth doing.

src/main/java/uk/gov/hmcts/ccd/data/IdListCacheKeyGenerator.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cache.interceptor.KeyGenerator;
+import org.springframework.stereotype.Component;
+
+@Component
+@Qualifier("idListCacheKeyGenerator")
+public class IdListCacheKeyGenerator implements KeyGenerator {
+
+    @Override
+    public Object generate(Object target, Method method, Object... params) {
```
gbenadikar: LGTM

infrastructure/prod.tfvars
```diff
@@ -18,4 +18,5 @@ http_client_max_total = 200
 http_client_max_client_per_route = 40
 #http_client_validate_after_inactivity = 0
 
-definition_latest_version_ttl_sec = 5
\ No newline at end of file
+definition_latest_version_ttl_sec = 5
+definition_jurisdiction_ttl_sec = 5
```
mario-paniccia: considering this operation is called only once when a user logs in, I doubt 5 sec ttl will have any effect on performance. It'll help only if 2 users that retrieve the same jurisdictions login max 5 sec apart from each other

infrastructure/demo.tfvars
```diff
@@ -11,4 +11,5 @@ data_store_max_pool_size = 48
 database_sku_name = "GP_Gen5_8"
 database_sku_capacity = "8"
 
-definition_latest_version_ttl_sec = 30
\ No newline at end of file
+definition_latest_version_ttl_sec = 30
+definition_jurisdiction_ttl_sec = 30
```
mario-paniccia: considering this operation is called only once when a user logs in, I doubt 30 sec ttl will have any effect on performance. It'll help only if 2 users that retrieve the same jurisdictions login max 30 sec apart from each other

src/main/resources/application.properties
```diff
@@ -30,7 +30,10 @@ spring.liquibase.change-log=classpath:/db/changelog/db.changelog-master.xml
 definition.cache.max-idle.secs=${DEFINITION_CACHE_MAX_IDLE_SEC:259200}
 
 #ttl for the latest version objects cached
-definition.cache.latest-version-ttl.secs=${DEFINITION_CACHE_LATEST_VERSION_TTL_SEC:30}
+definition.cache.latest-version-ttl=${DEFINITION_CACHE_LATEST_VERSION_TTL_SEC:30}
```
mario-paniccia: I think we should keep the `.sec` suffix to avoid confusion

src/main/java/uk/gov/hmcts/ccd/data/IdListCacheKeyGenerator.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+
+import org.apache.tomcat.util.buf.StringUtils;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cache.interceptor.KeyGenerator;
+import org.springframework.stereotype.Component;
+
+@Component
+@Qualifier("idListCacheKeyGenerator")
+public class IdListCacheKeyGenerator implements KeyGenerator {
```
mario-paniccia: Not for now. But I think a better approach in the future would be rather than caching the whole list of jurisdiction, to refactor the code to fetch 1 jurisdiction at the time so that we can cache the single Jurisdictions by jurisdiction id.
This will make better usage of the cache. Less cached data, because we can the single jurs rather that all the possible combinations of jur. And more cache hits.
But then will increase the number of requests to the backend in normal situations when cache is not used. So there are pros and cons. But worth pointing this out.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
MSancaktutar: Well, it later on turned out that doing this is causing functional failures. The reason is the return value of the method is also a List and if we treat the id lists as sets, we will have to treat the Jurisdiction lists also as sets. Otherwise, same set of Jurisdictions returned in different order is breaking the code. That's why I cancelled the implementation of this comment, or otherwise it would take further refactorings in other places of the code.

infrastructure/prod.tfvars
MSancaktutar: We can change this number, but the caching of the Jurisdiction list is already serving for purpose of that particular use. Approximately 3K calls per day, this optimisation can still be useful, if not sensible across the entire system.

infrastructure/demo.tfvars
MSancaktutar: We can change this number, but the caching of the Jurisdiction list is already serving for purpose of that particular use. Approximately 3K calls per day, this optimisation can still be useful, if not sensible across the entire system.

src/main/resources/application.properties
MSancaktutar: We can later on refactor to include the unit of measurement in the value, like 5sec. But for now, let the unit of measurement stay in the describing the property, rather than in the property name itself.

src/main/java/uk/gov/hmcts/ccd/data/IdListCacheKeyGenerator.java
MSancaktutar: Ok, removing this.

infrastructure/prod.tfvars
```diff
@@ -19,3 +19,4 @@ http_client_max_client_per_route = 40
 #http_client_validate_after_inactivity = 0
 
 definition_cache_latest_version_ttl_sec = 30
+definition_cache_jurisdiction_ttl_sec = 5
```
mario-paniccia: Can we increase both `definition_cache_latest_version_ttl_sec` and `definition_cache_jurisdiction_ttl_sec` to 60 sec on Prod? We must notify services to just wait 1 min after a definition was uploaded. That is an acceptable compromise to get an increase in performance.

infrastructure/sprod.tfvars
```diff
@@ -16,6 +16,6 @@ http_client_max_client_per_route = 40
 #http_client_validate_after_inactivity = 0
 
 definition_cache_latest_version_ttl_sec = 5
+definition_cache_jurisdiction_ttl_sec = 5
```
mario-paniccia: Both of `definition_cache_latest_version_ttl_sec` and `definition_cache_jurisdiction_ttl_sec` must match the values on Prod. Same for AAT

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
```diff
@@ -77,7 +79,14 @@ public CaseTypeDefinitionVersion getLatestVersion(String caseTypeId) {
 
     @Override
     public List<Jurisdiction> getJurisdictions(List<String> ids) {
-        return this.caseDefinitionRepository.getJurisdictions(ids);
```
mario-paniccia: prefer: 
```
        return ids.stream().map(this::getJurisdiction).collect(Collectors.toList());
```

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -172,31 +181,39 @@ public CaseTypeDefinitionVersion doGetLatestVersion(String caseTypeId) {
...
-    @Cacheable("caseTypeDefinitionsCache")
-    public CaseType getCaseType(int version, String caseTypeId) {
-        return this.getCaseType(caseTypeId);
+    public List<Jurisdiction> getJurisdictions(List<String> ids) {
+        ArrayList<Jurisdiction> jurisdictions = new ArrayList<>(ids.size());
+        ids.forEach(id -> jurisdictions.add(getJurisdiction(id)));
+        return jurisdictions;
     }
 
+    @Cacheable(value = "jurisdictionListCache")
```
mario-paniccia: shall we rename `jurisdictionListCache` to `jurisdictionCache`?

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -172,31 +181,39 @@ public CaseTypeDefinitionVersion doGetLatestVersion(String caseTypeId) {
     }
 
     @Override
-    @Cacheable("caseTypeDefinitionsCache")
-    public CaseType getCaseType(int version, String caseTypeId) {
-        return this.getCaseType(caseTypeId);
+    public List<Jurisdiction> getJurisdictions(List<String> ids) {
+        ArrayList<Jurisdiction> jurisdictions = new ArrayList<>(ids.size());
```
mario-paniccia: this method can potentially be called bypassing caching IF this repository is injected without wrapping it with the cached repository. In that case this implementation has got performance issues. Because it makes many single calls rather than one. I would suggest we reimplement this to make 1 http call only that fetches all the jurisdictions. And in the cached repository we keep the current implementation that loops and call the getJurisdiction and triggers that caching

infrastructure/demo.tfvars
```diff
@@ -11,4 +11,5 @@ data_store_max_pool_size = 48
 database_sku_name = "GP_Gen5_8"
 database_sku_capacity = "8"
 
-definition_cache_latest_version_ttl_sec = 30
\ No newline at end of file
+definition_cache_latest_version_ttl_sec = 30
```
mario-paniccia: can we have 2 min on Demo?

infrastructure/prod.tfvars
MSancaktutar: Done.

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
MSancaktutar: Done. Thank you.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
MSancaktutar: Already was done. :)

infrastructure/demo.tfvars
MSancaktutar: Done for both jurisdiction and last version caches on both prod and demo.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
MSancaktutar: Done as discussed. Thank you.

MSancaktutar:
Closing without merge, as there will be another PR for this work.

PR #426, opened 2019-03-21
Rmd 4402 cache pre-warm

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4402

### Change description ###

Disabled by default and with fail safe logic so unsuccessful warm up (due to auth or def store calls) does not fail the app start up.

There is a related PR on def store which should be merged first:
https://github.com/hmcts/ccd-definition-store-api/pull/356

No docker changes to follow as although this logic is default it is fail safe and does not affect app startup.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
```diff
@@ -37,11 +81,43 @@ private void configCaches(int definitionCacheTTL, Config config) {
         config.addMapConfig(newMapConfig("userRolesCache", definitionCacheTTL));
     }
 
+    private void warmUp(final HazelcastInstance hazelcastInstance) {
+        IMap<String, CaseType> caseTypeDefinitionsCache = hazelcastInstance.getMap("caseTypeDefinitionsCache");
```
mario-paniccia: I don't think this is going to work. We are trying to fetch from def store all the case type definitions at once. Consider that at the moment retrieving 1 definition
is a very expensive operation. Which takes about 10 seconds. I'm worried this single call will kill Def Store. 
Also with this approach we introduced a risky endpoint in Def Store which fetches all the definitions. We already had that endpoint in the past,
and I removed it to make sure no one attempts to do something like that. Which can kill Def Store db.
I think a better approach would be to retrieve from the Def Store only the ids of the case types. Then in Data Store we can do a looping logic which 
retrieves the case types one by one. 
If you do this, you can even avoid adding all this logic here in this class, you can just call the existing getCaseType, which already does caching automatically.
We could also customise the fetching to say for example wait 1 second between each single case type fetch. To avoid overloading the Def Store.

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
```diff
@@ -37,11 +81,43 @@ private void configCaches(int definitionCacheTTL, Config config) {
...
+        List<CaseType> allCaseTypes = getAllCaseTypes();
+        allCaseTypes.stream().forEach(caseType -> {
+            caseTypeDefinitionsCache.put(caseType.getId(), caseType);
```
mario-paniccia: Can we introduce a CachePreWarmingService Interface with on implementation for the moment DefaultCachePreWarmingServiceImpl please. So we have all the logic
for the pre warming nicely encapsulated

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
andrewfolga: Sure will change.

mario-paniccia: cheers 👍

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
andrewfolga: Yes I will introduce DefaultCacheWarmUpService as that name sounds better to me.

src/main/java/uk/gov/hmcts/ccd/CachingConfiguration.java
andrewfolga: I cannot call the existing getCaseType(id) as I do not have request scope here. Remember this is an app startup so I have to build the headers for the cache warm up user manually.

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
```diff
@@ -0,0 +1,63 @@
+package uk.gov.hmcts.ccd;
```
mario-paniccia: can we move this into a `cache` package?

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.List;
+
+@Service
+@Qualifier(DefaultCacheWarmUpService.QUALIFIER)
+public class DefaultCacheWarmUpService implements CacheWarmUpService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CachingConfiguration.class);
+    public static final String QUALIFIER = "default";
+    private static final long TWO_SECONDS = 2000l;
```
mario-paniccia: can we have the sleep interval configurable per env please

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
```diff
@@ -0,0 +1,63 @@
...
+        this.idamHelper = idamHelper;
+    }
+
+    @Override
+    public void warmUp() {
+        try {
+            HttpHeaders httpHeaders = authorizationHeaders();
+            List<String> caseTypesReferences = caseDefinitionRepository.getCaseTypesReferences(httpHeaders);
+            for (String reference : caseTypesReferences) {
+                Thread.sleep(TWO_SECONDS);
```
mario-paniccia: I'm always a bit concerned on using directly Thread sleep. Can we see if there's any more modern approach to execute operations with sleep time in between? perhaps some library, spring support, etc...

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -77,11 +78,37 @@ public DefaultCaseDefinitionRepository(final ApplicationParams applicationParams
         }
     }
 
+    @Override
+    public List<String> getCaseTypesReferences() {
+        return getCaseTypesReferences(null);
+    }
+
+    public List<String> getCaseTypesReferences(final HttpHeaders httpHeaders) {
```
mario-paniccia: Can we declare `httpHeaders` Optional? rather than passing a null?

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -77,11 +78,37 @@ public DefaultCaseDefinitionRepository(final ApplicationParams applicationParams
...
+            }
+        }
+    }
+
     @Override
     public CaseType getCaseType(final String caseTypeId) {
+        return getCaseType(caseTypeId, null);
+    }
+
+    public CaseType getCaseType(final String caseTypeId, final HttpHeaders httpHeaders) {
```
mario-paniccia: Can we declare httpHeaders Optional? rather than passing a null?

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
mario-paniccia: Just come to my mind that we will probably need this Pre Warming to be an asynchronous activity. As it's now, if it takes 1 min to load all the case type definition,
we will be slowing down Data Store startup time by 1 min?

src/main/resources/application.properties
```diff
@@ -137,3 +137,11 @@ management.health.elasticsearch.enabled=false
 management.health.status.order=DOWN, OUT_OF_SERVICE, UNKNOWN, UP, OUT_OF_SYNC, PROBLEM, COULD_NOT_CONNECT
 
 spring.banner.location=classpath:ccd-data-store-banner.txt
+
+# OAuth2
+
+ccd.oauth2.client_id: ${OAUTH2_CLIENT_ID:ccd_gateway}
+ccd.oauth2.client_secret: ${OAUTH2_CLIENT_SECRET:OOOOOOOOOOOOOOOO}
+ccd.oauth2.redirect_url: ${OAUTH2_REDIRECT_URL:http://localhost:3451/oauth2redirect}
+ccd.cache.warm.up.email: ${CCD_CACHE_WARM_UP_EMAIL:ccdcachewarmer@gmail.com}
```
dwaynebailey: Please no gmail addresses.  See https://tools.hmcts.net/confluence/display/RCCD/Automation+and+CCD

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
andrewfolga: Optional as method parameters? It is advised to use it as a return type only. Please see:
http://dolszewski.com/java/java-optional-parameters/
3.3. Why not Optional as method argument?

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
andrewfolga: Optional as method parameters? It is advised to use it as a return type only. Please see:
http://dolszewski.com/java/java-optional-parameters/
3.3. Why not Optional as method argument?

src/main/resources/application.properties
andrewfolga: Thanks. I will address it with @hemantt as he was the creator.

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
andrewfolga: Done.

andrewfolga: Do you want the `CachingConfiguration` to go to `cache` package too?

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
andrewfolga: What name would you suggest? Can we go with cacheWarmUpDelay?

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
mario-paniccia: yes would be a good move 👍

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
mario-paniccia: `cache.warmup.sleep.time` - cacheWarmUpSleepTime ?

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
andrewfolga: Yes I will look at the async way.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
mario-paniccia: As the guys says, this is very controversial. I see some of his points. You can leave it as you prefer

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
mario-paniccia: 👍

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
andrewfolga: Cool sounds good to me.

src/main/java/uk/gov/hmcts/ccd/DefaultCacheWarmUpService.java
andrewfolga: Cool

src/main/java/uk/gov/hmcts/ccd/cache/CachingConfiguration.java
```diff
@@ -35,13 +46,14 @@ private void configCaches(int definitionCacheTTL, Config config) {
         config.addMapConfig(newMapConfig("caseTabCollectionCache", definitionCacheTTL));
         config.addMapConfig(newMapConfig("wizardPageCollectionCache", definitionCacheTTL));
         config.addMapConfig(newMapConfig("userRolesCache", definitionCacheTTL));
+        cacheWarmUpService.warmUp();
```
mario-paniccia: the ticket also mentions to add the ability to enable/disable the pre warm up per environment. Can we add that please? 
I might be wrong but I don't see it. 
For local development better to have it disabled. Initially we will enable this on Demo only. If all looks good we'll then enable on AAT and Prod

andrewfolga: Is this because you do not want to hit def store db in some envs, correct? I would imagine for local dev it does not really matter. Remember it is fail safe so it won't stop the app starting up. But yes if that is comprehended and still the performance is a concern then I can add it per env.

andrewfolga: I have added the flag anyway. Better to have more control as you say 👍

src/main/resources/application.properties
```diff
@@ -137,3 +137,13 @@ management.health.elasticsearch.enabled=false
 management.health.status.order=DOWN, OUT_OF_SERVICE, UNKNOWN, UP, OUT_OF_SYNC, PROBLEM, COULD_NOT_CONNECT
 
 spring.banner.location=classpath:ccd-data-store-banner.txt
+
+# OAuth2
+
+ccd.oauth2.client_id: ${OAUTH2_CLIENT_ID:ccd_gateway}
+ccd.oauth2.client_secret: ${OAUTH2_CLIENT_SECRET:OOOOOOOOOOOOOOOO}
+ccd.oauth2.redirect_url: ${OAUTH2_REDIRECT_URL:http://localhost:3451/oauth2redirect}
+ccd.cache.warm.up.enabled: ${CCD_CACHE_WARM_UP_ENABLED:false}
```
mario-paniccia: where are we overriding this default value for Demo and AAT to be true?

Jenkinsfile_CNP
```diff
@@ -33,6 +32,13 @@ def secrets = [
         secret('ccd-private-cross-case-type-solicitor-password', 'CCD_PRIVATE_CROSS_CASE_TYPE_SOLICITOR_PASSWORD'),
         secret('ccd-restricted-cross-case-type-caseworker-email', 'CCD_RESTRICTED_CROSS_CASE_TYPE_CASEWORKER_EMAIL'),
         secret('ccd-restricted-cross-case-type-caseworker-password', 'CCD_RESTRICTED_CROSS_CASE_TYPE_CASEWORKER_PASSWORD'),
+        secret('ccd-cache-warm-up-email', 'CCD_CACHE_WARM_UP_EMAIL'),
+        secret('ccd-cache-warm-up-password', 'CCD_CACHE_WARM_UP_PASSWORD'),
+        secret('ccd-cache-warm-up-sleep-time-sec', 'CCD_CACHE_WARM_UP_SLEEP_TIME_SEC'),
```
mario-paniccia: rather than having `ccd-cache-warm-up-sleep-time-sec` and `ccd-cache-warm-up-enabled` declared as secrets,
can't we have them like terraform variables in the `main.tf` and overriding them per env? Like we do for example for `DEFINITION_CACHE_TTL_SEC`

src/main/java/uk/gov/hmcts/ccd/cache/DefaultCacheWarmUpService.java
```diff
@@ -0,0 +1,66 @@
...
+import uk.gov.hmcts.ccd.idam.AuthenticatedUser;
+import uk.gov.hmcts.ccd.idam.IdamHelper;
+import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+@Service
+@Qualifier(DefaultCacheWarmUpService.QUALIFIER)
+public class DefaultCacheWarmUpService implements CacheWarmUpService {
```
mario-paniccia: shall we add a test?

src/main/resources/application.properties
andrewfolga: Done.

Jenkinsfile_CNP
andrewfolga: Done

src/main/java/uk/gov/hmcts/ccd/cache/DefaultCacheWarmUpService.java
andrewfolga: Tests added.

src/main/resources/application.properties
andrewfolga: New email were created following the guidelines.

andrewfolga: However, @hemantt is not sure how he can confirm registration via email as he won't have access to that email account.

dwaynebailey: Fine with this being temporary as long as there is a ticket to resolve this.

PR #427, opened 2019-03-22
RDM-4363 The roles defined in CaseTypeTab worksheet are applied

tzarsmango:
The roles defined in CaseTypeTab worksheet are applied as a part of CRUD operations. The tab is filtered if a UserRole is defined in CaseTypeTab and the current user doesn't have that role. If no role is defined in that column it works as it used to do and that tab is visible for all logged in users.

https://tools.hmcts.net/jira/browse/RDM-4363

```
[] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/AuthorisedGetCaseViewOperation.java
```diff
@@ -6,6 +6,7 @@
 import static uk.gov.hmcts.ccd.domain.service.common.AccessControlService.CAN_READ;
 import static uk.gov.hmcts.ccd.domain.service.common.AccessControlService.CAN_UPDATE;
 
+import org.apache.commons.lang.StringUtils;
```
mario-paniccia: Fatih would it make sense to have a separate filter dedicated to role authorisation? for example RoleAuthorisedGetCaseViewOperation?
This class is mainly for CRUD, aren't we mixing separate things?

mario-paniccia: I bet this code for checking the role authorisation is potentially highly reusable. And could be used for the other thickets too related to role based authorisation.
What if for example we have CaseView implement an interface called `RoleBased'. The we could have the `RoleBasedAuthorisation` operate on generic entities of type `RoleBased'. Which provide a method `getRole()`

tzarsmango: Role Based Tabs is yet another CRUD implementation for Tabs. Since Tabs are meaningful only for Case Views, it's R of CRUD we have here. I assume the name Role based may be misleading since whole CRUD implementation is about role based access to jurisdiction, case type, state or case fields. 
The current implementation is at the point where we're filtering tabs, so I feel it's somewhat a relevant/good place. I would say we keep the current implementation, finish this task and look for an optimisation/refactoring on the oncoming tasks.

mario-paniccia: ok then. Let's keep an eye on avoiding potential duplication we might introduce in the next role related tickets tough

README.md
```diff
@@ -92,6 +92,25 @@ You don't need to migrate database manually since migrations are executed every
...
+export S2S_URL=http://localhost:4502
+export CCD_GW_SERVICE_NAME=ccd_gw
+export CCD_GW_SERVICE_SECRET=AAAAAAAAAAAAAAAC
+export CCD_CASEWORKER_AUTOTEST_EMAIL=auto.test.cnp@gmail.com
+export CCD_CASEWORKER_AUTOTEST_PASSWORD=Monday123
+```
+
+These tests also rely on the `CCD_CNP_27.xlsx` file to be already imported. The latest version of this file could be found 
+in [ccd-definition-store-api](https://github.com/hmcts/ccd-definition-store-api) project under the `aat/src/resource` folder. You may want to import it to your local installation 
+before trying to run the functional tests against yoru local environment.
```
andrewfolga: Small typo.

andrewfolga: 👍 for adding this

PR #429, opened 2019-03-26
Introduced configuration of the hazelcast HealthMonitor and JMX

mario-paniccia:
https://tools.hmcts.net/jira/browse/RDM-4447

```
[ ] Yes
[x ] No
```


infrastructure/aat.tfvars
```diff
@@ -13,3 +13,5 @@ database_sku_name = "GP_Gen5_8"
 database_sku_capacity = "8"
 
 http_client_read_timeout = "180000"
+
+hazelcast_health_monitoring_level = "NOISY"
```
dwaynebailey: Do we really want noisy? That will log all the time, not just when we're over a threshold.

infrastructure/variables.tf
```diff
@@ -73,6 +73,18 @@ variable "definition_cache_eviction_policy" {
   default = "NONE"
 }
 
+variable "hazelcast_health_monitoring_level" {
+  default = "SILENT"
+}
+
+variable "hazelcast_health_monitoring_delay_seconds" {
+  default = "3000"
```
dwaynebailey: 3000 second? 50 minutes.  Sounds like a missing 0 or something.

mario-paniccia: let's log every hour?

infrastructure/aat.tfvars
mario-paniccia: I need this to be noisy for now on AAT because can help investigate caching issues on AAT.

charts/ccd-data-store-api/values.template.yaml
```diff
@@ -40,6 +40,9 @@ java:
     DEFINITION_CACHE_TTL_SEC: 259200
     DEFINITION_CACHE_MAX_SIZE: 5000
     DEFINITION_CACHE_EVICTION_POLICY: NONE
+    HAZELCAST_HEALTH_MONITORING_LEVEL: OFF
+    HAZELCAST_HEALTH_MONITORING_DELAY_SECONDS: 3000
```
dwaynebailey: I think this one wants `3600`or `300` like the Terraform / default config

PR #430, opened 2019-03-27
RDM-3515 Alter code that discloses the callback url

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-3515

This PR fixes the the error message discloses callback URL security issue

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
@@ -55,8 +55,10 @@ public class CallbackService {
         }
         final CallbackRequest callbackRequest = new CallbackRequest(caseDetails, caseDetailsBefore, caseEvent.getId(), ignoreWarning);
         final Optional<ResponseEntity<CallbackResponse>> responseEntity = sendRequest(url, CallbackResponse.class, callbackRequest);
-        return responseEntity.map(re -> Optional.of(re.getBody())).orElseThrow(() ->
-            new CallbackException("Callback to service has been unsuccessful for event " + caseEvent.getId()));
+        return responseEntity.map(re -> Optional.of(re.getBody())).orElseThrow(() -> {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -55,7 +55,8 @@ public CallbackService(final SecurityUtils securityUtils,
         }
         final CallbackRequest callbackRequest = new CallbackRequest(caseDetails, caseDetailsBefore, caseEvent.getId(), ignoreWarning);
         final Optional<ResponseEntity<CallbackResponse>> responseEntity = sendRequest(url, CallbackResponse.class, callbackRequest);
-        return responseEntity.map(re -> Optional.of(re.getBody())).orElseThrow(() -> new CallbackException("Unsuccessful callback to " + url));
+        return responseEntity.map(re -> Optional.of(re.getBody())).orElseThrow(() ->
+            new CallbackException("Callback to service has been unsuccessful for event " + caseEvent.getId()));
```
mario-paniccia: can we define this error message in a constant field so we can reuse it below? Minor, ignore if you want

mario-paniccia: What about a LOG.WARN of `Unsuccessful callback to " + url`? it's ok to not expose it to the user, but the failing callback url is an important info we could log

mario-paniccia: Furthermore, maybe better to ask a BA what would be the best message to send to the end user? Perhaps something like 'Could not contact external service this event caseEvent.getId() depends on'?

tzarsmango: added a log.info line with url, caseType and event information. Updated the message with event name instead of the event Id. I assume the service name Probate, Divorce etc should be somehow clear to the user.

PR #449, opened 2019-04-11
RPE-1005 Fix AKS AAT

timja:
https://tools.hmcts.net/jira/browse/RPE-1005

### Change description ###
Add missing password definition to bootstrap.yaml

Rework files to use new values.yaml default and release name templating for postgres

```
[ ] Yes
[ ] No
```


charts/ccd-data-store-api/values.aat.template.yaml
```diff
@@ -1,10 +1,40 @@
...
-    CCD_DM_DOMAIN: ^https?://(?:api-gateway\\.preprod\\.dm\\.reform\\.hmcts\\.net|dm-store-aat\\.service\\.core-compute-aat\\.internal(?::\\d+)?)
-    #IDAM_USER_URL: https://preprod-idamapi.reform.hmcts.net:3511
+    DATA_STORE_DB_HOST: ccd-data-store-api-postgres-db-aat.postgres.database.azure.com
+    DATA_STORE_DB_USERNAME: ccd@ccd-data-store-api-postgres-db-aat
+    DATA_STORE_DB_NAME: ccd_data_store
+    DATA_STORE_DB_OPTIONS: "?stringtype=unspecified&sslmode=require"
+
+    IDAM_S2S_URL: http://rpe-service-auth-provider-java.software-engineering
+    DEFINITION_STORE_HOST: http://ccd-definition-store-api-java
+    USER_PROFILE_HOST: http://ccd-user-profile-api-java
```
adusumillipraveen: As these files are used for staging slot deployment, these URLs might dependant on how we deploy the staging slot in AAT ? Will the staging apps be in same namespace? I thought we would put them in a different namespace like "ccd-staging" ? So, may be using  http://ccd-user-profile-api-java.ccd is not a bad idea?

charts/ccd-data-store-api/values.preview.template.yaml
```diff
@@ -0,0 +1,5 @@
+java:
+  image: ${IMAGE_NAME}
+  ingressHost: ${SERVICE_FQDN}
+  postgresql:
```
adusumillipraveen: This should be at root level, outside java.

charts/ccd-data-store-api/values.yaml
```diff
@@ -0,0 +1,50 @@
+java:
+  image: hmcts.azurecr.io/hmcts/ccd-data-store-api:latest
+  applicationPort: 4452
+  applicationInsightsInstrumentKey: "fake-key"
```
adusumillipraveen: May be get rid of it as java app puts a fake one?

charts/ccd-data-store-api/values.preview.template.yaml
timja: why's that? I assumed since its a dependency of the java app it would go inside of it

charts/ccd-data-store-api/values.aat.template.yaml
timja: probably easiest to keep them in the same namespace so they use the same identity, I assume its safe to always specify the namespace so will add it in anyway

charts/ccd-data-store-api/values.preview.template.yaml
adusumillipraveen: https://github.com/helm/helm/issues/4490#issuecomment-469423429 . Limitation from Helm

PR #453, opened 2019-04-12
RDM-4456 Filter searchInput and workbasketInput per user role

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-4456
https://tools.hmcts.net/jira/browse/RDM-4465

Refactor search input fields and workbasket input fields
Filter searchInput and workbasketInput per user role
Alter tests and add tests for new requirements

This PR is dependent on https://github.com/hmcts/ccd-definition-store-api/pull/386

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[X] No
```

PR #457, opened 2019-04-15
RDM-3460 Search and Workbasket changes to support complex fields

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-3460

### Change description ###
Complex element *only* as list screen input.

```
[ ] Yes
[ X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseType.java
```diff
@@ -153,12 +152,17 @@ public void setSearchAliasFields(List<SearchAliasField> searchAliasFields) {
     }
 
     @JsonIgnore
-    public Optional<CaseField> getCaseField(String caseFieldId) {
-        return caseFields.stream().filter(caseField -> caseField.getId().equalsIgnoreCase(caseFieldId)).findFirst();
+    public CaseField getCaseField(String caseFieldId) {
+        return caseFields.stream().filter(caseField -> caseField.getId().equalsIgnoreCase(caseFieldId)).findFirst()
```
mario-paniccia: can I know what's the rationale for this change? I think the previous solution is cleaner. I don't see why we need to throw an Exception here

src/main/java/uk/gov/hmcts/ccd/domain/model/search/Field.java
```diff
@@ -5,6 +5,7 @@
 
 public class Field {
     private String id;
+    private String elementPath;
```
mario-paniccia: Do we need to add elementPath to Field? it's a quite specific property needed only for this particular use case. We already have added it to WorkbasketInputField and SearchInputField can't we use that property?

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultFindSearchInputOperation.java
```diff
@@ -51,19 +51,15 @@ private SearchInput toSearchInput(final SearchInputField in, final CaseType case
         final SearchInput result = new SearchInput();
         result.setLabel(in.getLabel());
         result.setOrder(in.getDisplayOrder());
+        CaseField caseField = caseType.getCaseField(in.getCaseFieldId());
```
mario-paniccia: you can use the original `getCaseField` method that returns Optional here and throw exception at this level

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
```diff
@@ -140,6 +144,46 @@ public void propagateACLsToNestedFields() {
...
+        if (StringUtils.isBlank(path)) {
+            return this.getFieldType();
+        } else {
+            CaseField caseFieldByPath = this.findNestedElementByPath(path);
+            return caseFieldByPath.getFieldType();
+        }
+    }
+
+    @JsonIgnore
+    public CaseField findNestedElementByPath(String path) {
```
mario-paniccia: can this be private? In that case we could remove the blank field validation because we already did it in `getFieldTypeByPath`

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
```diff
@@ -140,6 +144,46 @@ public void propagateACLsToNestedFields() {
         propagateACLsToNestedFields(this, this.accessControlLists);
     }
 
+    @JsonIgnore
+    public FieldType getFieldTypeByPath(String path) {
```
mario-paniccia: I might be wrong but such a method would be better placed in the FieldType class rather than CaseField. You could return an Optional<FieldType>. And the following check:
`if (StringUtils.isBlank(path)) {
            return this.getFieldType();
 }`
could be removed from here and placed in the `DefaultFindSearchInputOperation`

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
```diff
@@ -140,6 +144,46 @@ public void propagateACLsToNestedFields() {
...
+        }
+    }
+
+    @JsonIgnore
+    public CaseField findNestedElementByPath(String path) {
+        if (StringUtils.isBlank(path)) {
+            throw new RuntimeException(format("Invalid blank element path for field %s.", this.id));
+        }
+        if (this.getFieldType().getChildren().isEmpty()) {
+            throw new RuntimeException(format("CaseField %s has no nested elements.", this.id));
```
mario-paniccia: I would expect the endpoint to return a 400 Bad Request exception when this.getFieldType().getChildren().isEmpty(). Can we double check that a RuntimeException thrown here gets mapped by the endpoint to a 400 status code and not a 50X type of error?

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
```diff
@@ -140,6 +144,46 @@ public void propagateACLsToNestedFields() {
...
+
+        return reduce(this.getFieldType().getChildren(), pathElements);
+    }
+
+    @JsonIgnore
+    private CaseField reduce(List<CaseField> caseFields, List<String> pathElements) {
+        String head = pathElements.get(0);
+        if (pathElements.size() == 1) {
+            return caseFields.stream().filter(e -> e.getId().equals(head)).findFirst()
+                .orElseThrow(() -> new RuntimeException(format("Nested element not found for %s", head)));
```
mario-paniccia: can we return a more descriptive error message and also the `path` as part of the message?

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
```diff
@@ -140,6 +144,46 @@ public void propagateACLsToNestedFields() {
...
+        } else {
+            CaseField caseField = caseFields.stream().filter(e -> e.getId().equals(head)).findFirst()
```
mario-paniccia: this block of code is duplicated with the above. Can we remove this duplication?

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseType.java
rafalkalita: Changed back

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
rafalkalita: Done

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
rafalkalita: head is a part of the path. Changed the name to reflect it better.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
rafalkalita: Returning BadRequestException that is our exception mapped to 400

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
rafalkalita: Moved to CaseType. Now it is named getCaseFieldByPath(String caseFieldId, String path)

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseField.java
rafalkalita: Now used by CaseType getCaseFieldByPath(...) method

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/DefaultFindSearchInputOperation.java
rafalkalita: Changed as in above

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseType.java
```diff
@@ -153,12 +154,34 @@ public void setSearchAliasFields(List<SearchAliasField> searchAliasFields) {
...
+     * Gets a caseField by caseField id and specified path.
+     *
+     * @param caseFieldId CaseField id
+     * @param path Path to a nested CaseField
+     * @return Optional of a CaseField
+     */
     @JsonIgnore
-    public boolean isCaseFieldACollection(String caseFieldId) {
-        return getCaseField(caseFieldId).map(CaseField::isCollectionFieldType).orElse(false);
+    public Optional<CaseField> getCaseFieldByPath(String caseFieldId, String path) {
```
mario-paniccia: I think you can simplify the method to:
```
Optional<CaseField> caseFieldOptional = getCaseField(caseFieldId);

        if (StringUtils.isBlank(path)) {
            return caseFieldOptional;
        }
        return caseFieldOptional.map(f-> f.findNestedElementByPath(path));
```

mario-paniccia: I'm just realising now that the path is relative to a field id, it's not an absolute path. In this case you were right this method belongs to the CaseField not the CaseType

tzarsmango:
This PR has lots of conflicting changes with [this PR](https://github.com/hmcts/ccd-data-store-api/pull/453) Probably it would be better to review after merging the other one and refreshing this one.

PR #464, opened 2019-04-26
RDM-4636 - sort search results for divorce by lat modified date

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-4636

### Change description ###
Sorting search results for DIVORCE cases by last modified date, rest of the case types remain the same (sorted by created date).

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
```diff
@@ -121,11 +121,13 @@ public CaseDetailsQueryBuilder whereMetadata(MetaData metadata) {
         return this;
     }
 
-    public CaseDetailsQueryBuilder orderByCreatedDate(String sortDirection) {
+    public CaseDetailsQueryBuilder orderBy(MetaData metadata) {
+        String sortDirection = metadata.getSortDirection().orElse("asc");
+        String sortField = metadata.getSortField().orElse(CREATED_DATE);
         if (sortDirection.equalsIgnoreCase("asc")) {
```
MSancaktutar: default sorting may be more useful to be in reverse chronological order for date / time fields. but not an issue at this time.

gbenadikar: agree, although I just retained existing behaviour.  Also DIVORCE asked specifically for oldest first (ascending).

src/main/java/uk/gov/hmcts/ccd/data/casedetails/query/CaseDetailsQueryBuilder.java
```diff
@@ -121,11 +121,13 @@ public CaseDetailsQueryBuilder whereMetadata(MetaData metadata) {
         return this;
     }
 
-    public CaseDetailsQueryBuilder orderByCreatedDate(String sortDirection) {
+    public CaseDetailsQueryBuilder orderBy(MetaData metadata) {
+        String sortDirection = metadata.getSortDirection().orElse("asc");
```
mario-paniccia: in theory this changes a bit the behaviour of the current standard API. Because before we always got sort by Created Date. Now this could be changed on the standard API for any case type. Not a big deal though

gbenadikar: It does change the behaviour of this method, although that sort field is not passed by the standard API, so it will always be null and use Created date to sort. The aggregated API explicitly sets the sort field for DIVORCE. So this change is safe.

gbenadikar:
Not required

PR #477, opened 2019-05-13
Rdm 4336 Implement and consume external POST Create Event endpoint in API v2

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4336

### Change description ###
Implement external POST Create Event endpoint in API v2

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseType.java
```diff
@@ -142,6 +142,13 @@ public boolean hasEventId(String eventId) {
         return events.stream().anyMatch(event -> event.getId().equals(eventId));
     }
 
+    public CaseEvent findCaseEvent(String eventId) {
```
gbenadikar: Can this new method return optional just to avoid any NPEs? In EventTriggerService you can probably return null as all calls to EventTriggerService.findCaseEvent check for null.

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperation.java
```diff
@@ -71,12 +64,15 @@ public CaseDetails createCaseEvent(String uid,
             throw new ValidationException("Cannot find user roles for the user");
         }
 
+        String caseTypeId = existingCaseDetails.getCaseTypeId();
+        final CaseType caseType = caseDefinitionRepository.getCaseType(caseTypeId);
+        if (caseType == null) {
```
gbenadikar: is this null check required? I think the repository already throws ResournceNotFoundException so this check is redundant.

src/main/java/uk/gov/hmcts/ccd/v2/external/controller/CaseController.java
```diff
@@ -72,4 +81,49 @@ public CaseController(
...
+        @ApiResponse(
+            code = 404,
+            message = V2.Error.EVENT_TRIGGER_NOT_FOUND
+        ),
+        @ApiResponse(
+            code = 409,
+            message = V2.Error.CASE_ALTERED
+        )
+    })
+    public ResponseEntity<CaseResource> createEvent(@PathVariable("caseId") String caseId,
```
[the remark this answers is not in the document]
andrewfolga: Good point. Done.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseType.java
andrewfolga: Good point. Done.

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperation.java
andrewfolga: There is no null check in the repository. In fact all the other users do the null check at the moment.

PR #483, opened 2019-06-04
RDM-4064 - fix broken swagger docs

smathangi:

https://tools.hmcts.net/jira/browse/RDM-4064

### Change description ###

fix to publish swagger docs to reform-api-docs. 

```
[ ] Yes
[ x] No
```


.travis.yml
```diff
@@ -19,6 +19,7 @@ script:
 env:
   global:
     - CODACY_PROJECT_TOKEN=24f7bf62b3fa4328aa3b9bf32c07b0d9
+    - AZURE_APPLICATIONINSIGHTS_INSTRUMENTATIONKEY=SomeRandomString
```
smathangi: This can be set at docker-compose level as well. Not sure preferred one.

dwaynebailey: Was that all that was needed?  I think better to sort out at at app level, rather have it running correctly and not expose app level things in travis.

src/main/resources/application.properties
```diff
@@ -142,3 +142,5 @@ management.health.elasticsearch.enabled=false
 management.health.status.order=DOWN, OUT_OF_SERVICE, UNKNOWN, UP, OUT_OF_SYNC, PROBLEM, COULD_NOT_CONNECT
 
 spring.banner.location=classpath:ccd-data-store-banner.txt
+
+azure.application-insights.instrumentation-key=${APPINSIGHTS_INSTRUMENTATIONKEY:dummy}
```
mario-paniccia: not sure if this is relevant, but checking at the official docs on how to publish swagger it looks like we are a bit out date? 
https://github.com/hmcts/reform-api-docs#publish-swagger-docs
Since we're there shall we follow the latest recommendations?

smathangi: We are using those recommendations only https://github.com/hmcts/ccd-data-store-api/blob/RDM-4064/.travis.yml#L23-L27. 
But the issue is that script try to start app and it is failing atm due to missing appInsights instrumentation property.

PR #493, opened 2019-06-24
RDM-4824 - commit caseDetails to DB before pass to submit callback

smathangi:
https://tools.hmcts.net/jira/browse/RDM-4824

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/DefaultCreateEventOperation.java
```diff
@@ -75,6 +76,7 @@
     private final SecurityClassificationService securityClassificationService;
     private final ValidateCaseFieldsOperation validateCaseFieldsOperation;
     private final UserAuthorisation userAuthorisation;
+    private final TransactionHelper transactionHelper;
 
     @Inject
     public DefaultCreateEventOperation(@Qualifier(CachedUserRepository.QUALIFIER) final UserRepository userRepository,
```
mario-paniccia: do we have a similar issue on case creation? i.e. `DefaultCreateCaseOperation`

smathangi: good point. let me check that code as well.

smathangi: Yes :) shall i apply same fix to that flow?

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/DefaultCreateEventOperation.java
```diff
@@ -117,33 +121,16 @@ public CaseDetails createCaseEvent(final String caseReference,
...
-        validateCaseFieldsOperation.validateData(caseDetails.getData(), caseType);
-        final CaseDetails savedCaseDetails = saveCaseDetails(caseDetails, eventTrigger, newState);
-        saveAuditEventForCaseDetails(aboutToSubmitCallbackResponse, content.getEvent(), eventTrigger, savedCaseDetails, caseType);
-
-        if (!isBlank(eventTrigger.getCallBackURLSubmittedEvent())) {
+        if (!isBlank(caseEventCreator.getEventTrigger().getCallBackURLSubmittedEvent())) {
             try { // make a call back
                 final ResponseEntity<AfterSubmitCallbackResponse> callBackResponse = callbackInvoker
-                    .invokeSubmittedCallback(eventTrigger, caseDetailsBefore, savedCaseDetails);
+                    .invokeSubmittedCallback(caseEventCreator.getEventTrigger(), caseEventCreator.getCaseDetailsBefore(),
```
mario-paniccia: rather than starting a nested transaction (increasing complexity perhaps?) to handle the save, perhaps we can reduce the scope of the current transaction to include only the save. I don't think we need the submitted callback to be inside the transaction

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/CreateCaseEventService.java
```diff
@@ -0,0 +1,233 @@
...
+        this.securityClassificationService = securityClassificationService;
+        this.validateCaseFieldsOperation = validateCaseFieldsOperation;
+        this.userAuthorisation = userAuthorisation;
+        mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+    }
+
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public CreateCaseEventResult createCaseEvent(String caseReference, CaseDataContent content) {
+
+        final CaseDetails caseDetails = lockCaseDetails(caseReference);
```
mario-paniccia: I've noticed this lock operation has been moved up a few lines. Any reason for that? Not sure but probably we could lock after retrieval of CaseType and CaseEvent? to keep the lock for less time

src/main/java/uk/gov/hmcts/ccd/domain/service/util/TransactionHelper.java
```diff
@@ -0,0 +1,26 @@
+package uk.gov.hmcts.ccd.domain.service.util;
+
+import org.springframework.stereotype.Component;
```
mario-paniccia: is this class a leftover? can be removed?

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CallbackTest.java
```diff
@@ -1,11 +1,15 @@
 package uk.gov.hmcts.ccd.endpoint.std;
 
 import javax.inject.Inject;
```
mario-paniccia: is this class a leftover?

src/main/java/uk/gov/hmcts/ccd/domain/service/util/TransactionHelper.java
smathangi: yeah.. removed.

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/CreateCaseEventService.java
smathangi: it's in same position as before in DefaultCreateEventOperation. 
I guess, that's lock plus read (select for update operation). So, to retrieve CaseType and CaseEvent you need case-id.

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CallbackTest.java
smathangi: nope.. they are callback tests

PR #494, opened 2019-06-25
RDM-5190 Add new previousServiceCaseReference index.

andrewfolga:
…eCaseReference index.

https://tools.hmcts.net/jira/browse/RDM-5190

### Change description ###
Add CMC jsonb indexes
Add index on `previousServiceCaseReference` field in all environments (DEMO, AAT, PRODUCTION)
No old index on the the old field called `referenceNumber` existed before so no need to drop.

```
[ ] Yes
[x] No
```


andrewfolga:
@ckalista We could however I do not have an insight into the PROD db and cannot say the CREATE INDEX would not fail if the data was no unique in this case.

PR #497, opened 2019-07-02
 RDM-5110 UNIQUE index on case external id

smathangi:
https://tools.hmcts.net/jira/browse/RDM-5110

### Change description ###

Create a UNIQUE constraint on case external id for CMC.

```
[ ] Yes
[ x] No
```


smathangi:
Using partial unique index for externalId. This way it wouldn’t apply uniqueness to other Jurisdictions data. And it’s efficient as well. 
https://www.postgresql.org/docs/current/indexes-partial.html#INDEXES-PARTIAL-EX3

PR #502, opened 2019-07-05
Rdm 4522 full solution

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4522

### Change description ###
Allow configuration of callback timeout and number of retries

The latest addition is to build RestTemplate on each callback call.

https://stackoverflow.com/questions/37125694/how-to-set-requestconfiguration-per-request-using-resttemplate

`In contrast, the RestTemplate calls createRequest(URI, HttpMethod) (defined in HttpAccessor) which uses the ClientHttpRequestFactory. In other words: there is no option to set a request configuration per request. 
I'm not sure why Spring left this option out, it seems a reasonable functional requirement (or maybe I'm still missing something)...`

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
+++ b/src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
@@ -17,7 +17,10 @@ import uk.gov.hmcts.ccd.endpoint.exceptions.ServiceException;
 @Singleton
 public class ApplicationParams {
     @Value("#{'${ccd.callback.retries}'.split(',')}")
-    private List<Integer> callbackRetries;
+    private List<Integer> callbackRetryIntervalsInSeconds;
+
+    @Value("${http.client.read.timeout}")
+    private Integer callbackReadTimeoutInMillis;
 
     @Value("${ccd.token.secret}")
     private String tokenSecret;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
```diff
@@ -17,7 +17,10 @@
 @Singleton
 public class ApplicationParams {
     @Value("#{'${ccd.callback.retries}'.split(',')}")
-    private List<Integer> callbackRetries;
+    private List<Integer> callbackRetryIntervalsInSeconds;
+
+    @Value("${http.client.read.timeout}")
```
rafalkalita: Should we add a default value to the application.properties?

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -22,29 +20,71 @@
...
+                                           final List<Integer> callbackRetryTimeouts,
+                                           final CaseEvent caseEvent,
+                                           final CaseDetails caseDetailsBefore,
+                                           final CaseDetails caseDetails) {
+
+        return send(url, callbackRetryTimeouts, caseEvent, caseDetailsBefore, caseDetails, false);
+    }
+
+    @SuppressWarnings("javasecurity:S5145")
+    public Optional<CallbackResponse> send(final String url,
```
rafalkalita: Can't we use the following send method using generics and call getBody() on the result?

src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
andrewfolga: The default value in application.properties is `60000`.

rafalkalita: Fine. Couldn't find it in the PR as the code was folded and the property was already there.

PR #503, opened 2019-07-08
Rdm 4522 full solution

andrewfolga:
JIRA link (if applicable)
https://tools.hmcts.net/jira/browse/RDM-4522

Change description
Allow configuration of callback timeout and number of retries

The latest addition is to build RestTemplate on each callback call.

https://stackoverflow.com/questions/37125694/how-to-set-requestconfiguration-per-request-using-resttemplate

In contrast, the RestTemplate calls createRequest(URI, HttpMethod) (defined in HttpAccessor) which uses the ClientHttpRequestFactory. In other words: there is no option to set a request configuration per request. I'm not sure why Spring left this option out, it seems a reasonable functional requirement (or maybe I'm still missing something)...

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
@@ -49,7 +49,7 @@ class RestTemplateConfiguration {
     @Bean(name = "restTemplate")
     public RestTemplate restTemplate() {
         final RestTemplate restTemplate = new RestTemplate();
-        HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory(getHttpClient());
+        HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory(httpClient());
         requestFactory.setReadTimeout(readTimeout);
         LOG.info("readTimeout: {}", readTimeout);
         restTemplate.setRequestFactory(requestFactory);
@@ -79,7 +79,8 @@ class RestTemplateConfiguration {
         }
     }
 
-    private HttpClient getHttpClient() {
+    @Bean(name = "httpClient")
+    public HttpClient httpClient() {
         return getHttpClient(connectionTimeout);
     }
 
@@ -99,15 +100,15 @@ class RestTemplateConfiguration {
         final RequestConfig
             config =
             RequestConfig.custom()
-                         .setConnectTimeout(timeout)
-                         .setConnectionRequestTimeout(timeout)
-                         .setSocketTimeout(timeout)
-                         .build();
+                .setConnectTimeout(timeout)
+                .setConnectionRequestTimeout(timeout)
+                .setSocketTimeout(timeout)
+                .build();
 
         return HttpClientBuilder.create()
-                                .useSystemProperties()
-                                .setDefaultRequestConfig(config)
-                                .setConnectionManager(cm)
-                                .build();
+            .useSystemProperties()
+            .setDefaultRequestConfig(config)
+            .setConnectionManager(cm)
+            .build();
     }
 }
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.ccd.domain.service.callbacks;
 
+import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -8,12 +9,11 @@ import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
-import org.springframework.retry.annotation.Backoff;
-import org.springframework.retry.annotation.Retryable;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.stereotype.Service;
 import org.springframework.web.client.RestClientException;
 import org.springframework.web.client.RestTemplate;
-
+import uk.gov.hmcts.ccd.ApplicationParams;
 import uk.gov.hmcts.ccd.data.SecurityUtils;
 import uk.gov.hmcts.ccd.domain.model.callbacks.CallbackRequest;
 import uk.gov.hmcts.ccd.domain.model.callbacks.CallbackResponse;
@@ -22,29 +22,71 @@ import uk.gov.hmcts.ccd.domain.model.definition.CaseEvent;
 import uk.gov.hmcts.ccd.endpoint.exceptions.ApiException;
 import uk.gov.hmcts.ccd.endpoint.exceptions.CallbackException;
 
+import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
+import static java.util.Optional.ofNullable;
 import static org.springframework.util.CollectionUtils.isEmpty;
-
-// RDM-4316 discarded timeout/backoff value in case event definition until requirements are cleared
 
 @Service
 public class CallbackService {
     private static final Logger LOG = LoggerFactory.getLogger(CallbackService.class);
+    public static final int CALLBACK_RETRY_INTERVAL_MULTIPLIER = 3;
 
     private final SecurityUtils securityUtils;
     private final RestTemplate restTemplate;
+    private final List<Integer> defaultCallbackRetryIntervalsInSeconds;
+    private final Integer defaultCallbackTimeoutInMillis;
+
+    static class CallbackRetryContext {
+        private final Integer callbackRetryInterval;
+        private final Integer callbackRetryTimeout;
+
+        CallbackRetryContext(final Integer callbackRetryInterval, final Integer callbackRetryTimeout) {
+            this.callbackRetryInterval = callbackRetryInterval;
+            this.callbackRetryTimeout = callbackRetryTimeout;
+        }
+
+        Integer getCallbackRetryInterval() {
+            return callbackRetryInterval;
+        }
+
+        Integer getCallbackRetryTimeout() {
+            return callbackRetryTimeout;
+        }
+
+    }
 
     @Autowired
     public CallbackService(final SecurityUtils securityUtils,
-                           @Qualifier("restTemplate") final RestTemplate restTemplate) {
+                           @Qualifier("restTemplate") final RestTemplate restTemplate,
+                           final ApplicationParams applicationParams) {
         this.securityUtils = securityUtils;
         this.restTemplate = restTemplate;
+        this.defaultCallbackRetryIntervalsInSeconds = applicationParams.getCallbackRetryIntervalsInSeconds();
+        this.defaultCallbackTimeoutInMillis = applicationParams.getCallbackReadTimeoutInMillis();
     }
 
-    // The retry will be on seconds T=1 and T=3 if the initial call fails at T=0
-    @Retryable(value = {CallbackException.class}, maxAttempts = 3, backoff = @Backoff(delay = 1000, multiplier = 3))
     public Optional<CallbackResponse> send(final String url,
+                                           final List<Integer> callbackRetryTimeouts,
+                                           final CaseEvent caseEvent,
+                                           final CaseDetails caseDetails) {
+        return send(url, callbackRetryTimeouts, caseEvent, null, caseDetails);
+    }
+
+    public Optional<CallbackResponse> send(final String url,
+                                           final List<Integer> callbackRetryTimeouts,
+                                           final CaseEvent caseEvent,
+                                           final CaseDetails caseDetailsBefore,
+                                           final CaseDetails caseDetails) {
+
+        return send(url, callbackRetryTimeouts, caseEvent, caseDetailsBefore, caseDetails, false);
+    }
+
+    @SuppressWarnings("javasecurity:S5145")
+    public Optional<CallbackResponse> send(final String url,
+                                           final List<Integer> callbackRetryTimeouts,
                                            final CaseEvent caseEvent,
                                            final CaseDetails caseDetailsBefore,
                                            final CaseDetails caseDetails,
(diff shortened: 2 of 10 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContext.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.hmcts.ccd.domain.service.callbacks.retrycontext;
+
+import lombok.ToString;
+
+@ToString
```
mario-paniccia: can you add javadoc please? explaining the retry strategy

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
```diff
@@ -0,0 +1,65 @@
+package uk.gov.hmcts.ccd.domain.service.callbacks.retrycontext;
+
+import com.google.common.collect.Lists;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.ccd.ApplicationParams;
+
+import java.util.List;
+
+@Service
+public class CallbackRetryContextBuilder {
```
mario-paniccia: can you add javadoc please? explaining the retry strategy

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+    private final Integer defaultCallbackTimeoutInSeconds;
+
+    public CallbackRetryContextBuilder(final ApplicationParams applicationParams) {
+        this.defaultCallbackRetryIntervalsInSeconds = applicationParams.getCallbackRetryIntervalsInSeconds();
+        this.defaultCallbackTimeoutInSeconds = applicationParams.getCallbackTimeoutInSeconds();
+    }
+
+    public List<CallbackRetryContext> buildCallbackRetryContexts(final List<Integer> callbackRetryTimeouts) {
+        List<CallbackRetryContext> retryContextList = Lists.newArrayList();
+        if (callbackRetryTimeouts.isEmpty() || isCallbackRetriesWithZeros(callbackRetryTimeouts)) {
```
mario-paniccia: can you please add a comment to the ticket or update the description mentioning that if the callback config contains a 0 we ignore the config and use the defaults?

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+            buildDefaultRetryContext(retryContextList);
+        } else if (isCallbackRetriesDisabled(callbackRetryTimeouts)) {
+            disableRetryContext(retryContextList);
```
mario-paniccia: same as above, can you document this in the ticket please, it's not there currently

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+        retryContextList.add(new CallbackRetryContext(0, defaultCallbackTimeoutInSeconds));
+    }
+
+    private void buildDefaultRetryContext(final List<CallbackRetryContext> retryContextList) {
+        this.defaultCallbackRetryIntervalsInSeconds.forEach(cbRetryInterval -> retryContextList.add(
+            new CallbackRetryContext(cbRetryInterval, defaultCallbackTimeoutInSeconds)));
+    }
+
+    private void buildCustomRetryContext(final List<Integer> callbackRetryTimeouts, final List<CallbackRetryContext> retryContextList) {
+        retryContextList.add(new CallbackRetryContext(0, callbackRetryTimeouts.remove(0)));
```
mario-paniccia: here and in line 48 we are hardcoding the values 0 and 1. This issue with this is that if we change the config `ccd.callback.retry.intervals=0,1,3` that is not picked up here and could have some misleading behaviour

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+        if (!callbackRetryTimeouts.isEmpty()) {
+            retryContextList.add(new CallbackRetryContext(1, callbackRetryTimeouts.remove(0)));
+            for (Integer callbackRetryTimeout : callbackRetryTimeouts) {
+                retryContextList.add(
+                    new CallbackRetryContext(
+                        getLastElement(retryContextList).getCallbackRetryInterval() * CALLBACK_RETRY_INTERVAL_MULTIPLIER,
```
mario-paniccia: you cold use `Iterables.getLast(retryContextList)`

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -185,13 +152,46 @@ private void sleep(final Integer timeout) {
...
+        } catch (ExecutionException e) {
+            handleException("Execution exception. ", url, e);
+            if (e.getCause() instanceof RuntimeException) {
+                throw (RuntimeException)e.getCause();
+            } else {
+                return Optional.empty();
+            }
+        } catch (TimeoutException e) {
+            future.cancel(true);
+            handleException("Future timed out. ", url, e);
```
mario-paniccia: is it possible to add an indication to the message that the callback will be retried?

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -185,13 +152,46 @@ private void sleep(final Integer timeout) {
...
-        return ofNullable(exchange);
+        Future<ResponseEntity> future = executorService.submit(() -> restTemplate.exchange(url, HttpMethod.POST, requestEntity, clazz));
+
+        return handleResponse(url, timeout, future, caseTypeId, eventId);
+    }
+
+    @SuppressWarnings("javasecurity:S5145")
+    private <T> Optional<ResponseEntity<T>> handleResponse(final String url, final Integer timeout, final Future<ResponseEntity> future,
+                                                           final String caseTypeId, final String eventId) {
+        try {
```
mario-paniccia: I have some concerns here. Have we considered using the retry and timeout features of Apache HTTP Client library? Rather than implementing a custom retry and timeout mechanism ourselves? From a (very) quick look I've seen that Apache HTTP Client allows setting timeouts per request. And has the ability to configure retries. Perhaps we could find a way to make it work?

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
andrewfolga: Sure done now.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
andrewfolga: Sure added to the ticket

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
andrewfolga: Done in the ticket now.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
andrewfolga: The `ccd.callback.retry.intervals` is only used for building default retry context and this is custom retry context. It works on the notion we multiple every successive interval value by 3. Do you mean we should provide a multiplier in the properties as well? That was not the requirement of the ticket however.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
andrewfolga: Done.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
andrewfolga: It will be logged anyway with the logging of `Trying url={} with timeout={}s` in the following retry attempt.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
andrewfolga: That was tried before and it failed in DEMO. Besides we thought (Fatih and I) that having the timeout defined encompassing the entire request is actually what services and anyone configuring the timeouts in the spreadsheet would expect. Defining it on http client level does not guarantee breaching the timeout would drop the request as explained before due to the read timeout being socket timeout.

See also:
https://stackoverflow.com/questions/9873810/using-apache-httpclient-how-to-set-the-timeout-on-a-request-and-response

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
mario-paniccia: I meant something different. According to the ticket, the retry interval should always be 0,1,3... for both default and custom callback settings.
The way things are implemented now it's possible instead that we can end up with different values for the retry interval used for the default behaviour and the custom behaviour:
Example: 
change ccd.callback.retry.intervals to say 2,3,10. Now default callbacks keep using 0,1,3 but the custom start using 2,3,10. So there is a divergence which makes no sense.
So we either remove 'ccd.callback.retry.intervals', so that it cannot be reconfigured and it's always hardcoded to 0,1,3. Or we change the code below to pick the new values for 'ccd.callback.retry.intervals'.
But we cannot let the 2 diverge.

andrewfolga: Yes ok I will change to keep the default retry intervals property and use it in the first 3 tries of custom context.

andrewfolga: done

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+ * Blank value means default retries (0s interval with 60s default timeout, 1s interval with 60s default timeout, 3s interval with 60s default timeout).
+ * Single 0 value means disable retries (single send at 0s interval with 60s default timeout).
+ * Any comma separated chain of values that DOES NOT include 0 means custom retries.
+ * Any comma separated chain of values that DOES include 0 means default retries (see blank value).
+ *
+ * The examples input callback retry timeouts and output callback retry contexts are as follows:
+ *
+ *      blank - default retries context (0s interval with 60s default timeout, 1s interval with 60s default timeout, 3s interval with 60s default timeout)
+ *      0 - disable retries (0s interval with 60s default callbacks timeout)
+ *      1 - custom retries (0s interval with 1s custom timeout)
```
mario-paniccia: `1` has no retries isn't it? `custom retries` is a bit misleading

src/main/resources/application.properties
```diff
@@ -51,6 +51,7 @@ ccd.token.secret=${DATA_STORE_TOKEN_SECRET:AAAAAAAAAA}
 
 #callback timeouts - comma separated integers in seconds
 ccd.callback.retry.intervals=0,1,3
+ccd.callback.timeout=${CCD_CALLBACK_TIMEOUT:60}
```
mario-paniccia: this is only used for defaults, so probably better call it `ccd.default.callback.timeout`

src/main/resources/application.properties
```diff
@@ -122,6 +123,8 @@ http.client.connection.timeout=${HTTP_CLIENT_CONNECTION_TIMEOUT:30000}
 http.client.read.timeout=${HTTP_CLIENT_READ_TIMEOUT:60000}
 http.client.connection.drafts.create.timeout=${HTTP_CLIENT_CONNECTION_DRAFTS_CREATE_TIMEOUT:1000}
 http.client.connection.drafts.timeout=${HTTP_CLIENT_CONNECTION_DRAFTS_TIMEOUT:500}
+http.client.connection.callbacks.timeout=${HTTP_CLIENT_CONNECTION_CALLBACKS_TIMEOUT:300000}
+http.client.read.callbacks.timeout=${HTTP_CLIENT_READ_CALLBACKS_TIMEOUT:300000}
```
mario-paniccia: this is interesting here. We always set on the library a fixed timeout of 5 min. Question, for my knowledge: imagine we configure a callback timeout of 40 secs. When the future that executes the callback times out after 40secs and returns, is the connection cancelled and released to the http connection pools or that will happen only after the 5 mins timeout is reached?

mario-paniccia: also, for my knowledge, this means that even if the service configures 10 min callback timeout on the definitions, the library will ignore that and timeout after max 5 min anyway am I correct?

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
andrewfolga: `0` is no retries. `1` is part of custom retries.

src/main/resources/application.properties
andrewfolga: Sure and changed now.

src/main/resources/application.properties
andrewfolga: First question: connection is release to the pool as soon as it is cancelled after the configured 40s custom timeout.
Second question: Yes the library will time out after 5 mins.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/retrycontext/CallbackRetryContextBuilder.java
mario-paniccia: ok but with `1` there are no retries am I correct?

andrewfolga: Yes with `1` there will be no retries. It will be initial interval of `0` and then a timeout of `1` and that is it.

src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
```diff
@@ -108,6 +141,7 @@ private HttpClient getHttpClient(final int timeout) {
                                 .useSystemProperties()
                                 .setDefaultRequestConfig(config)
                                 .setConnectionManager(cm)
+                                .setRetryHandler(new DefaultHttpRequestRetryHandler(3, false))
```
andrewfolga: During performance testing Fatih and I spotted an issue (org.apache.http.NoHttpResponseException) with server dropping connections under certain load and a solution on spring forums was suggested to add retry handler and also increase http.client.validate.after.inactivity property to 100 from intial 0 to help detect connections that have become stale (half-closed) while kept inactive in the pool.

src/main/resources/application.properties
```diff
@@ -117,11 +118,13 @@ spring.application.name=ccd-data-store
 http.client.max.total=${HTTP_CLIENT_MAX_TOTAL:100}
 http.client.seconds.idle.connection=${HTTP_CLIENT_SECONDS_IDLE_CONNECTION:120}
 http.client.max.client_per_route=${HTTP_CLIENT_MAX_CLIENT_PER_ROUTE:20}
-http.client.validate.after.inactivity=${HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY:0}
+http.client.validate.after.inactivity=${HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY:100}
```
andrewfolga: During performance testing Fatih and I spotted an issue (org.apache.http.NoHttpResponseException) with server dropping connections under certain load and a solution on spring forums was suggested to add retry handler and also increase http.client.validate.after.inactivity property to 100 from intial 0 to help detect connections that have become stale (half-closed) while kept inactive in the pool.

PR #516, opened 2019-07-26
#RDM-5118_functional_tests_for_ccd-am_integration_datastore

vinoth-senthamaraikannan:
JIRA link (if applicable)
https://tools.hmcts.net/jira/browse/RDM-5118

Change description
Added a new functional test methods to test the following functionalities given a caseworker has different access (case_type_acl) on a Case Type.
1. Create Case
2. Update Case
3. Get Case
4. Search Case

No


src/aat/java/uk/gov/hmcts/ccd/datastore/tests/functional/UpdateCaseTest.java
```diff
@@ -68,4 +68,188 @@ void shouldUpdateSingleField() {
...
+            .as(asAutoTestCaseworker())
+            .withData(
+                CaseData.builder()
+                    .numberField(UPDATED_NUMBER)
+                    .build()
+            )
+            .submit()
+
+            .then()
+            .statusCode(201);
```
MSancaktutar: When testing update, more fields should be updated and whether the values are indeed updated should be verified. For this, we should set a value to first X, verify that it's indeed X, then update it to Y, and verify that it's indeed Y. Otherwise, we may get a 201 response but may be not updating the fields. Ideally, we should verify this with all the fields we're can be updating with an update call.

src/aat/java/uk/gov/hmcts/ccd/datastore/tests/functional/UpdateCaseTest.java
```diff
@@ -68,4 +68,188 @@ void shouldUpdateSingleField() {
...
+            .withData(
+                CaseData.builder()
+                    .numberField(UPDATED_NUMBER)
+                    .build()
+            )
+            .submit()
+
+            .then()
+            .statusCode(404);
+    }
```
MSancaktutar: We should verify that another user with enough access to this case cen see this case, as a verification of impossible deletion despite existence of the case.

src/aat/java/uk/gov/hmcts/ccd/datastore/tests/functional/UpdateCaseTest.java
vinoth-senthamaraikannan: Mutlu,

  Only for few test methods where the user role doesn't have the "READ" privilege I didn't get the response content back as they only have the privilege to update the case but don't have read permission. Anyways, I have updated the these test methods, by fetching the response using the super user and asserted that the case data was successfully updated.

src/aat/java/uk/gov/hmcts/ccd/datastore/tests/functional/UpdateCaseTest.java
vinoth-senthamaraikannan: Mutlu,

  In the same test method, I had infact created and retrieved the case data using the super user (private case worker). So, this already proves that another user have access to read the case data. Please confirm if that fine or I need to add a code snippet to update the same case super user. I thought that is not needed as we already have other test methods to test those scenarios.

PR #520, opened 2019-08-01
Rdm 5545 Option to disable retires from ccd definition file.

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-5545

### Change description ###
Option to disable retires from ccd definition file.

This should only disable current default logic (1, 5, 10 intervals each having 60s timeouts) when one specifies 0 meaning no retries and just one initial call. All other values put in there will go to current default behaviour.

Do not be intimidated by the amount of changes - by introducing the option to disable retries we have to get rid of @Retryable annotation as we will only know about that flag at runtime.

The changes are also a small prep work before we deliver RDM-4522.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
```diff
@@ -33,7 +33,10 @@
     private List<String> readFromAMCaseTypes;
 
     @Value("#{'${ccd.callback.retries}'.split(',')}")
-    private List<Integer> callbackRetries;
+    private List<Integer> callbackRetryIntervalsInSeconds;
```
mario-paniccia: this new name could be confused with the interval of time in seconds between callbacks, which is something different. Can we call it `callbackDurationInSeconds`?

andrewfolga: This actually is the interval of time in seconds between callbacks.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -22,79 +23,118 @@
...
+        ResponseEntity<CallbackResponse> responseResponseEntity = send(url, callbackRetryTimeouts, caseEvent,
+            caseDetailsBefore, caseDetails, CallbackResponse.class, false);
+        return Optional.ofNullable(responseResponseEntity.getBody());
     }
 
-    // The retry will be on seconds T=1 and T=3 if the initial call fails at T=0
-    @Retryable(value = {CallbackException.class}, maxAttempts = 3, backoff = @Backoff(delay = 1000, multiplier = 3))
+    @SuppressWarnings("javasecurity:S5145")
     public <T> ResponseEntity<T> send(final String url,
+                                      final List<Integer> callbackRetryTimeouts,
```
mario-paniccia: `callbackRetryTimeouts` is not consistent with the name chosen for the default setting, which is `defaultCallbackRetryIntervalsInSeconds`. A bit confusing

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -106,4 +146,49 @@ public void validateCallbackErrorsAndWarnings(final CallbackResponse callbackRes
                 .withWarnings(callbackResponse.getWarnings());
         }
     }
+
+    private List<CallbackRetryContext> buildCallbackRetryContexts(final List<Integer> callbackRetryTimeouts) {
```
mario-paniccia: I think we need to be more uniform with the naming of variables we chose. Otherwise it's complicated to follow the code. In this method parameter we introduce a new name `callbackRetryTimeouts` for the same concept. We have already `callbackRetryIntervalsInSeconds` and `RetryTimeouts`

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
```diff
@@ -106,4 +146,49 @@ public void validateCallbackErrorsAndWarnings(final CallbackResponse callbackRes
...
+        } else {
+            this.defaultCallbackRetryIntervalsInSeconds.forEach(cbRetryInterval -> {
+                retryContextList.add(new CallbackRetryContext(cbRetryInterval, defaultCallbackTimeoutInMillis));
+            });
+        }
+        return retryContextList;
+    }
+
+    private boolean isCallbackRetriesDisabled(final List<Integer> callbackRetryTimeouts) {
+        return callbackRetryTimeouts.size() == 1 && callbackRetryTimeouts.get(0) == 0;
```
mario-paniccia: we already discussed about this, so you can ignore this comment. Commenting anyway just for future reference.
Why we are mandating that the value must be `0` for disabling retries? That was not a requirement of RDM-4522. And we can achieve disabling retries without forcing the value to `0`. Retries are disabled by passing one single value rather than a list. The values denote the max callback timeout

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
andrewfolga: I think there is just a confusion here. It is indeed `callbackRetryTimeouts` that we want here - it is the timeout each consecutive callback should last for. `callbackRetryIntervalsInSeconds` is a different thing - it is the interval each consecutive callback should be be delayed for before execution. The `InSeconds` suffix which could be followed up on similar named variables that do not have it will be addressed as part of 4522 not to delay this any longer.

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
andrewfolga: Sure and it was already explained 👍

andrewfolga: `0` was chosen as after 4522 it would become an invalid case (0 is not a valid callback timeout) and should not be honoured so we can treat it as a missed requirement.
Also allowing value other `0` could confuse ppl using it now (we would have to announce that too that single values other than `0` also mean disable timeouts) which would confuse people even more. This is counterintuitive and would not make sense to services. We would have to then revert it's meaning when 4522 comes in because we would want single non `0` values to now have a callback timeout meaning rather than merely disabling retry timeouts.
This ticket was to introduce an optional disable of retries whilst making the least amount of functional change possible

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
andrewfolga: As mentioned above this is actually desired.
`callbackRetryTimeouts` is the callbacks duration dynamically configurable by spreadsheet. `defaultCallbackRetryIntervalsInSeconds` on the other hand is the default interval of time in seconds between callbacks (it is not callback duration) that is static here and comes from the application properties.

src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
mario-paniccia: ok I see. Then we should rename `ccd.callback.retries` as well?

src/main/java/uk/gov/hmcts/ccd/domain/service/callbacks/CallbackService.java
mario-paniccia: yep got it now got confused by the `ccd.callback.retries` left in the code

andrewfolga:
Yes we could have done what you suggest but we would have to scrap that in future anyway (we have to get rid of retryables annotations anyway) and with this solution we do not have to scrap anything in the future. Also this solution does not touch the thing that was breaking last time which was the actual changing of the timeouts from call to call. Finally and most importantly it allows us to introduce more logging before the callback is made, just after the callback is being made and any execution exceptions that the callback has generated which will help us investigate any issues. The Retryable code was lacking that.

src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
andrewfolga: Sure changed now. Ta

PR #529, opened 2019-08-15
RDM-5650 - remove duplicate search criteria (JPA) 

smathangi:

https://tools.hmcts.net/jira/browse/RDM-5650

### Change description ###
removed duplicate search criteria (JPA) in DefaultCaseDetaRespository.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -151,3 +151,5 @@ ccd.am.write.to_both=${CCD_AM_WRITE_TO_BOTH:}
 
 ccd.am.read.from_ccd=${CCD_AM_READ_FROM_CCD:AUTOTEST1,TEST}
 ccd.am.read.from_am=${CCD_AM_READ_FROM_AM:}
+
+ccd.case.search.jpa.criteria.enabled=${JPA_CRITERIA_IN_SEARCH_ENABLED:false}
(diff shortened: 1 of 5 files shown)


src/main/resources/application.properties
```diff
@@ -151,3 +151,5 @@ ccd.am.write.to_both=${CCD_AM_WRITE_TO_BOTH:}
 
 ccd.am.read.from_ccd=${CCD_AM_READ_FROM_CCD:AUTOTEST1,TEST}
 ccd.am.read.from_am=${CCD_AM_READ_FROM_AM:}
+
+ccd.case.search.jpa.criteria.enabled=${JPA_CRITERIA_IN_SEARCH_ENABLED:false}
```
mario-paniccia: we need to define this as a var in Terraform and update the terraform files so that we can set this back to true in case of issues on Prod

smathangi: yeah.. i thought of it.. but dev-ops has to update that value in app_settings and restart webapp. If that's the case they can add a new param itself to revert 'JPA_CRITERIA_IN_SEARCH_ENABLED'

mario-paniccia: I'd stick with conventions :)

PR #541, opened 2019-08-25
Document restore

mario-paniccia:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/DocLinksRestoreService.java
```diff
@@ -0,0 +1,133 @@
...
+
+@Component
+public class DocLinksRestoreService {
+    private static final Logger LOG = LoggerFactory.getLogger(DocLinksRestoreService.class);
+
+    public static final String START_TIME = "'2019-08-20 15:50:00.000'";
+    public static final String END_TIME = "'2019-08-21 13:30:00.000'";
+    public static final String BETWEEN_CLAUSE = "BETWEEN  " + START_TIME + " AND " + END_TIME;
+
+    public static final String OLD_CASE_QUERY = "SELECT * FROM case_data WHERE last_modified " + BETWEEN_CLAUSE + " AND created_date < " + START_TIME;
```
mario-paniccia: OLD_CASE_QUERY : aren't we skipping cases that have events in the bug period but that have then been modified after the bug period? Those should be included as well.

mario-paniccia: I think the correct way would be to find the id of any case which has any event in the bug timeframe.

smathangi: Yes, that's right.

src/main/java/uk/gov/hmcts/ccd/domain/service/DocLinksRestoreService.java
```diff
@@ -0,0 +1,133 @@
...
+            .collect(Collectors.toList());
+    }
+
+    private Map<Long, CaseDetailsEntity> getCaseIdMap(List<CaseDetailsEntity> caseDetailsEntities) {
+        return caseDetailsEntities.stream()
+                .collect(Collectors.toMap(CaseDetailsEntity::getId, Function.identity()));
+    }
+
+    private boolean hasMissingDocuments(CaseDetailsEntity caseDetails, CaseAuditEventEntity caseEvent) {
+        List<String> eventDocValues = caseEvent.getData().findValuesAsText(DOCUMENT_URL);
```
mario-paniccia: we are not distinguishing between a doc in a collection and a doc is a simple field. As a result we might be classifying as 'Needs Review by User' more cases than we should. Because users can only delete themselves from collections. So if a case has got all non collections doc fields, and is recovered, the recovery is correct without doubts. We should not mark the case as 'Needs Review by User'. I think it's worth improving the program to do that

mario-paniccia: I guess we could add this check as part of the Phase 2 and ignore it here.

smathangi: The current logic correctly identifies impacted case irrespective of where the doc links are missing. So, we need a classification here something like `impacted but not required a user review` . 
First of all, are we sure docLinks in a simple filed actually missing because of this bug? I mean anyone verified locally?

mario-paniccia: Yes agree we need a classification `impacted but not required a user review`. And yes simple document fields are impacted too.

src/main/java/uk/gov/hmcts/ccd/domain/service/DocLinksRestoreService.java
```diff
@@ -0,0 +1,134 @@
...
+        List<String> caseDocValues = caseDetails.getData().findValuesAsText(DOCUMENT_URL);
+        return !caseDocValues.containsAll(eventDocValues);
```
mario-paniccia: `caseDocValues.containsAll(eventDocValues)`:
there could be false positives here. For example imagine `docFieldA` was wiped out, then the user after the bug period re uploads a doc to `docFieldA` but with a different name. The user has manually fixed the issue itself but we are marking the case as `hasMissingDocuments` because the old name doesn't match the new one.

mario-paniccia: I think we should compare the event not with the case data but with the last event in the bug timeframe. That'll give us more accurate results

smathangi: Distinguishing manually corrected cases in a collection is very difficult, probably not possible. Where as in simple fields it's possible and we can treat presence of a value at that path means it's corrected or not missed at all. 
The more i think about event states more cases are coming up 😄

src/main/java/uk/gov/hmcts/ccd/domain/service/DocLinksRestoreService.java
```diff
@@ -0,0 +1,134 @@
...
+        return ListUtils.union(oldCases, newCases);
+    }
+
+    private List<CaseDetailsEntity> findDocLinksMissedOldCases(List<String> jurisdictionList) {
+        Query query = jurisdictionList.isEmpty() ? em.createNativeQuery(OLD_CASE_QUERY, CaseDetailsEntity.class) :
+            em.createNativeQuery(OLD_CASE_QUERY_WITH_JUIDS, CaseDetailsEntity.class).setParameter("jids", jurisdictionList);
+        return findInOldCasesBasedOnEventHistory(query.getResultList());
+    }
+
+    private List<CaseDetailsEntity> findDocLinksMissedNewCases(List<String> jurisdictionList) {
```
mario-paniccia: minor. feel free to ignore. we could also reduce even more the scope by passing case type rather than jurisdiction.

src/main/java/uk/gov/hmcts/ccd/domain/service/DocLinksRestoreService.java
smathangi: > I think we should compare the event not with the case data but with the last event in the bug timeframe. That'll give us more accurate results
Actually `document_url` is unique, so does it matter?

src/main/java/uk/gov/hmcts/ccd/domain/service/DocLinksRestoreService.java
```diff
@@ -0,0 +1,134 @@
...
+    public static final String NEW_CASE_QUERY_WITH_JUIDS = NEW_CASE_QUERY +  " AND jurisdiction IN :jids";
+
+    public static final String NEW_CASE_EVENTS = "SELECT * FROM case_event WHERE case_data_id IN :caseIds";
+
+    public static final String DOCUMENT_URL = "document_url";
+
+    @PersistenceContext
+    private EntityManager em;
+
+    public List<CaseDetailsEntity> findDocLinksMissedCases(List<String> jurisdictionList) {
```
mario-paniccia: shall we return for each case the list or number of docs missing?

src/main/java/uk/gov/hmcts/ccd/domain/service/DocLinksRestoreService.java
mario-paniccia: what I'm saying is that if you compare the event with the case, you will generate false positives also for simple fields. We can and should avoid that. We can do by comparing first event of bug timeframe with the last one rather then with the case

PR #542, opened 2019-08-27
RDM-5545 Minimal solution that disables callback retries.

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-5545

### Change description ###
Option to disable retires from ccd definition file.

This should only disable current default retry logic (1 and 3 s intervals each having 60s timeouts) when one specifies 0 meaning no retries and just one initial call. All other values put in there will go to current default behaviour.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
```diff
@@ -108,6 +109,7 @@ private HttpClient getHttpClient(final int timeout) {
                                 .useSystemProperties()
                                 .setDefaultRequestConfig(config)
                                 .setConnectionManager(cm)
+                                .setRetryHandler(new DefaultHttpRequestRetryHandler(3, false))
```
mario-paniccia: can you explain this change?

andrewfolga: During performance testing of 4522 Fatih and I spotted an issue (`org.apache.http.NoHttpResponseException`) with server dropping connections under certain load and a solution on spring forums was suggested to add retry handler and also increase `http.client.validate.after.inactivity` property to `100` from intial `0` to help detect connections that have become stale (half-closed) while kept inactive in the pool.

However as this is a simple fix for only the PROD issue we could remove it from here and apply as part of 4522.

Let me know your thoughts.

src/main/resources/application.properties
```diff
@@ -117,7 +117,7 @@ spring.application.name=ccd-data-store
 http.client.max.total=${HTTP_CLIENT_MAX_TOTAL:100}
 http.client.seconds.idle.connection=${HTTP_CLIENT_SECONDS_IDLE_CONNECTION:120}
 http.client.max.client_per_route=${HTTP_CLIENT_MAX_CLIENT_PER_ROUTE:20}
-http.client.validate.after.inactivity=${HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY:0}
+http.client.validate.after.inactivity=${HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY:100}
```
andrewfolga: Defines period of inactivity in milliseconds after which persistent connections must be re-validated prior to being leased to the consumer. Non-positive value passed to this method disables connection validation. This check helps detect connections that have become stale (half-closed) while kept inactive in the pool.

This is changed to rectify the problem with `org.apache.http.NoHttpResponseException` that was spotted as part of perf testing of 4522 but could be removed from this PR (which deals solely with PROD issue) and added to 4522 PR only.

Let me know.

src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
mario-paniccia: yes please let's remove it from here if they don't belong strictly to this ticket. Cheers

andrewfolga: Sure done.

PR #567, opened 2019-10-02
moved test environment variables a level up so that they are available for AKS testing

smathangi:


Jenkinsfile_CNP
```diff
@@ -53,6 +53,13 @@ def vaultOverrides = [
     'preview' : 'aat',
     'spreview': 'saat'
 ]
+// Vars needed for AKS testing
```
mario-paniccia: can make it a bit more explicit please:
Vars needed for functional and smoke tests run against AKS

PR #570, opened 2019-10-04
added app-insights key config.

smathangi:


charts/ccd-data-store-api/values.yaml
```diff
@@ -12,6 +12,7 @@ java:
         - ccd-ELASTIC-SEARCH-URL
         - ccd-ELASTIC-SEARCH-DATA-NODES-URL
         - ccd-ELASTIC-SEARCH-PASSWORD
+        - AppInsightsInstrumentationKey
```
MSancaktutar: I'm concerned with these names not following a convention, but not an issue of this ticket, I see.

timja: somewhat a requirement of the pipeline, not easily changeable now.

PR #572, opened 2019-10-07
Pool size & Replica config

smathangi:


charts/ccd-data-store-api/values.aat.template.yaml
```diff
@@ -1,5 +1,4 @@
 java:
-  replicas: 2
```
mario-paniccia: shouldn't AAT be an exact replica of Prod?

charts/ccd-data-store-api/values.yaml
```diff
@@ -20,6 +20,7 @@ java:
     DATA_STORE_DB_HOST: ccd-data-store-api-postgres-db-{{ .Values.global.environment }}.postgres.database.azure.com
     DATA_STORE_DB_USERNAME: ccd@ccd-data-store-api-postgres-db-{{ .Values.global.environment }}
     DATA_STORE_DB_OPTIONS: "?stringtype=unspecified&sslmode=require"
+    DATA_STORE_DB_MAX_POOL_SIZE: 48
```
mario-paniccia: Can we double check in AAT this has the same value as Prod?

charts/ccd-data-store-api/values.aat.template.yaml
smathangi: This config is staging pod only which is used by functional tests only.

charts/ccd-data-store-api/values.yaml
smathangi: yes, it's 48 for aat as well in terraform.

charts/ccd-data-store-api/values.aat.template.yaml
adusumillipraveen: yes, aat.template.yaml is used only for staging deployment where tests are run. Its a bit redundant config which we will try to get rid of soon.

mario-paniccia: 👍 That'd be great @adusumillipraveen if you can remove it. Much less maintainability effort

PR #578, opened 2019-10-15
RDM-6005 Change retry to only work on UUID collision

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-6005

### Change description ###
CMC doesn't want the retry on externalId collision, so I have changed the retry to work only when UUID collision happens.

```
[ ] Yes
[X ] No
```


src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -76,8 +77,13 @@ public CaseDetails set(final CaseDetails caseDetails) {
         } catch (PersistenceException e) {
             LOG.warn("Failed to store case details", e);
             if (e.getCause() instanceof ConstraintViolationException) {
-                throw new CaseConcurrencyException(e.getMessage());
+                LOG.warn("ConstraintViolationException happen for UUID={}. Cause: {}",
+                    caseDetails.getReference(), ((ConstraintViolationException) e.getCause()).getConstraintName());
+                if (((ConstraintViolationException) e.getCause()).getConstraintName().equals(UNIQUE_INDEX_EXTERNAL_ID_CONSTRAINT)) {
```
smathangi: this is very specific to CMC index.. we might have different unique indexes as well (Ethos).

ckalista: Indeed, only CMC Jurisdiction has this index in place. It may not be needed to check the exact name of the constraint here to make your solution slightly more generic?

rafalkalita: Reversed this and now the code has nothing CMC specific.

ckalista: Looks good to me 👍 

Are we still going ahead with this change then? You mentioned that you would investigate the option of removing the retry on CaseConcurrencyExcpetion altogether...?

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/ReferenceKeyConcurrencyException.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmcts.ccd.endpoint.exceptions;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.CONFLICT)
+public class ReferenceKeyConcurrencyException extends ApiException {
```
MSancaktutar: Could we change the name ReferenceKeyConcurrencyException to something like, DuplicateReferenceKeyException? There doesn't have to be a concurrency matter involved in the case. So, concurrency word may be irrelevant.

src/main/java/uk/gov/hmcts/ccd/domain/service/createcase/SubmitCaseTransaction.java
```diff
@@ -28,10 +20,18 @@
...
 import uk.gov.hmcts.ccd.infrastructure.user.UserAuthorisation.AccessLevel;
 
+import javax.inject.Inject;
+import javax.transaction.Transactional;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+
+import static javax.transaction.Transactional.TxType.REQUIRES_NEW;
+import static uk.gov.hmcts.ccd.data.caseaccess.GlobalCaseRole.CREATOR;
+
```
MSancaktutar: imports seems to have changed in order. could we make sure we're using the standard code format activated to be applied on auto-save action?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -25,12 +24,12 @@
 import javax.persistence.PersistenceContext;
 import javax.persistence.PersistenceException;
 import javax.persistence.Query;
-
-import org.hibernate.StaleObjectStateException;
-import org.hibernate.exception.ConstraintViolationException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Qualifier;
+import java.time.LocalDateTime;
```
MSancaktutar: imports seems to have changed in order. could we make sure we're using the standard code format activated to be applied on auto-save action?

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/ReferenceKeyConcurrencyException.java
rafalkalita: Changed to ReferenceKeyUniqueConstraintException

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -75,9 +76,16 @@ public CaseDetails set(final CaseDetails caseDetails) {
...
+
+            if (e.getCause() instanceof ConstraintViolationException
+                && ((ConstraintViolationException) e.getCause()).getConstraintName()
+                .equals(UNIQUE_REFERENCE_KEY_CONSTRAINT)) {
+
+                LOG.warn("ConstraintViolationException happen for UUID={}. ConstraintName: {}",
+                    caseDetails.getReference(), UNIQUE_REFERENCE_KEY_CONSTRAINT);
+                throw new ReferenceKeyConcurrencyException(e.getMessage());
             }
+            throw new CaseConcurrencyException(e.getMessage());
```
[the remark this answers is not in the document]
rafalkalita: Introduced CasePersistenceException with the same 409 mapping.

src/main/java/uk/gov/hmcts/ccd/domain/service/createcase/SubmitCaseTransaction.java
rafalkalita: Imports are fine. Ordered by the formatter.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
rafalkalita: Imports are fine. Ordered by the formatter.

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/CasePersistenceException.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmcts.ccd.endpoint.exceptions;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.CONFLICT)
```
mario-paniccia: perhaps we can find a better error code than 409 for a PersistenceException?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -75,9 +77,16 @@ public CaseDetails set(final CaseDetails caseDetails) {
             throw new CaseConcurrencyException("The case data has been altered outside of this transaction.");
         } catch (PersistenceException e) {
             LOG.warn("Failed to store case details", e);
-            if (e.getCause() instanceof ConstraintViolationException) {
-                throw new CaseConcurrencyException(e.getMessage());
+
+            if (e.getCause() instanceof ConstraintViolationException
+                && ((ConstraintViolationException) e.getCause()).getConstraintName()
+                .equals(UNIQUE_REFERENCE_KEY_CONSTRAINT)) {
```
mario-paniccia: for readability can you move this into a private method `isDuplicateReference(e)` or similar please

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -75,9 +77,16 @@ public CaseDetails set(final CaseDetails caseDetails) {
...
+
+                LOG.warn("ConstraintViolationException happen for UUID={}. ConstraintName: {}",
```
mario-paniccia: I like this detailed exception, but we are already logging at line 79 a more generic one. Logging an exception twice is bad practice. Can you refactor so that we log this type of exception only once?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java
```diff
@@ -75,9 +77,16 @@ public CaseDetails set(final CaseDetails caseDetails) {
...
+                    caseDetails.getReference(), UNIQUE_REFERENCE_KEY_CONSTRAINT);
+                throw new ReferenceKeyUniqueConstraintException(e.getMessage());
             }
+            throw new CasePersistenceException(e.getMessage());
```
mario-paniccia: note a change of behaviour here. Before we were just logging non `ConstraintViolationException` PersistenceExceptions, now we are throwing them. But it seems better throwing them to me than swallowing them so 👍

MSancaktutar: We can throw new Exceptions with the original one kept in the "cause".

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/CasePersistenceException.java
MSancaktutar: 412 could be another option, but I think 409 is the best in the case of version conflict in opt locking.

PR #580, opened 2019-10-18
RDM-6031 - added xui_webapp to the list of data-store s2s authorised services

smathangi:


charts/ccd-data-store-api/values.yaml
```diff
@@ -24,7 +24,7 @@ java:
 
     CCD_DRAFT_TTL_DAYS: 180
 
-    DATA_STORE_S2S_AUTHORISED_SERVICES: ccd_data,ccd_gw,ccd_ps,probate_backend,divorce_ccd_submission,sscs,sscs_bulkscan,cmc,cmc_claim_store,cmc_claim_external_api,jui_webapp,pui_webapp,bulk_scan_orchestrator,fpl_case_service,iac,finrem_ccd_data_migrator,finrem_case_orchestration,employment_tribunals,ethos_repl_service,ccpay_bubble,ctsc_work_allocation,em_ccd_orchestrator
+    DATA_STORE_S2S_AUTHORISED_SERVICES: ccd_data,ccd_gw,ccd_ps,probate_backend,divorce_ccd_submission,sscs,sscs_bulkscan,cmc,cmc_claim_store,cmc_claim_external_api,jui_webapp,pui_webapp,bulk_scan_orchestrator,fpl_case_service,iac,finrem_ccd_data_migrator,finrem_case_orchestration,employment_tribunals,ethos_repl_service,ccpay_bubble,ctsc_work_allocation,em_ccd_orchestrator,xui_webpp
```
MSancaktutar: Is there a spelling issue in "xui_webpp"?

dependency-check-suppressions.xml
```diff
@@ -191,10 +191,11 @@
             for an attacker to gain access to response header containing sensitive data from another user.</notes>
         <vulnerabilityName>CVE-2019-7614</vulnerabilityName>
     </suppress>
-    <suppress>
+    <suppress until="2019-10-31">
         <notes><![CDATA[ file name: jackson-databind-2.9.10.jar ]]></notes>
         <packageUrl regex="true">^pkg:maven/com\.fasterxml\.jackson\.core/jackson\-databind@.*$</packageUrl>
         <cve>CVE-2019-16942</cve>
         <cve>CVE-2019-16943</cve>
+        <cve>CVE-2019-17531</cve>
```
MSancaktutar: We have to open a proper ticket to resolve this CVE and put it into the current sprint, before we temporarily suppress a CVE. And the suppression should ideally contain the Jira ticket number to let the code reviewer/observer be able to track the actual resolution activity (plan, progress, etc.).

smathangi: Build will fail anyway on 31st October unless we bump to new Jackson data bind release. I feel this way is easy to track and upgrade.

charts/ccd-data-store-api/values.yaml
smathangi: just copied what's there in the ticket. But very good spot. Thanks!

dependency-check-suppressions.xml
MSancaktutar: I see, but we should immediately initiate the path to resolution the first time we see any new CVE flag. We shouldn't let any new piece of information about a security matter live without a due process triggered after it's first observation. I do believe this is part of the best and most disciplined treatment of CVE flags.

PR #581, opened 2019-10-18
Fetch case events directly from case reference

banderous:

https://tools.hmcts.net/jira/browse/RDM-5550

### Change description ###

This is an optimisation to retrieve case events by case reference with
a single query, instead of first fetching a case and then issuing
another query to fetch its events.

This change performs an inner join between case_data and case_event
that can be executed efficiently by postgres using our existing indexes;
 in my test DB with 200k cases each with 10 events, fetching events for
 a case took a couple of milliseconds.

```
[ ] Yes
[x] No
```


banderous:
Closing in favor of https://github.com/hmcts/ccd-data-store-api/pull/586

PR #591, opened 2019-10-25
Fix N+1 query in CreatorSearchOperation

banderous:
https://tools.hmcts.net/jira/browse/RDM-6188

### Change description ###
The filtering in CreatorSearchOperation was issuing the same sql query  for every record in the results set, so up to 25 queries in total.

    This change ensures the query only happens once. This is a tactical fix,  a better long term strategy would be to push this filtering to the initial database query and avoid the need for additional queries at all.

I also deleted the test of the filtering in CreatorSearchOperationTest since this filtering is delegated to CaseAccessService (where it is now tested there).

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/search/CreatorSearchOperation.java
```diff
@@ -38,8 +37,6 @@ public CreatorSearchOperation(@Qualifier("authorised") SearchOperation searchOpe
             return Lists.newArrayList();
         }
 
-        return results.stream()
-            .filter(caseAccessService::canUserAccess)
-            .collect(Collectors.toList());
+        return caseAccessService.filterByUserAccess(results);
```
mario-paniccia: this is a very good improvement. However this solution still retrieves from the db potentially much more cases than needed. Because filtering based on user access happens in a second stage. I was wondering whether we could improve things even further by modifying the criteria passed to execute the very first query: 
```        
final List<CaseDetails> results = searchOperation.execute(metaData, criteria);
```
For users with restricted access to cases, we could as first thing check the list of case the user has access to, e.g.
```
 final List<Long> grantedCases = caseUserRepository.findCasesUserIdHasAccessTo(userRepository.getUserId());
```
Then enhance the criteria mechanism so that we can restrict a search on a range of cases only, in this example the `grantedCases`. 
And then we execute:
```
searchOperation.execute(metaData, criteriaWithRestrictedCases);
```

banderous:
@mario-paniccia Just written up the commit message explaining that the CreatorSearchOperation is redundant:

The CreatorSearchOperation is intended to restrict a search to only
    return cases the user has been granted access to in case_users, however
    this filtering is already done by DefaultCaseDetailsRepository so the
    CreatorSearchOperation is redundant.

    DefaultSearchOperation
      CachedCaseDetailsRespository
        DefaultCaseDetailsRepository
         -> SearchQueryFactoryOperation.build()
                 // Same filter logic as CreatorSearchOperation
              -> addUserCaseAccessClause()
         -> // Count queries also filtered by CaseDetailsQueryBuilderFactory

We already have tests for the DefaultSearchRepository that check it filters cases by access granted according to the user's role

PR #595, opened 2019-10-30
RDM-6116- Revert "Document links detection / restore program"

smathangi:
Reverts hmcts/ccd-data-store-api#548

Doclink recovery code is isolation to main ccd functional code.


build.gradle
```diff
@@ -486,9 +486,8 @@ jacoco {
 
 sonarqube {
```
mario-paniccia: @smathangi is it same to revert these changes or it's something you want to keep?

mario-paniccia: ah ok ignore this, that's the doc restore endpoint so we should remove it 👍

PR #599, opened 2019-11-04
Update dependencies to address CVEs

banderous:
https://tools.hmcts.net/jira/browse/RDM-6331

### Change description ###
Bump versions on multiple dependencies to address CVEs.

Additionally removes redundant suppressions.

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -224,7 +226,7 @@ dependencies {
     compile group: 'org.awaitility', name: 'awaitility', version: '3.1.6'
 
     // CVE-2019-12086 - force update of jackson-databind - remove when dependencies move off <2.9.9
-    compile "com.fasterxml.jackson.core:jackson-databind:2.9.10"
+    compile "com.fasterxml.jackson.core:jackson-databind:2.9.10.1"
```
smathangi: may be bump to latest version 2.10.0?

build.gradle
```diff
@@ -287,10 +292,7 @@ dependencyManagement {
 configurations.all {
     resolutionStrategy {
         eachDependency { DependencyResolveDetails details ->
-            if (details.requested.group in ['com.fasterxml.jackson.core', 'com.fasterxml.jackson.module', 'com.fasterxml.jackson.datatype']) {
-                details.useVersion '2.9.10'
-            }
-            if (details.requested.name == 'jackson-databind') {
+            if (details.requested.group in ['com.fasterxml.jackson.module', 'com.fasterxml.jackson.datatype']) {
```
smathangi: do we still need this config..? May be bump jackson.datatype ?

banderous: Good spot, I have removed this whole section.

build.gradle
```diff
@@ -258,7 +258,7 @@ dependencies {
     compile(group: 'io.searchbox', name: 'jest', version: '6.3.1') {
         exclude group: 'org.apache.logging.log4j', module: 'log4j-core'
     }
-    compile(group: 'org.elasticsearch', name: 'elasticsearch', version: '6.4.2') {
+    compile(group: 'org.elasticsearch', name: 'elasticsearch', version: '6.8.4') {
```
mario-paniccia: Alex do we need to bump up this? The reason why I'm asking is that the ES backend we use is on version `6.4.2` so ideally it would be better to use a client of corresponding version

banderous: As per the other PR, there are CVEs for 6.4.2; 6.8.2 was the lowest unaffected version I could find.

build.gradle
banderous: I'm advised to keep the version delta as small as possible to minimise the scope of regression testing required whilst resolving outstanding CVEs - we should tackle this in a follow up.

PR #623, opened 2019-11-18
Rdm 4324 DisplayOrder required for ComplexType elements

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4324

### Change description ###
DisplayOrder required for ComplexType elements

This ticket means no internal ordering of compound case fields is required in consumers like case ui toolkit (hmcts/ccd-case-ui-toolkit#472).

Additionally order that is taken from `DisplayOrder` column on `ComplexTypes` tab can be overwritten by `FieldDisplayOrder` column on `EventToComplexTypes` tab. If At least one `FieldDisplayOrder` value is present the ordering will be derived from `EventToComplexTypes` tab. As not all members of a complex field have to be defined on `EventToComplexTypes` the remaining fields without `FieldDisplayOrder` will be added in the encounter order (as defined in spreadsheet). Note that all members of a complex type that are returned from def store will always have value for `DisplayOrder` column on `ComplexTypes` therefore it is not necessary to handle any missing entries in this case (definition store is adding validation for `DisplayOrder` column to have order values for all members of the complex type and also ordering of these values on retrieval: https://github.com/hmcts/ccd-definition-store-api/pull/473)

Finally the WizardPageComplexFieldOverride is cleaned up from order as that is already handled by above.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CompoundFieldOrderService.java
```diff
@@ -0,0 +1,146 @@
...
+ * 1) If caseEventComplexFields is non empty array it will sort on its elements order values.
+ *      (EventToComplexTypes has at least one value for DisplayOrder column defined it will sort according to this column)
+ * 2) If caseEventComplexFields is empty array then it will sort caseFields' children order values.
+ *      (ComplexTypes tab's DisplayOrder column)
+ *
+ * Additionally it sorts the field type's FixedListItem list of case field of FixedList, FixedRadioList and MultiSelectList types.
+ */
+@Named
+@Singleton
+public class CompoundFieldOrderService {
```
andrewfolga: I will try to move it to definition store now.

andrewfolga: Done now - the compound case field complex fields ordering as well as fixed list items ordering are done in def store now.

src/test/java/uk/gov/hmcts/ccd/domain/model/aggregated/CompoundFieldOrderServiceTest.java
```diff
@@ -0,0 +1,270 @@
...
+            .withFieldType(simpleType())
+            .withOrder(order)
+            .build();
+    }
+
+    private FieldType simpleType() {
+        return aFieldType().withType(TEXT_TYPE).build();
+    }
+
+    private FieldType complexType() {
```
rafalkalita: Not used

PR #636, opened 2019-11-21
Adding back domain validation

mrganeshraja:
https://tools.hmcts.net/jira/browse/RDM-6298

### Change description ###
Adding back the domain validtion on document links.

```
[x] Yes
```


src/test/java/uk/gov/hmcts/ccd/domain/types/DocumentValidatorTest.java
```diff
@@ -53,10 +55,41 @@ public static void setUpClass() {
...
+        when(ap.getValidDMDomain()).thenReturn(url);
+        return new DocumentValidator(ap);
+    }
+
     @Before
     public void setUp() throws Exception {
-        validator = new DocumentValidator();
-
+        final ApplicationParams applicationParams = mock(ApplicationParams.class);
+//        when(applicationParams.getValidDMDomain()).thenReturn("https://dm.reform.hmcts.net");
```
banderous: Why is this commented out?

src/main/java/uk/gov/hmcts/ccd/domain/types/DocumentValidator.java
```diff
@@ -54,7 +61,7 @@ public BaseType getType() {
         }
 
         final String documentUrlValue = documentUrl.textValue();
-        final String urlPatternString = "http.*/documents/[A-Za-z0-9-]+(?:/binary)?";
+        final String urlPatternString = applicationParams.getValidDMDomain() + "/documents/[A-Za-z0-9-]+(?:/binary)?";
```
banderous: Can we be sure this won't exclude valid domains?  Is there a way to check all existing domains in usage?

src/test/java/uk/gov/hmcts/ccd/domain/types/DocumentValidatorTest.java
mrganeshraja: Missed it, was testing out something locally 👍

src/main/java/uk/gov/hmcts/ccd/domain/types/DocumentValidator.java
mrganeshraja: I am only putting back the old code and unit test back in action. I think the existing unit tests should be sufficient to test the code. I have tested with my local docker setup to make sure the domain regex used works with exisiting domains in AAT and production.

rafalkalita: I suggest asking Mutlu for a review. He was working in this area: https://tools.hmcts.net/jira/browse/RDM-4146 https://tools.hmcts.net/jira/browse/RDM-4146 We don't want to introduce the bug again.

mrganeshraja: @MSancaktutar can also review this PR if he need too, look a his comments in the JIRA ticket.

MSancaktutar: Happy to bring back the validation.

src/test/java/uk/gov/hmcts/ccd/domain/types/DocumentValidatorTest.java
hemantt: @mrganeshraja are you uncommenting this out?

mrganeshraja:
> Only question I have is whether this would block any valid domains

It wouldn't block any valid domains as build a regex this that will not block a valid domain and we can test this in the demo and aat environments before deploying to production.

PR #647, opened 2019-11-26
Rdm 6447

kiran-yenigala-hmcts:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/stdapi/CallbackInvoker.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/stdapi/CallbackInvoker.java
@@ -218,9 +218,8 @@ public class CallbackInvoker {
                                     final CaseType caseType,
                                     final CaseDetails caseDetails,
                                     final Map<String, JsonNode> responseData) {
-        // TODO: Change this to caseAccessService.getCreateRoles();
         accessControlService.verifyCreateAccess(eventId, caseType,
-                                                caseAccessService.getUserRoles(),
+                                                caseAccessService.getCaseCreationRoles(),
                                                 MAPPER.convertValue(responseData, JsonNode.class));
         caseTypeService.validateData(responseData, caseType);
         caseDetails.setData(caseSanitiser.sanitise(caseType, responseData));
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/stdapi/CallbackInvoker.java
```diff
@@ -36,19 +35,27 @@
     private final CaseDataService caseDataService;
     private final CaseSanitiser caseSanitiser;
     private final SecurityValidationService securityValidationService;
+    private final AccessControlService accessControlService;
+    private final CaseAccessService caseAccessService;
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
mario-paniccia: we have some OM already configured that you could inject, have a look at: JacksonObjectMapperConfig

src/main/java/uk/gov/hmcts/ccd/domain/service/stdapi/CallbackInvoker.java
```diff
@@ -205,10 +215,14 @@ private boolean callbackResponseHasCaseAndDataClassification(CallbackResponse ca
...
 
-    private void validateAndSetData(final CaseType caseType,
+    private void validateAndSetData(final String eventId,
+                                    final CaseType caseType,
                                     final CaseDetails caseDetails,
                                     final Map<String, JsonNode> responseData) {
         caseTypeService.validateData(responseData, caseType);
+        accessControlService.verifyCreateAccess(eventId, caseType,
+            caseAccessService.getCaseCreationRoles(),
+            MAPPER.convertValue(responseData, JsonNode.class));
```
mario-paniccia: can we extract to a method called `verifyCRUD`

mario-paniccia: 1) AuthorisedCreateCaseOperation also does a `verifyReadAccess` other than a `verifyCreateAccess`. Shouldn't we add that here too?
2)The issue with this solution is that these callbacks methods are called not only on case creation. But also on case update. So we are changing the behaviour of case updates, which is probably something we shouldn't do.
3)We have to be careful here. When I raised this ticket I was kinda sure that we are already verify CRUD on callbacks. But it looks like that's not the case. So now we are adding CRUD validation. The problem with this is that this is a breaking change that might introduce some errors in Production

mario-paniccia:
it looks like we are missing some CRUD verifications on callbacks at the moment. This needs to be confirmed and fixed. We will do this outside of the Respondent Journey work. I've raised the following new ticket: https://tools.hmcts.net/jira/browse/RDM-6775

PR #661, opened 2019-11-29
RDM-5007 Fix the WizardPage fields displayContext override when no access

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-5007

### Change description ###
The code missed update of the displayContext in WizardPageFields when user has no access. Eg.:
      "wizard_page_fields": [
        {
          "case_field_id": "defendantFirstName",
          "order": 1,
          "page_column_no": 1,
          "display_context": "OPTIONAL", <--- THIS will get updated to READONLY
          "complex_field_overrides": []
        },

This is for the update event, when the case has been already previously created.

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
@@ -255,13 +255,10 @@ public class AccessControlService {
 
     private void setReadonlyOnWizardPageFieldDisplayContext(List<WizardPage> wizardPages,
                                                             String caseViewFieldId) {
-        wizardPages.forEach(wizardPage -> {
-            wizardPage.getWizardPageFields().stream()
-                .filter(wizardPageField -> wizardPageField.getCaseFieldId().equals(caseViewFieldId)).collect(toList())
-                .forEach(wizardPageField -> {
-                    wizardPageField.setDisplayContext(READONLY);
-                });
-        });
+       wizardPages.forEach(wizardPage -> {
+           Optional<WizardPageField> wizardPageField = wizardPage.getWizardPageField(caseViewFieldId);
+           wizardPageField.ifPresent(field -> field.setDisplayContext(READONLY));
+       });
     }
 
     private void setChildrenAsReadOnlyIfNoAccess(final List<WizardPage> wizardPages, final String rootFieldId, final CaseField caseField, final Predicate<AccessControlList> access, final Set<String> userRoles, final CommonField caseViewField) {
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
@@ -1,6 +1,33 @@
 package uk.gov.hmcts.ccd.domain.service.common;
 
-import java.util.*;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.ccd.domain.model.aggregated.CaseEventTrigger;
+import uk.gov.hmcts.ccd.domain.model.aggregated.CaseViewField;
+import uk.gov.hmcts.ccd.domain.model.aggregated.CaseViewTrigger;
+import uk.gov.hmcts.ccd.domain.model.aggregated.CommonField;
+import uk.gov.hmcts.ccd.domain.model.definition.AccessControlList;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseEvent;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseField;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseState;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseType;
+import uk.gov.hmcts.ccd.domain.model.definition.WizardPage;
+import uk.gov.hmcts.ccd.domain.model.definition.WizardPageComplexFieldOverride;
+import uk.gov.hmcts.ccd.domain.model.definition.WizardPageField;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceTest.java
```diff
@@ -2846,7 +2907,12 @@ void shouldSetReadonlyFlagIfRelevantAclGrantingAccessAndEventNameNotMatching() t
                 USER_ROLES,
                 CAN_UPDATE);
 
-            assertThat(eventTrigger.getCaseFields(), everyItem(hasProperty("displayContext", is(READONLY))));
+            assertAll(
+                () -> assertThat("CaseField displayContext should change to READONLY",
+                    eventTrigger.getCaseFields().get(0).getDisplayContext(), is(READONLY)),
+                () -> assertThat("WizardPage CaseField displayContext should change to READONLY",
+                    eventTrigger.getWizardPages().get(0).getWizardPageFields().get(0).getDisplayContext(), is(READONLY))
+                     );
```
Thor-tech-of-metal: the size of ventTrigger.getWizardPages() can be asserted before get(0)

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -228,17 +240,30 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
...
     }
 
+    private void setReadonlyOnWizardPageFieldDisplayContext(List<WizardPage> wizardPages,
+                                                            String caseViewFieldId) {
+        wizardPages.forEach(wizardPage -> {
+            wizardPage.getWizardPageFields().stream()
+                .filter(wizardPageField -> wizardPageField.getCaseFieldId().equals(caseViewFieldId)).collect(toList())
+                .forEach(wizardPageField -> {
+                    wizardPageField.setDisplayContext(READONLY);
+                });
```
kiran-yenigala-hmcts: update wizardPageField only if display context is not READONLY

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -228,17 +240,30 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
...
                 }
             });
         return caseEventTrigger;
     }
 
+    private void setReadonlyOnWizardPageFieldDisplayContext(List<WizardPage> wizardPages,
+                                                            String caseViewFieldId) {
+        wizardPages.forEach(wizardPage -> {
+            wizardPage.getWizardPageFields().stream()
+                .filter(wizardPageField -> wizardPageField.getCaseFieldId().equals(caseViewFieldId)).collect(toList())
```
mario-paniccia: we could introduce a method on wizardPage called `getField` and move this code there

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -228,17 +240,30 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
                     CaseField field = caseFieldOpt.get();
                     if (!hasAccessControlList(userRoles, access, field.getAccessControlLists())) {
                         caseViewField.setDisplayContext(READONLY);
+                        setReadonlyOnWizardPageFieldDisplayContext(caseEventTrigger.getWizardPages(), caseViewField.getId());
                     }
                     if (field.isCompound()) {
```
mario-paniccia: don't we need to apply the same logic for compounds fields?

rafalkalita: It is covered already. Please check the unit tests. For compound fields it will change the displayContext in the "complex_field_overrides": []. This one is for top level displayContext

mario-paniccia: 👍

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -228,17 +240,27 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
...
                 }
             });
         return caseEventTrigger;
     }
 
+    private void setReadonlyOnWizardPageFieldDisplayContext(List<WizardPage> wizardPages,
+                                                            String caseViewFieldId) {
+       wizardPages.forEach(wizardPage -> {
+           Optional<WizardPageField> wizardPageField = wizardPage.getWizardPageField(caseViewFieldId);
+           wizardPageField.ifPresent(field -> field.setDisplayContext(READONLY));
```
mario-paniccia: This fix would work. But I think it's not the optimal fix. This is a regression on a feature that worked in the past. The bug is happening because recently we have done some changes (not sure if by mistake or voluntarily) to the display context of a field. 
In the original CCD design the display context of a field **for an event** is stored on the case field directly (the various`CaseField` classes in the toolkit and Data Store backend). 
When we remove update permission Data Store sets the display context of that event field as Readonly.
2 things changed recently:
1) a duplicate display context property has been added to the WizardPageField which is returned to the UI together inside an event field.
2) the UI has been changed to copy the WizardPageField display context to the event field, basically ignoring and overwriting the value Data Store is returning.
But Data Store has not been updated to also change this additional display context on update permission removal. Data store is thereby returning to the UI two display context values not in sync. 
This fix solves the problem by making sure the 2 display contexts have the same value.
But in my opinion the correct fix would be to remove this duplication by keeping only 
one of them. Keeping these 2 display contexts is increasing the complexity of Data Store and the UI for no reason. And this bug is a proof. Probably safer to keep the one on the case field to minimise the impact of the change. 
This will make sure we avoid similar bugs in Data Store in the future because there's only one place that controls the display context of a field.

rafalkalita:
We have decided to go with the change in the ui-toolkit.

PR #664, opened 2019-11-29
RDM-6376- Functional test cases for "fetch a case for display" endpoint

aashishaggarwal123:
JIRA link (if applicable)
https://tools.hmcts.net/jira/browse/RDM-6376

Change description
Implement functional test cases for "fetch a case for display" endpoint

[ ] Yes
[x] No


src/aat/resources/features/F-027/F-027.feature
```diff
@@ -0,0 +1,58 @@
...
+
+  @S-038
+  Scenario: must return status 200 along with the case-view object successfully
+    Given a user with [an active profile in CCD]
+    And a case that has just been created as in [Standard_Full_Case]
+    When a request is prepared with appropriate values
+    And the request [contains as input a valid case-reference that exists in CCD]
+    And it is submitted to call the [Fetch a case for display for Case Worker] operation of [CCD Data Store]
+    Then a positive response is received
+    And the response [has the 200 return code]
```
MSancaktutar: Are we not using the case reference of the case object just created? If so, this fact not is reflected to scenario wording. Moreover, when we get the response, it should contain the details of the case just created, shouldn't it. If what I understand is correct, this scenario is not worded well to describe the relations with the case record just created.

src/aat/resources/features/F-027/F-027.feature
```diff
@@ -0,0 +1,58 @@
...
+    And the request [contains in input parameters a structurally valid but non-existing case-reference]
+    And it is submitted to call the [Fetch a case for display for Case Worker] operation of [CCD Data Store]
+    Then a negative response is received
+    And the response [has the 400 return code]
+    And the response has all other details as expected
+
+  @S-036
+  Scenario: must return appropriate negative response when request does not provide valid authentication credentials
+    Given a user with [an active profile in CCD]
+    And a case that has just been created as in [Standard_Full_Case]
```
MSancaktutar: why create a case before a call known to be going to fail? That is unnecessary and confusing.

src/aat/resources/features/F-027/F-027.feature
```diff
@@ -0,0 +1,58 @@
...
+    And the request [does not provide valid authentication credentials]
+    And it is submitted to call the [Fetch a case for display for Case Worker] operation of [CCD Data Store]
+    Then a negative response is received
+    And the response [has the 403 return code]
+    And the response has all other details as expected
+
+  @S-037
+  Scenario: must return appropriate negative response when request does not provide an authorized access
+    Given a user with [an active profile in CCD]
+    And a case that has just been created as in [Standard_Full_Case]
```
MSancaktutar: Why create a case before a call that will fail?

src/aat/resources/features/F-027/F-027.feature
```diff
@@ -0,0 +1,58 @@
...
+    And the request [does not provide authorised access to the operation]
+    And it is submitted to call the [Fetch a case for display for Case Worker] operation of [CCD Data Store]
+    Then a negative response is received
+    And the response [has the 403 return code]
+    And the response has all other details as expected
+
+  @S-034
+  Scenario: must return appropriate negative response for a user not having a profile in CCD
+    Given a user with [no profile in CCD]
+    And a case that has just been created as in [Standard_Full_Case]
```
MSancaktutar: again, this seems to be unnecessary

src/aat/resources/features/F-027/F-027.feature
aashishaggarwal123: It is already there on the line number 10. If required I can remove line no 12 completely.

src/aat/resources/features/F-027/F-027.feature
aashishaggarwal123: It is kept for the purpose that expect the scenario related spec rest everything is correct and correct case reference is also part of this otherwise that spec and prepared data will be misleading.

aashishaggarwal123: It is kept for the purpose that request contains everything apart from the specified scenario otherwise prepared data and spec would be misleading. however I can remove that if you want.

src/aat/resources/features/F-027/F-027.feature
aashishaggarwal123: case creation removed.

src/aat/resources/features/F-027/F-027.feature
aashishaggarwal123: case creation removed.

src/aat/resources/features/F-027/F-027.feature
aashishaggarwal123: case creation removed.

src/aat/resources/features/F-027/F-027.feature
aashishaggarwal123: Updated

MSancaktutar: we should relate the case just created with the request being made

PR #681, opened 2019-12-06
Rdm 6758

kiran-yenigala-hmcts:
JIRA link (if applicable)
https://tools.hmcts.net/jira/browse/RDM-6758

When using case role case history doesn't show case events

[ ] Yes
[x] No

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
@@ -72,6 +72,13 @@ public class CaseAccessService {
         return new HashSet<>(caseUserRepository.findCaseRoles(Long.valueOf(caseId), userRepository.getUserId()));
     }
 
+    public Set<String> getAccessRoles(String caseId) {
+        if (caseId == null || caseId.trim().length() == 0) {
+            return userRepository.getUserRoles();
+        }
+        return Sets.union(userRepository.getUserRoles(), getCaseRoles(caseId));
+    }
+
     public Set<String> getCaseCreationCaseRoles() {
         return Collections.singleton(CREATOR.getRole());
     }
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
@@ -73,9 +73,6 @@ public class CaseAccessService {
     }
 
     public Set<String> getAccessRoles(String caseId) {
-        if (caseId == null || caseId.trim().length() == 0) {
-            return userRepository.getUserRoles();
-        }
         return Sets.union(userRepository.getUserRoles(), getCaseRoles(caseId));
     }
 
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
@@ -1,9 +1,16 @@
 package uk.gov.hmcts.ccd.domain.service.common;
 
 import com.google.common.collect.Sets;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.stereotype.Service;
+import uk.gov.hmcts.ccd.data.caseaccess.CachedCaseUserRepository;
 import uk.gov.hmcts.ccd.data.caseaccess.CaseUserRepository;
 import uk.gov.hmcts.ccd.data.user.CachedUserRepository;
 import uk.gov.hmcts.ccd.data.user.UserRepository;
@@ -11,9 +18,6 @@ import uk.gov.hmcts.ccd.domain.model.definition.CaseDetails;
 import uk.gov.hmcts.ccd.endpoint.exceptions.ValidationException;
 import uk.gov.hmcts.ccd.infrastructure.user.UserAuthorisation.AccessLevel;
 import uk.gov.hmcts.reform.auth.checker.spring.serviceanduser.ServiceAndUserDetails;
-
-import java.util.*;
-import java.util.regex.Pattern;
 
 import static uk.gov.hmcts.ccd.data.caseaccess.GlobalCaseRole.CREATOR;
 
@@ -37,7 +41,7 @@ public class CaseAccessService {
         = Pattern.compile(".+-solicitor$|.+-panelmember$|^citizen(-.*)?$|^letter-holder$|^caseworker-.+-localAuthority$");
 
     public CaseAccessService(@Qualifier(CachedUserRepository.QUALIFIER) UserRepository userRepository,
-                             CaseUserRepository caseUserRepository) {
+                             @Qualifier(CachedCaseUserRepository.QUALIFIER)  CaseUserRepository caseUserRepository) {
         this.userRepository = userRepository;
         this.caseUserRepository = caseUserRepository;
     }
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
@@ -1,23 +1,18 @@
 package uk.gov.hmcts.ccd.domain.service.getevents;
 
 import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
-import uk.gov.hmcts.ccd.data.caseaccess.CachedCaseUserRepository;
-import uk.gov.hmcts.ccd.data.caseaccess.CaseUserRepository;
 import uk.gov.hmcts.ccd.data.definition.CachedCaseDefinitionRepository;
 import uk.gov.hmcts.ccd.data.definition.CaseDefinitionRepository;
-import uk.gov.hmcts.ccd.data.user.CachedUserRepository;
-import uk.gov.hmcts.ccd.data.user.UserRepository;
 import uk.gov.hmcts.ccd.domain.model.definition.CaseDetails;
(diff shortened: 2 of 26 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
```diff
@@ -74,12 +85,19 @@ public AuthorisedGetEventsOperation(@Qualifier("classified") GetEventsOperation
...
-        if (userRoles == null || userRoles.isEmpty()) {
+        if (caseUserRoles == null || caseUserRoles.isEmpty()) {
             throw new ValidationException("Cannot find user roles for the user");
         }
 
-        return verifyReadAccess(events, userRoles, caseType);
+        return verifyReadAccess(events, caseUserRoles, caseType);
+    }
+
+    private Set<String> getUserRoles(String caseId) {
```
mario-paniccia: can you rename this to `getAccessRoles` since these are no longer only user roles but it's now  a union of user roles and case roles

mario-paniccia: we should also move this method to `CaseAccessService` to prevent duplicating the same logic in the future

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
```diff
@@ -74,12 +85,19 @@ public AuthorisedGetEventsOperation(@Qualifier("classified") GetEventsOperation
...
+        return Sets.union(userRepository.getUserRoles(),
+            caseUserRepository
```
mario-paniccia: Important: I've noticed that CaseUserRepository does not have a Cached version. Compare with UserRepository. This means every call to retrieve the case roles done **as part of the same endpoint call** will reach the db. Can we introduce a CachedCaseUserRepository?

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
```diff
@@ -74,12 +85,19 @@ public AuthorisedGetEventsOperation(@Qualifier("classified") GetEventsOperation
             throw new ValidationException("Cannot find case type definition for  " + caseTypeId);
         }
 
-        Set<String> userRoles = userRepository.getUserRoles();
-        if (userRoles == null || userRoles.isEmpty()) {
+        if (caseUserRoles == null || caseUserRoles.isEmpty()) {
```
mario-paniccia: shall we rename `caseUserRoles` to `accessRoles`?

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
```diff
@@ -28,30 +31,38 @@
...
-        return secureEvents(caseDetails.getCaseTypeId(), events);
+
+        return secureEvents(caseDetails.getCaseTypeId(), getUserRoles(caseDetails.getId()), events);
     }
 
     @Override
     public List<AuditEvent> getEvents(String jurisdiction, String caseTypeId, String caseReference) {
-        return secureEvents(caseTypeId, getEventsOperation.getEvents(jurisdiction, caseTypeId, caseReference));
+        List<AuditEvent> auditEvents = getEventsOperation.getEvents(jurisdiction, caseTypeId, caseReference);
+        Set<String> caseUserRoles = auditEvents != null && auditEvents.size() > 0
```
mario-paniccia: shall we rename caseUserRoles to accessRoles?

mario-paniccia: I think I preferred the previous version of the code where the access roles where retrieved inside the `secureEvents` itself. Much less duplication. 1 call to `getUserRoles` vs 3 now. And if we do that we could remove lines 62 -> 64 that seem a bit of a complication to me.

src/test/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseControllerCaseRolesIT.java
```diff
@@ -0,0 +1,120 @@
...
+
+        MockUtils.setSecurityAuthorities(UID_NO_EVENT_ACCESS, authentication, MockUtils.ROLE_CASEWORKER_PUBLIC);
+
+        mockMvc = MockMvcBuilders.webAppContextSetup(wac).build();
+        template = new JdbcTemplate(db);
+    }
+
+    @Test
+    @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = { "classpath:sql/insert_cases_event_access_case_roles.sql" })
+    public void shouldNotReturnEventDataForCitizenWhoHasNoAccessToEvents() throws Exception {
```
mario-paniccia: can you rename to `shouldNotReturnEventHistoryDataForCitizenWhoHasNoAccessToEvents`

src/test/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseControllerCaseRolesIT.java
```diff
@@ -0,0 +1,120 @@
...
+        String content = result.getResponse().getContentAsString();
+        assertNotNull("Content Should not be null", content);
+        UICaseViewResource savedCaseResource = mapper.readValue(content, UICaseViewResource.class);
+        assertNotNull("Saved Case Details should not be null", savedCaseResource);
+        assertEquals("Should not contain events with case role access", 1, savedCaseResource.getEvents().length);
+    }
+
+    @Test
+    @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = { "classpath:sql/insert_cases_event_access_case_roles.sql" })
+    public void shouldReturnEventDataForCitizenWhoHasCaseRoleAccess() throws Exception {
```
mario-paniccia: can you rename to `shouldReturnEventHistoryDataForCitizenWhoHasCaseRoleAccess`

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
kiran-yenigala-hmcts: We have to pass caseId to the secureEvents... method to get the caseroles.. is that fine....
Still we have to do lines 62-64 to get caseid for that method

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
kiran-yenigala-hmcts: Will look into this....

kiran-yenigala-hmcts: CachedCaseUserRepository has been introduced... Please review...

mario-paniccia: 👍

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
```diff
@@ -74,12 +85,19 @@ public AuthorisedGetEventsOperation(@Qualifier("classified") GetEventsOperation
...
         }
 
-        return verifyReadAccess(events, userRoles, caseType);
+        return verifyReadAccess(events, accessRoles, caseType);
+    }
+
+    private Set<String> getAccessRoles(String caseId) {
+        return Sets.union(userRepository.getUserRoles(),
+            caseUserRepository
+                .findCaseRoles(Long.valueOf(caseId), userRepository.getUserId())
```
mario-paniccia: we should add this logic to a new method in `CaseAccessService` alongside the similar methods we have there already

kiran-yenigala-hmcts: Ok... I am going to move this to CaseAccessService, I need to update test cases as well for this.. I am working on this

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
```diff
@@ -27,31 +25,31 @@
...
-
-        return secureEvents(caseDetails.getCaseTypeId(), events);
+        return secureEvents(caseDetails.getCaseTypeId(), caseDetails.getId(), events);
     }
 
     @Override
     public List<AuditEvent> getEvents(String jurisdiction, String caseTypeId, String caseReference) {
-        return secureEvents(caseTypeId, getEventsOperation.getEvents(jurisdiction, caseTypeId, caseReference));
+        List<AuditEvent> auditEvents = getEventsOperation.getEvents(jurisdiction, caseTypeId, caseReference);
+        String caseId = auditEvents != null && auditEvents.size() > 0 ? auditEvents.get(0).getCaseDataId() : "";
```
mario-paniccia: we could simplify this by considering that audit event's for a case cannot be null. There will be always at least on event in the case history of a case, the creation event. I'd say we can assume that's the case and remove these extra checks.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessService.java
```diff
@@ -68,6 +72,13 @@ public AccessLevel getAccessLevel(ServiceAndUserDetails serviceAndUserDetails) {
         return new HashSet<>(caseUserRepository.findCaseRoles(Long.valueOf(caseId), userRepository.getUserId()));
     }
 
+    public Set<String> getAccessRoles(String caseId) {
+        if (caseId == null || caseId.trim().length() == 0) {
+            return userRepository.getUserRoles();
+        }
```
mario-paniccia: we could simplify this by considering that audit event's for a case cannot be null. There will be always at least on event in the case history of a case, the creation event. I'd say we can assume that's the case and remove these extra checks.

kiran-yenigala-hmcts: Done

src/main/java/uk/gov/hmcts/ccd/domain/service/getevents/AuthorisedGetEventsOperation.java
kiran-yenigala-hmcts: Done

PR #684, opened 2019-12-09
RDM-6385: functional test cases for "Get case data with UI layout"

SivaKanukolluQA:
https://tools.hmcts.net/jira/browse/RDM-6385

### Change description ###
Creating functional test cases for "Get case data with UI layout" endpoint

[ ] Yes
[x] No

### DIFF
--- a/src/aat/resources/features/F-026/F-026.feature
+++ b/src/aat/resources/features/F-026/F-026.feature
@@ -47,7 +47,7 @@ Feature: F-026: Get a case data with UI layout
     And the response [has an HTTP-403 return code]
     And the response has all other details as expected
 
-  @S-076
+  @S-076 @Ignore #This scenario is not returning http 412 code. Jira: RDM-6879
   Scenario: must return 412 when the case type is not present in Definition store workbasket input fields
     Given a user with [an active profile in CCD]
     When a request is prepared with appropriate values
--- a/src/aat/resources/features/F-026/F-026.feature
+++ b/src/aat/resources/features/F-026/F-026.feature
@@ -4,7 +4,7 @@ Feature: F-026: Get a case data with UI layout
   Background: Load test data for the scenario
     Given an appropriate test context as detailed in the test data source
 
-  @S-078 @Ignore
+  @S-078
   Scenario: must return the list of cases and status code 200 for correct inputs
     Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
     And a user with [an active profile in CCD]
--- a/src/aat/resources/features/F-026/F-026.feature
+++ b/src/aat/resources/features/F-026/F-026.feature
@@ -1,0 +1,17 @@
+@F-026
+Feature: F-026: Get a case data with UI layout
+
+  Background: Load test data for the scenario
+    Given an appropriate test context as detailed in the test data source
+
+  @S-078
+  Scenario: must return the list of cases and status code 200 for correct inputs
+    Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
+    And a user with [an active profile in CCD]
+    When a request is prepared with appropriate values
+    And the request [uses the case-reference of the case just created]
+    And it is submitted to call the [Get a case data with UI layout] operation of [CCD Data Store]
+    Then a positive response is received
+    And the response [contains the details of the case just created, along with an HTTP-200 OK]
+    And the response has all other details as expected
+
--- a/src/aat/resources/features/F-026/F-026.feature
+++ b/src/aat/resources/features/F-026/F-026.feature
@@ -9,7 +9,7 @@ Feature: F-026: Get a case data with UI layout
     Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
     And a user with [an active profile in CCD]
     When a request is prepared with appropriate values
-    And the request [uses the case-reference of the case just created]
+    And the request [uses case-reference of the case just created]
     And it is submitted to call the [Get a case data with UI layout] operation of [CCD Data Store]
     Then a positive response is received
     And the response [contains the details of the case just created, along with an HTTP-200 OK]
@@ -20,8 +20,8 @@ Feature: F-026: Get a case data with UI layout
     Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
     And a user with [an active profile in CCD]
     When a request is prepared with appropriate values
-    And the request [uses the case-reference of the case just created]
-    And the request [uses a case-type which is not in CCD]
+    And the request [uses case-reference of the case just created]
+    And the request [uses an existing case-type which doesn't have any associated cases]
     And it is submitted to call the [Get a case data with UI layout] operation of [CCD Data Store]
     Then a positive response is received
     And the response [contains an empty SearchResultView, along with an HTTP-200 OK]
@@ -51,7 +51,7 @@ Feature: F-026: Get a case data with UI layout
   Scenario: must return 412 when the case type is not present in Definition store workbasket input fields
     Given a user with [an active profile in CCD]
     When a request is prepared with appropriate values
-    And the request [the case type is not present in Definition store workbasket]
+    And the request [uses case type which is not present in Definition store workbasket]
     And it is submitted to call the [Get a case data with UI layout] operation of [CCD Data Store]
     Then a negative response is received
     And the response [has an HTTP-412 return code]
--- a/src/aat/resources/features/F-026/F-026.feature
+++ b/src/aat/resources/features/F-026/F-026.feature
@@ -4,7 +4,7 @@ Feature: F-026: Get a case data with UI layout
   Background: Load test data for the scenario
     Given an appropriate test context as detailed in the test data source
 
-  @S-078
+  @S-078 @Ignore
   Scenario: must return the list of cases and status code 200 for correct inputs
     Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
     And a user with [an active profile in CCD]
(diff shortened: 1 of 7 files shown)


src/aat/resources/features/F-026/S-074.td.json
```diff
@@ -0,0 +1,27 @@
...
+    "pathVariables": {
+      "uid": "InvalidUID"
+    }
+  },
+  "expectedResponse": {
+    "_extends_": "Common_403_Response",
+    "headers": {
+      "Content-Length": "[[ANY]]",
+      "Content-Encoding": "gzip",
+      "Content-Type": "application/json;charset=UTF-8"
```
aashishaggarwal123: You can move these 3 lines to the Test data base files as these are common in all the responses.

src/aat/resources/features/F-026/S-076.td.json
```diff
@@ -0,0 +1,27 @@
+{
+  "title": "must return appropriate negative response when request does not provide valid authentication credentials",
```
aashishaggarwal123: Title doesn't match with the scenario description in the feature file.

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    And the response [contains the details of the case just created, along with an HTTP-200 OK]
+    And the response has all other details as expected
+
+  @S-077
+  Scenario: must return an empty SearchResultView envelope and status code 200 if case type is not found
+    Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
+    And a user with [an active profile in CCD]
+    When a request is prepared with appropriate values
+    And the request [uses the case-reference of the case just created]
+    And the request [uses a case-type which is not in CCD]
```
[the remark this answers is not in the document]
SivaKanukolluQA: Updated:   And the request [uses an existing case-type which doesn't have any associated cases]

src/aat/resources/features/F-026/S-076.td.json
SivaKanukolluQA: Updated:   "title": "must return 412 when the case type is not present in Definition store workbasket input fields",

src/aat/resources/features/F-026/S-074.td.json
SivaKanukolluQA: Updated: in Test Data Base file
  "expectedResponse": {
    "headers": {
      "Content-Length": "[[ANY]]",
      "Content-Encoding": "gzip",
      "Content-Type": "application/json;charset=UTF-8"
    }

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    Then a positive response is received
+    And the response [contains the details of the case just created, along with an HTTP-200 OK]
+    And the response has all other details as expected
+
+  @S-077
+  Scenario: must return an empty SearchResultView envelope and status code 200 if case type is not found
+    Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
+    And a user with [an active profile in CCD]
+    When a request is prepared with appropriate values
+    And the request [uses the case-reference of the case just created]
```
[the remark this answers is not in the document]
SivaKanukolluQA: Updated.

shrivsksavit:
as per : Siva Kanukollu 3:33 PM
Hi Sumit, Mutlu and team agreed to stick with the scenarios listed in excel to deliver in sprint. There is another review round later to verify these scenarios against Swagger.

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
+@F-026
+Feature: F-026: Get a case data with UI layout
```
MSancaktutar: Feature title is wrong. With an 'a' inserted, it reads like an operation that returns just 1 case data. Starting from this title with just a single letter difference, the rest of the files became much more confusing to understand. I initially thought, like, if this is to get 'a' case data, where is the case id for it in the test dat files, etc.

MSancaktutar: Above comment preserved, the original title will be made more clear and self descriptive after a total review of original swagger documentation.

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+
+  Background: Load test data for the scenario
+    Given an appropriate test context as detailed in the test data source
+
+  @S-078
+  Scenario: must return the list of cases and status code 200 for correct inputs
+    Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
```
MSancaktutar: Why create a case at all? Totally non sense.

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    And a user with [an active profile in CCD]
+    When a request is prepared with appropriate values
+    And the request [uses case-reference of the case just created]
```
MSancaktutar: Why use the case reference of the case just created? And where is it actually used at all? No where...
This is promise to be met by the respective test data, which it indeed lists among spec, but there is no use of such case reference. Indeed, there shouldn't be, either.

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    And it is submitted to call the [Get a case data with UI layout] operation of [CCD Data Store]
+    Then a positive response is received
+    And the response [contains the details of the case just created, along with an HTTP-200 OK]
```
MSancaktutar: For God's sake, where is the check for the case just created?

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    And the response has all other details as expected
+
+  @S-077
+  Scenario: must return an empty SearchResultView envelope and status code 200 if case type is not found
+    Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
```
MSancaktutar: None sense. Why create new case at all?

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    And a user with [an active profile in CCD]
+    When a request is prepared with appropriate values
+    And the request [uses case-reference of the case just created]
```
MSancaktutar: None sense.

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    And it is submitted to call the [Get a case data with UI layout] operation of [CCD Data Store]
+    Then a positive response is received
+    And the response [contains the details of the case just created, along with an HTTP-200 OK]
+    And the response has all other details as expected
+
+  @S-077
+  Scenario: must return an empty SearchResultView envelope and status code 200 if case type is not found
```
MSancaktutar: If case type is not found, how can we return a 200? Isn't it already so obvious that it musn't return a 200 if case type doesn't exist?

src/aat/resources/features/F-026/F-026.feature
```diff
@@ -0,0 +1,58 @@
...
+    Given a case that has just been created as in [Standard_Full_Case_Creation_Data]
+    And a user with [an active profile in CCD]
+    When a request is prepared with appropriate values
+    And the request [uses case-reference of the case just created]
+    And the request [uses an existing case-type which doesn't have any associated cases]
```
MSancaktutar: This is inconsistent with the scenario title.
The case for a case type with no cases under it could be a separate test scenario. But that would require a dedicated case type with no cases ever to be created under it.

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: Corrected this typo and copy paste issue.
Feature: F-026: Get case data with UI layout

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: My assumption: A new case created, because when these tests run on any environment, it should have at least one case to return.
However as per your explanation, assuming cases are already available in our environments, removed this case creation step from tests.

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: Updated: this is not necessary. removed.

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: Updated:   And the response [contains details of existing cases associated, along with an HTTP-200 OK]

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: Removed.

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: Removed.

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: Updated: 
  Scenario: must return an empty SearchResultView envelope and status code 200 if case type has no associated cases

SivaKanukolluQA: There is another scenario (S-076) for case type is not found. 
So updated the title for S-077.

src/aat/resources/features/F-026/F-026.feature
SivaKanukolluQA: Created new case type which does't have any cases
  "ctid": "DO_NOT_CREATE_CASES"

PR #703, opened 2019-12-13
RDM-6719: Add Banner Feature to CCD UI

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-6719

### Change description ###

Display Banner on CCD UI for each Jurisdiction configured.

```
[ ] Yes
[ X] No

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/JurisdictionMapper.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/JurisdictionMapper.java
@@ -14,7 +14,6 @@ public class JurisdictionMapper {
         result.setName(jurisdiction.getName());
         result.setDescription(jurisdiction.getDescription());
         result.setCaseTypes(jurisdiction.getCaseTypes());
-        result.setBanners(jurisdiction.getBanners());
         return result;
     }
 }
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/JurisdictionMapper.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/JurisdictionMapper.java
@@ -1,10 +1,9 @@
 package uk.gov.hmcts.ccd.data.casedetails;
 
-import uk.gov.hmcts.ccd.domain.model.aggregated.JurisdictionDisplayProperties;
-import uk.gov.hmcts.ccd.domain.model.definition.Jurisdiction;
-
 import javax.inject.Named;
 import javax.inject.Singleton;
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: src/main/java/uk/gov/hmcts/ccd/data/casedetails/JurisdictionMapper.java [hunk 1]
```diff
@@ -15,6 +15,10 @@ public JurisdictionDisplayProperties toResponse(Jurisdiction jurisdiction) {
         result.setName(jurisdiction.getName());
```
mario-paniccia: would it be possible to properly model the banner in Object Oriented terms with a Banner object?
Furthermore, but this is my personal opinion, I wouldn't have added a banner directly into a Jurisdiction. I would have had reverted the association, i.e. kept it as a separate object and have each Banner refer to a JurisdictionId.

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UIDefinitionController.java
```diff
@@ -95,4 +106,35 @@ public UIDefinitionController(@Qualifier(AuthorisedGetCriteriaOperation.QUALIFIE
...
+    @ApiResponses({
+        @ApiResponse(
+            code = 200,
+            message = "Success",
+            response = UIBannerResource.class
+        ),
+        @ApiResponse(
+            code = 404,
+            message = "Banners not found"
+        )
```
MSancaktutar: For better design, an empty list of banners should be returned with 200, rather than a 404.

MSancaktutar: A 404 is meant to designate a case where a client refers to a resource by mistake, and it is client's fault to refer to that. It would be the case if the client was to ask for, say, banner object of b1, which wasn't present. However, when the user asks a list of resources, the list may be an empty list, and still the query is valid. A 4XX response would be wrong.

kiran-yenigala-hmcts: Code is fine, it is returning empty banners result if banners not available for the passed jurisdiction references. only the documentation need to be removed 

@ApiResponse(
            code = 404,
            message = "Banners not found"
        )

src/main/java/uk/gov/hmcts/ccd/data/casedetails/JurisdictionMapper.java
kiran-yenigala-hmcts: Done changes

PR #729, opened 2020-01-07
RDM-7070 Functional test for get printable documents (AM operation)

bmcts:

https://tools.hmcts.net/jira/browse/AM-436

### Change description ###

Added scenario `must return successfully all and only the respective documents with read access for solicitors`

```
[ ] Yes
[x] No
```


src/aat/resources/features/F-032/F-032.feature
```diff
@@ -55,3 +55,19 @@ Feature: F-032: Get printable documents
...
+  @S-123
+  Scenario: must return successfully all and only the respective documents with read access for solicitors
+    Given a user with [an active solicitor 1 profile in CCD with read permissions on limited documents for a given case type]
+    And a user with [another active solicitor 2 profile in CCD with all permissions on all documents for a given case type]
+    And a user with [another active solicitor 3 profile in CCD with read permissions on different limited documents for a given case type]
+    And a successful call [to create a token for case creation] as in [Befta_Jurisdiction2_Default_Token_Creation_Data_For_Case_Creation]
+    And another successful call [by a privileged user with full ACL to create a case of this case type] as in [Befta_Jurisdiction2_Default_Full_Case_Creation_Data]
+    When a request is prepared with appropriate values
+    And the request [is made by solicitor 2 who just created the case]
+    And it is submitted to call the [get printable documents] operation of [CCD Data Store]
```
dharmendra8186: get printable documents operation returning one document while expectation is 7 documents.Due to this Scenario is failing.Please have a look in CallbackEndPoint Controller to modified the document printable logic.

bmcts: There is only one endpoint we are testing in this feature file and it is hitting that one. The test is failing because of issues with the implementation of the endpoint, but fixing that is out of scope for this PR

bmcts: After further investigation, the single document URL returned by the endpoint is correct and the test data has been updated to reflect this

PR #745, opened 2020-01-14
RDM-6757 Populate display_context_parameter for collections using case role

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-6757

### Change description ###
The main problem was the user profile on the front-end has no case roles and we cannot use them to show/hide collection Add new/Remove buttons.

We have decided to put this information in the display_context_parameter:
#COLLECTION(allowInsert, allowDelete) - should show both Add New and Remove buttons
#COLLECTION(allowInsert) - should show Add New and hide Remove button
#COLLECTION(allowDelete) - should show Remove and hide Add New button
#COLLECTION() - should hide both Add New and Remove buttons

There will be a Front-End change to consume the display_context_parameter.

```
[ ] Yes
[ X] No
```


src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test
+    @DisplayName("should update parameter)")
+    void shouldUpdateParameter() {
+        assertAll(
+            () -> assertThat(updateDisplayContextParameter("#COLLECTION(allowInsert,allowDelete)", COLLECTION, asList("allowInsert", "allowDelete")),
+                is("#COLLECTION(allowDelete,allowInsert)")),
+            () -> assertThat(updateDisplayContextParameter("#COLLECTION(allowInsert,allowDelete)", COLLECTION, asList("allowDelete")),
+                is("#COLLECTION(allowDelete)")),
+            () -> assertThat(updateDisplayContextParameter("#COLLECTION()", COLLECTION, asList("allowInsert")),
+                is("#COLLECTION(allowInsert)")),
```
mario-paniccia: can you add a scenario:
`"#COLLECTION("someOtherOption")", COLLECTION, asList("allowInsert")`
and verify that the end result is 
`"#COLLECTION("someOtherOption","allowInsert" )"`
we don't want "someOtherOption" to be wiped out, because it could be some other display option for collections we introduce in the future

mario-paniccia: can you add a scenario for:
"#COLLECTION(allowInsert,allowDelete)", COLLECTION, asList("allowDelete", "allowInsert")

src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
```diff
@@ -0,0 +1,82 @@
...
+            () -> assertThat(updateDisplayContextParameter("#TABLE(param1,param2)", TABLE, asList("postcode")),
```
mario-paniccia: not sure this is correct. Data store should not interfere or override any TABLE option set on the definition. There's no reason to do that. I think we can restrict Data Store action only on the COLLECTION option and ignore all the rest

src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
```diff
@@ -0,0 +1,82 @@
...
+        assertAll(
+            () -> assertThat(updateDisplayContextParameter("#TABLE()", COLLECTION, asList("allowInsert", "allowDelete")),
+                is("#COLLECTION(allowDelete,allowInsert),#TABLE()")),
+            () -> assertThat(updateDisplayContextParameter("#TABLE(),#COLLECTION(allowInsert,allowDelete)", COLLECTION, asList("allowDelete")),
+                is("#COLLECTION(allowDelete),#TABLE()")),
+            () -> assertThat(updateDisplayContextParameter("#TABLE(),#COLLECTION(allowInsert,allowDelete)", TABLE, asList("postcode")),
+                is("#COLLECTION(allowDelete,allowInsert),#TABLE(postcode)")),
+            () -> assertThat(updateDisplayContextParameter("#TABLE(postcode),#COLLECTION(allowInsert,allowDelete)", COLLECTION, asList("other")),
+                is("#COLLECTION(other),#TABLE(postcode)"))
+                 );
```
mario-paniccia: mmm not sure about this but it seems to me these tests are describing a feature we don't provide. We only allow overriding of "allowDelete, allowInsert" of COLLECTION, so adding tests for all the rest seems a bit misleading to me. But we can leave them

src/main/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtil.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.joining;
+
+public class DisplayContextParameterUtil {
+    public enum Parameter {
+        COLLECTION("#COLLECTION"),
+        TABLE("#TABLE");
```
mario-paniccia: don't think we need to define anything else other than COLLECTION. That's all Data Store should concern with

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -239,6 +252,46 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
         return caseEventTrigger;
     }
 
+    public CaseEventTrigger updateCollectionDisplayContextParameterByAccess(final CaseEventTrigger caseEventTrigger,
+                                                                            final Set<String> userRoles) {
+        caseEventTrigger.getCaseFields()
```
mario-paniccia: consider one of these alternative options:
1) introduce a method on CaseEventTrigger called getCollectionFields
2) caseEventTrigger.getCaseFields().filter(caseViewField.isCollectionFieldType())....

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -239,6 +252,46 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
...
+        if (hasAccessControlList(userRoles, CAN_DELETE, field.getAccessControlLists())) {
+            collectionAccess.add("allowDelete");
+        }
+        if (hasAccessControlList(userRoles, CAN_UPDATE, field.getAccessControlLists())) {
+            collectionAccess.add("allowInsert");
+            collectionAccess.add("allowDelete");
+        }
+        return DisplayContextParameterUtil.updateDisplayContextParameter(field.getDisplayContextParameter(),
+            DisplayContextParameterUtil.Parameter.COLLECTION, collectionAccess);
+    }
```
mario-paniccia: I would move this method to DisplayContextParameterUtil so it know's about `"allowInsert"` and `"allowDelete" rather than AccessControlService

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -239,6 +252,46 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
...
+                setChildrenCollectionDisplayContextParameter(caseViewField.getFieldType().getChildren(), userRoles);
+            });
+        return caseEventTrigger;
+    }
+
+    private void setChildrenCollectionDisplayContextParameter(final List<CaseField> caseFields,
+                                                              final Set<String> userRoles) {
+        caseFields.forEach(childField -> {
+            if (childField.isCollectionFieldType()) {
+                childField.setDisplayContextParameter(getCollectionAccess(userRoles, childField));
```
mario-paniccia: I would rename `getCollectionAccess` to `calculateDisplayContextParamer`

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -239,6 +252,46 @@ public CaseEventTrigger setReadOnlyOnCaseViewFieldsIfNoAccess(final CaseEventTri
...
+            }
+
+            setChildrenCollectionDisplayContextParameter(childField.getFieldType().getChildren(), userRoles);
+        });
+    }
+
+    private String getCollectionAccess(Set<String> userRoles, CommonField field) {
+        List<String> collectionAccess = new ArrayList<>();
+        if (hasAccessControlList(userRoles, CAN_CREATE, field.getAccessControlLists())) {
```
mario-paniccia: we need to be careful here. Strictly speaking this ticket should only have effect in case of case roles. But because userRoles also contains idam roles, when EUI implements the UI part of this ticket, it's possible the UI behaviour will change also for permissions set on Idam roles...

mario-paniccia: what I mean is this code adds COLLECTION(allowInsert) if there is an idam role with C permission, even if there's no case roles. We need to discuss about this being correct

src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
rafalkalita: ok, changing the behaviour to preserve other properties

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
rafalkalita: Going for option 2, because introducing getCollectionFields is tricky. We have two methods, one with List<CaseViewField> and one with List<CaseFiled>. It is easier to do the filter manually.

src/main/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtil.java
rafalkalita: done 👍

src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
rafalkalita: both done 👍

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
rafalkalita: As discussed I will put a comment for ExUI this is for both Case and Idam roles

src/main/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtil.java
```diff
@@ -0,0 +1,74 @@
...
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static uk.gov.hmcts.ccd.domain.model.common.DisplayContextParameterCollectionOptions.ALLOW_DELETE;
+import static uk.gov.hmcts.ccd.domain.model.common.DisplayContextParameterCollectionOptions.ALLOW_INSERT;
+
+public class DisplayContextParameterUtil {
+
+    public static String COLLECTION = "COLLECTION";
+
+    public static String updateCollectionDisplayContextParameter(final String currentDisplayContextParameter,
```
mario-paniccia: this code could seems a bit too complex would be good if we could further simplify by breaking it down into more private method with self explanatory names

src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
```diff
@@ -0,0 +1,84 @@
...
+                is("#COLLECTION()"))
+                 );
+    }
+
+    @Test
+    @DisplayName("should update parameter)")
+    void shouldUpdateParameter() {
+        assertAll(
+            () -> assertThat(updateCollectionDisplayContextParameter("#COLLECTION(allowInsert,allowDelete)", asList("allowInsert", "allowDelete")),
+                is("#COLLECTION(allowDelete,allowInsert)")),
```
mario-paniccia: this one is funny, how come it's inverting the order?

src/test/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceFilterTest.java
```diff
@@ -1152,6 +1154,162 @@ void filterComplexCaseFieldChildrenByUpdateAccessWhenAllAccessIsOnCreate() {
...
+            assertAll(
+                () -> assertNotNull(people),
+                () -> assertTrue(people.getDisplayContextParameter().contains("#COLLECTION(")),
+                () -> assertFalse(people.getDisplayContextParameter().contains("allowInsert")),
+                () -> assertFalse(people.getDisplayContextParameter().contains("allowDelete"))
+                     );
+        }
+    }
+
     @Nested
```
mario-paniccia: We should add few integration tests as well, especially one testing collection of complex of collection...

src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
rafalkalita: Because we always order the results.

src/test/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceFilterTest.java
rafalkalita: Yes, can add few.

src/main/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtil.java
rafalkalita: Will try to make it more clear.

rafalkalita:
Added integration tests. Only tricky part is, we cannot drop 'create' access to test displayContextParameter, because the filed is going to be filtered out.

src/test/java/uk/gov/hmcts/ccd/domain/model/common/DisplayContextParameterUtilTest.java
mario-paniccia: 👍

PR #757, opened 2020-01-17
Rdm 7225

Thor-tech-of-metal:
RDM-7225: Endpoint validations for get case..
    
      [RDM-7225] (https://tools.hmcts.net/jira/browse/RDM-7225).


src/main/java/uk/gov/hmcts/ccd/validators/annotations/CcdAlphaNumericId.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({PARAMETER, FIELD})
+@Retention(RUNTIME)
+@Constraint(validatedBy = CcdAlphaNumericIdValidator.class)
+@Documented
+public @interface CcdAlphaNumericId {
```
MSancaktutar: Did we have to mention "Id" in the naming of this? This looks like a more generic annotation for alpha-numeric.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
```diff
@@ -108,19 +113,25 @@ public CaseDetailsEndpoint(@Qualifier(CreatorGetCaseOperation.QUALIFIER) final G
     @ApiOperation(value = "Get case", notes = "Retrieve an existing case with its state and data")
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = "Case found for the given ID"),
-        @ApiResponse(code = 400, message = "Invalid case ID"),
+        @ApiResponse(code = 400, message = "Invalid request"),
```
MSancaktutar: We can keep this as "Bad Request" as a standard across CCD, particularly as we already have catalogue response object that contains business reading of this response?

mario-paniccia:
This API improvement proposal is valuable but some disagreements have emerged on the new API structure proposed. A common agreement on the new JSON structure of the API response could not be achieved. Tech Leads (Mutlu Sancaktutar, Mario Paniccia) and Architects (Lee Ash) decided then not to go on with this change for now but to instead tackle this as part of a future proposal for a v3 of the CCD API after a common agreement has been reached.

PR #766, opened 2020-01-21
Rdm 7224/5 common

Thor-tech-of-metal:
Rdm 7224 5. This Pr contains all common  code for Rdm 7224 and   Rdm 7224.

  [RDM-7224] (https://tools.hmcts.net/jira/browse/RDM-7224).
  [RDM-7225] (https://tools.hmcts.net/jira/browse/RDM-7225).


src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponseCode.java
```diff
@@ -0,0 +1,39 @@
+package uk.gov.hmcts.ccd.domain.model.common;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+public enum CatalogueResponseCode {
```
MSancaktutar: Could we please rename this to CatalogueResponseElement? This is because the words code and message repeating in multiple places causing a partial ambiguity in smooth reading and understanding of code.

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponseCode.java
```diff
@@ -0,0 +1,39 @@
...
+
+    // CCD.02 Callback codes
+    CALLBACK_FAILURE(CatalogueResponseGroup.CALLBACK, "01", "Callback failure."),
+    CALLBACK_PAYMENT_REQUIRED(CatalogueResponseGroup.CALLBACK, "02", "Payment required by callback.");
+
+    final CatalogueResponseGroup group;
+    final String code;
+    final String message;
+
+    CatalogueResponseCode(final CatalogueResponseGroup group, final String code, final String message) {
```
MSancaktutar: The word 'code' is used in multiple places with different content/meaning. Better to rename. I suggest to use an integer for this, and name it as "sequenceNo". The getCode() should then return a formatted String to contain ".01" for sequence is 1, for example. There should be a method "int getSequenceNo()", and no attribute like 'code' in the class. An attribute X and a method getX containing/returning different values is not good.

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponseGroup.java
```diff
@@ -0,0 +1,19 @@
...
+
+public enum CatalogueResponseGroup {
+
+    SUCCESS("00"),
+    VALIDATION("01"),
+    CALLBACK("02");
+
+    final String code;
+
+    CatalogueResponseGroup(final String code) {
```
MSancaktutar: code as input to constructor and code returned from getCode() are inconsistent, reducing readability. Let's rename the parameter like int groupId. We will also need a groupId attribute and it's getter.

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponseGroup.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.hmcts.ccd.domain.model.common;
+
+public enum CatalogueResponseGroup {
+
+    SUCCESS("00"),
+    VALIDATION("01"),
+    CALLBACK("02");
+
+    final String code;
```
MSancaktutar: we should remove this. getCode() should compute this from groupId

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponse.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final String code;
+    private final String message;
+    private final T details;
+
+    private CatalogueResponse(final String code, final String message, final T details) {
+        this.code = code;
+        this.message = message;
+        this.details = details;
+    }
```
MSancaktutar: One more overload to take no details.

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponseGroup.java
```diff
@@ -0,0 +1,19 @@
...
+
+    CatalogueResponseGroup(final String code) {
+        this.code = code;
+    }
+
+    public String getCode() {
+        return String.format("CCD.%s", code);
```
MSancaktutar: CCD part should also be an attribute. That may help us later in processing similar structure messages from different domains. Let's name it like "domain".

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponse.java
Thor-tech-of-metal: Is it what is required default constructor ? 
public CatalogueResponse()
{
  super();
   this.code =null;
        this.message =null;
        this.details = null;
}

MSancaktutar: I meant "private CatalogueResponse(final String code, final String message)"

MSancaktutar: @Thor-tech-of-metal  - Kindly for your attention.. ^^^

src/main/java/uk/gov/hmcts/ccd/domain/model/common/HttpError.java
```diff
@@ -23,29 +25,63 @@
...
 
-    private HttpStatus getHttpStatus(ResponseStatus responseStatus) {
+    private HttpStatus getHttpStatus(final Integer status) {
+        // NB: don't return value if matches default
+        if (HttpStatus.INTERNAL_SERVER_ERROR.value() != status.intValue()) {
+            return HttpStatus.resolve(status.intValue());
+        }
+
+        return null;
+    }
```
MSancaktutar: The logic here doesn't look consistent, as it is returning a null for 500. Either the code should be fixed or the method be renamed to imply the logic inside it.

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponse.java
Thor-tech-of-metal: Perfect Mutlu .  I will do it :)

src/main/java/uk/gov/hmcts/ccd/domain/model/common/HttpError.java
```diff
@@ -23,29 +25,59 @@
...
 
-    private Integer getStatus(ResponseStatus responseStatus) {
-        if (null != responseStatus) {
+    private Integer getStatusValue(final Integer status, final ResponseStatus responseStatus) {
+        // Option 1: use status as it may be an override value for responseStatus
+        if (status != null) {
+            return status;
+
+        // Option 2: use responseStatus
+        } else if (null != responseStatus) {
```
MSancaktutar: I got what this method is doing now. But, if status codes annotated for Exceptions take precedence, couldn't it be a programming error to provide a status value for an Exception that already has an inconsistent annotated status?

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponse.java
```diff
@@ -0,0 +1,77 @@
...
+    private final String code;
+    private final String message;
+    private final T details;
+
+    public CatalogueResponse() {
+        super();
+        this.code = null;
+        this.message = null;
+        this.details = null;
+    }
```
MSancaktutar: This wasn't my point, as agreed already. :)

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponse.java
```diff
@@ -0,0 +1,74 @@
...
+        this(code,message,null);
+    }
+
+    public CatalogueResponse(final CatalogueResponseElement catalogueResponseElement, final T details) {
+        this(catalogueResponseElement.getCode(), catalogueResponseElement.getMessage(), details);
+    }
+
+    public CatalogueResponse(final CatalogueResponseElement catalogueResponseElement) {
+        this(catalogueResponseElement, null);
+    }
```
mario-paniccia: the whole point of having a builder is to avoid having telescoping constructor pattern, which is what you have here anyway. We should get rid of all these overloaded constructors.

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/ApiException.java
```diff
@@ -11,16 +12,28 @@
 )
 public class ApiException extends RuntimeException {
 
+    private final transient CatalogueResponse catalogueResponse;
```
mario-paniccia: for consistency with the other properties I would remove `final`

src/test/java/uk/gov/hmcts/ccd/endpoint/exceptions/ApiExceptionTest.java
```diff
@@ -0,0 +1,122 @@
...
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.*;
+
+public class ApiExceptionTest {
+
+    @Test
+    public void shouldExtractMessageFromCatalogueResponse() {
+
+        // ARRANGE
+        final CatalogueResponse testCatalogueResponse =
```
mario-paniccia: there is little value in having `final` in test classes. It makes the code very heavy to read in my opinion. I would remove it. Would also remove it in general. It make the code unclean for very little benefit

src/test/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponseTest.java
```diff
@@ -0,0 +1,78 @@
...
+            "{ \"code\": \"%s\", \"message\": \"%s\", \"details\": { \"test1\": 1, \"test2\": \"Test String\" } }",
+            CatalogueResponseElement.CALLBACK_FAILURE.getCode(), CatalogueResponseElement.CALLBACK_FAILURE.getMessage());
+        final ObjectMapper mapper = new ObjectMapper();
+
+        // ACT
+        final CatalogueResponse actualCatalogueResponse
+            = mapper.readerFor(CatalogueResponse.class).readValue(testJson);
+
+        // ASSERT
+        assertTrue(EqualsBuilder.reflectionEquals(expectedCatalogueResponse, actualCatalogueResponse));
```
mario-paniccia: better using Hamcrest: 
``` 
assertThat(expectedCatalogueResponse, samePropertyValuesAs(actualCatalogueResponse));
```

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponse.java
mario-paniccia: http://www.developerdave.co.uk/2016/05/using-the-builder-pattern-in-jackson/

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CatalogueResponse.java
```diff
@@ -0,0 +1,74 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonDeserialize(builder = CatalogueResponse.Builder.class)
+public class CatalogueResponse<T> {
+
+    private final String code;
+    private final String message;
+    private final T details;
```
mario-paniccia: how come this has been modeled as a generic? isn't a bit overcomplicating things?
If details is dynamic in structure, we could model it as a Map<String, JsonNode> same way as we do for the `data` field of the Case payload.

mario-paniccia:
This API improvement proposal is valuable but some disagreements have emerged on the new API structure proposed. A common agreement on the new JSON structure of the API response could not be achieved. Tech Leads (Mutlu Sancaktutar, Mario Paniccia) and Architects (Lee Ash) decided then not to go on with this change for now but to instead tackle this as part of a future proposal for a v3 of the CCD API after a common agreement has been reached.

PR #769, opened 2020-01-22
Rdm 6240

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-6240

### Change description ###

Sanitise inputs in SortOrderQueryBuilder. Replace SortOrderQueryBuilderTest with integration tests in DefaultCaseDetailsRepositoryTest.

SortOrderQueryBuilder generates SQL queries. To verify that the generated SQL is correct, it needs to be tested by executing it against the Hibernate/Postgres stack and validating the results. Unit testing this means testing implementation rather than behaviour, leading to brittle tests that are tightly coupled to the implementation. I have ensured coverage for the deleted tests by expanding DefaultCaseDetailsRepositoryTest, adding new tests where necessary.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
```diff
@@ -12,23 +21,39 @@
...
+                                              CaseDetailsEntity.STATE_FIELD_COL,
+                                              CaseDetailsEntity.REFERENCE_FIELD_COL,
+                                              CaseDetailsEntity.SECURITY_CLASSIFICATION_FIELD_COL)
+                                          .collect(Collectors.toCollection(HashSet::new)));
+
     public String buildSortOrderClause(MetaData metaData) {
         StringBuilder sb = new StringBuilder();
         metaData.getSortOrderFields().forEach(sortOrderField -> {
-            if (sortOrderField.isMetadata()) {
+            if (sortOrderField.isMetadata() && VALID_COLUMNS_FOR_ORDER_BY.contains(getMataFieldName(sortOrderField.getCaseFieldId()))) {
```
smathangi: this validation already occurs in definition-store. If we re-force it here then we should throw an error immediately rather than fallback to JSONB group. 

Also `VALID_FIELD_NAME_REGEX` should in sync with fieldNames pattern of definition.

src/test/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilderTest.java
```diff
@@ -1,80 +0,0 @@
-package uk.gov.hmcts.ccd.data.casedetails.search;
-
```
MSancaktutar: Looks good overall, but couldn't grasp the reason to remove this test class at first sight.

andrewfolga:
@smathangi I agree that no sql injection occurs in this class and suggested we do not add validation there. This class looks unlikely to be reused in any other context and also it introduces duplication as whenever the change occurs in the valid field name regexp in def store we have to keep it up to date in data store too.

andrewfolga:
Tests in this class were fully replaced by integration tests - see description.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
```diff
@@ -12,23 +21,39 @@
...
-            sb.append(COMMA);
-            sb.append(SPACE);
         });
         // always sort with creation_date as a last order so that it supports cases where no values at all for the configured fields and also default fallback.
         return sb.append(CREATED_DATE + SPACE + SortDirection.fromOptionalString(metaData.getSortDirection())).toString();
     }
 
+    private void appendOrder(final StringBuilder sb, final SortOrderField sortOrderField) {
+        sb.append(SPACE);
+        sb.append(sortOrderField.getDirection());
```
banderous: I would put the validation here, at the exact point where the data is being converted to SQL, and throw an exception if the sort order is invalid. Otherwise I have to look around a bunch of different places to figure out if the validation always happens.

andrewfolga: Not sure I understand - the appending happens in the if else as well?

banderous:
Re: why it's worth fixing this even though definition store performs its own validation.

1. If someone was able to compromise definition store they could immediately compromise data store by tweaking a definition to exploit this vulnerability and access all case data
2. It follows that it is dangerous to trust any input whether from a user or anywhere else in the organisation -see [the owasp guide on sql injection](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.md)

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
andrewfolga: ok I will add a top level if block looking for special characters:

`       metaData.getSortOrderFields().forEach(sortOrderField -> {
            if (!sortOrderField.getCaseFieldId().matches(SPECIAL_CHARS)) {
                if (sortOrderField.isMetadata()) {
                    sb.append(getMataFieldName(sortOrderField.getCaseFieldId()));
                } else {
                    sb.append(convertFieldNameToJSONBsqlFormat(sortOrderField.getCaseFieldId()));
                }
                sb.append(SPACE);
                sb.append(sortOrderField.getDirection());
                sb.append(COMMA);
                sb.append(SPACE);
            }
        });`

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
```diff
@@ -12,23 +21,39 @@
...
+    private static final Set<String> VALID_COLUMNS_FOR_ORDER_BY
+        = Collections.unmodifiableSet((Set<? extends String>) Stream
+                                          .of(CaseDetailsEntity.CREATED_DATE_FIELD_COL,
+                                              CaseDetailsEntity.LAST_MODIFIED_FIELD_COL,
+                                              CaseDetailsEntity.JURISDICTION_FIELD_COL,
+                                              CaseDetailsEntity.CASE_TYPE_ID_FIELD_COL,
+                                              CaseDetailsEntity.STATE_FIELD_COL,
+                                              CaseDetailsEntity.REFERENCE_FIELD_COL,
+                                              CaseDetailsEntity.SECURITY_CLASSIFICATION_FIELD_COL)
+                                          .collect(Collectors.toCollection(HashSet::new)));
```
MSancaktutar: Based on the latest discussions around logic duplication between def store and data store, I think we can indeed remove this allowed-list approach as well. We can stay within the purpose of just SQL-i prevention by just rejecting a few special chars in case field ids, and indeed, case field IDs can reasonable assumed to not contain such chars, no matter how their validity conventions change in the foreseeable future.

andrewfolga:
@banderous yes after a quick chat with @MSancaktutar we came to the same conclusions re the need of the validation here and I am adding it now 👍

andrewfolga:
@banderous @smathangi @MSancaktutar can you re-review now please?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
banderous: The [OWASP guide](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.md) recommends blacklisting only as a last resort:

> This technique should only be used as a last resort, when none of the above are feasible. Input validation is probably a better choice as this methodology is frail compared to other defenses and we cannot guarantee it will prevent all SQL Injection in all situations.

If in this case we know the range of possible values then Andrew's original whitelisting would be my preference but either approach is an improvement.

Regarding the duplication between definition and data stores, I see this as an indicator that these services should be merged together, which would also simplifying testing (no wiremock or duplication of models, cannot commit breaking changes to ccd into def store).

andrewfolga: Yes I was following the OWASP guide and initially tried to do white listing solution. The problem with that is it creates dependency on whatever change is introduced to `case_data` table's columns. I am ok with either solution here but the the common denominator between @MSancaktutar @smathangi and myself was blacklisting...
@MSancaktutar let me know what you think?

MSancaktutar: Sanitising input against a disallowed list is in this case more efficient and well-secure. It is an ORDER BY clause we're building and preventing those disallowed chars solves the same problem to the same effect with less coupling with other parts of CCD.

banderous: The OWASP guide is explicit that blacklisting should be the last resort and only used for securing legacy code; there are issues such as Unicode smuggling where special characters can get past a blacklist and get converted to dangerous characters by the database.

In this case the range of valid input is known so whitelisting is the secure way to go; a switch statement that only outputs known valid constants.

Database queries are inherently coupled to the schema they operate on, which is why it's problematic to split schema out of a database into its own service as we have done. If we are going to persist with the split then duplication is inevitable.

src/test/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepositoryTest.java
```diff
@@ -201,6 +222,35 @@ public void getFindByMetadataAndFieldDataSortDesc() {
...
+                                       .direction("ASC")
+                                       .build());
+        metadata.addSortOrderField(SortOrderField.sortOrderWith()
+                                       .caseFieldId("PersonLastName")
+                                       .metadata(false)
+                                       .direction("DESC")
+                                       .build());
+
+        HashMap<String, String> searchParams = new HashMap<>();
+        searchParams.put("PersonFirstName", "Janet");
```
andrewfolga: remove to not confuse - only order by

src/test/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepositoryTest.java
```diff
@@ -201,6 +222,35 @@ public void getFindByMetadataAndFieldDataSortDesc() {
...
+        final List<CaseDetails> byMetaDataAndFieldData = caseDetailsRepository.findByMetaDataAndFieldData(metadata,
+            searchParams);
+
+        // See the timestamps in insert_cases.sql.
+        // Should be ordered by person last name, last modified desc, creation date asc.
```
andrewfolga: double check comment

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
smathangi: I think this whitelist check for metadata fields are already happening against a ENUM here   https://github.com/hmcts/ccd-data-store-api/blob/75fbac732d9e24f220afa20f4f56f487f5c5a03c/src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java#L42

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
```diff
@@ -4,26 +4,34 @@
...
-                sb.append(convertFieldNameToJSONBsqlFormat(sortOrderField.getCaseFieldId()));
+            Matcher matcher = SPECIAL_CHARS_PATTERN.matcher(sortOrderField.getCaseFieldId());
+            if (!matcher.find()) {
+                if (sortOrderField.isMetadata()) {
+                    sb.append(getMataFieldName(sortOrderField.getCaseFieldId()));
+                } else {
+                    sb.append(convertFieldNameToJSONBsqlFormat(sortOrderField.getCaseFieldId()));
+                }
+                sb.append(SPACE);
+                sb.append(sortOrderField.getDirection());
```
banderous: We are appending an arbitrary String here - this should be an Enum, or it looks like you could use SortDirection.fromOptionalString

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
```diff
@@ -4,26 +4,34 @@
...
-            if (sortOrderField.isMetadata()) {
-                sb.append(getMataFieldName(sortOrderField.getCaseFieldId()));
-            } else {
-                sb.append(convertFieldNameToJSONBsqlFormat(sortOrderField.getCaseFieldId()));
+            Matcher matcher = SPECIAL_CHARS_PATTERN.matcher(sortOrderField.getCaseFieldId());
+            if (!matcher.find()) {
+                if (sortOrderField.isMetadata()) {
+                    sb.append(getMataFieldName(sortOrderField.getCaseFieldId()));
+                } else {
+                    sb.append(convertFieldNameToJSONBsqlFormat(sortOrderField.getCaseFieldId()));
```
banderous: convertFieldNameToJSONBsqlFormat is also vulnerable since it just passes through the input with some substitutions - can we apply a whitelist of known allowed characters to it?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
```diff
@@ -1,37 +1,55 @@
...
         return sb.append(CREATED_DATE + SPACE + SortDirection.fromOptionalString(metaData.getSortDirection())).toString();
     }
 
     private String getMataFieldName(String fieldName) {
         String metaFieldName = fieldName.startsWith("[") ? StringUtils.substringBetween(fieldName, "[", "]") : fieldName;
-        return CaseField.valueOf(metaFieldName).getDbColumnName();
+        String dbColumnName = null;
+        try {
+            dbColumnName = CaseField.valueOf(metaFieldName).getDbColumnName();
+        } catch (IllegalArgumentException iae) {
```
banderous: You could let the exception propagate and not catch it, or rethrow it. Returning a null values will result in a more cryptic error later on.

src/test/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepositoryTest.java
```diff
@@ -153,6 +153,27 @@ public void sanitisesInputs() {
...
+            .thenReturn(asList(evil));
+
+        when(userAuthorisation.getAccessLevel()).thenReturn(AccessLevel.GRANTED);
+        when(userAuthorisation.getUserId()).thenReturn(evil);
+
+        MetaData metadata = new MetaData("TestAddressBookCase", "PROBATE");
+        metadata.addSortOrderField(SortOrderField.sortOrderWith()
+                                       .caseFieldId(evil)
+                                       .metadata(false)
+                                       .direction("DESC")
```
banderous: Test passing evil in to the direction too

banderous:
Just a general point on strategy - I appreciate that validation is being applied elsewhere in the organisation.

However having to think about all the places input might be coming from and whether it is safe is a lot of mental effort! And easy to be make a mistake, plus someone might make a change elsewhere and not realise they've just created an injection vulnerability by bypassing sanitisation.

Ideally I can look at every call to stringbuilder.append and be confident that it will only be putting in safe input without having to look anywhere else.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
banderous: Yes @smathangi this is the right approach, a couple more places where the same needs to be done.

andrewfolga: @banderous can you let me know which places do you mean?

banderous: @andrewfolga individual comments are in the review I made requesting changes.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
andrewfolga: Good point - done.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
andrewfolga: We already handle that input with blacklisting as per @MSancaktutar comments? See `Matcher matcher = SPECIAL_CHARS_PATTERN.matcher(sortOrderField.getCaseFieldId());`

andrewfolga: This can be any field from withing the case data json so not sure how we would do whitelisting here. Can you elaborate please?

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
andrewfolga: I followed your approach from here:
```    @Test
    public void sanitisesInputsCountQuery() {
        String evil = "foo');insert into case users values(1,2,3);--";
        when(authorisedCaseDefinitionDataService.getUserAuthorisedCaseStateIds("PROBATE", "TestAddressBookCase", CAN_READ))
            .thenReturn(asList(evil));

        when(userAuthorisation.getAccessLevel()).thenReturn(AccessLevel.GRANTED);
        when(userAuthorisation.getUserId()).thenReturn(evil);

        MetaData metadata = new MetaData("TestAddressBookCase", "PROBATE");
        final PaginatedSearchMetadata byMetaData = caseDetailsRepository.getPaginatedSearchMetadata(metadata, Maps.newHashMap());

        // If any input is not correctly sanitized it will cause an exception since query result structure will not be as hibernate expects.
        assertThat(byMetaData.getTotalResultsCount(), is(0));
    }

Where you also do not throw or propagate exception but do not append the section of sql query being built that contains a validation problem.

**Comment by andrewfolga** [comment]
@banderous I've considered and proposed ditching StringBuilder too but it would mean rebuilding your `SearchQueryFactoryOperation` with CriteriaQuery and would be outside of this ticket.

If the `SearchQueryFactoryOperation` was done with CriteriaQuery it would make it possible to do this `SortOrderQueryBuilder` within this ticket.

File: src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java (hunk 8 above)
Comment (banderous) [reply src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java]: Ah yes sorry, I think we are back to the whitelisting vs blacklisting thing again. Whitelisting would involve asserting that the column name only consists of the set of allowable characters in it but we've been over that one :)

File: src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java (hunk 9 above)
Comment (banderous) [reply src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java]: The fix in that case was using OWASP option 1 - parameterised queries - it doesn't matter what the input is, it should still result in a valid query (that returns zero results).

In this case the column name is a parameter so we cannot use a parameterised query and must use OWASP option three - explicitly rejecting the bad input. So in this case I would design the test to expect that IllegalArgument exception and fail if not thrown.

**Comment by banderous** [comment]
> @banderous I've considered and proposed ditching StringBuilder too but it would mean rebuilding your `SearchQueryFactoryOperation` with CriteriaQuery and would be outside of this ticket.

Unfortunately in this case I think we are stuck with StringBuilder since column names cannot be parameterised. From the [guide](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.md):

> Various parts of SQL queries aren't legal locations for the use of bind variables, such as the names of tables or columns

File: src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java (hunk 9 above)
Comment (banderous) [reply src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java]: I would remove this exception handler completely and assert that the illegalargumentexception is thrown in your test

File: src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java [hunk 11]
```diff
@@ -1,37 +1,55 @@
...
+                sb.append(SortDirection.fromOptionalString(ofNullable(sortOrderField.getDirection())));
+                sb.append(SPACE);
             }
-            sb.append(SPACE);
-            sb.append(sortOrderField.getDirection());
-            sb.append(COMMA);
-            sb.append(SPACE);
         });
         // always sort with creation_date as a last order so that it supports cases where no values at all for the configured fields and also default fallback.
         return sb.append(CREATED_DATE + SPACE + SortDirection.fromOptionalString(metaData.getSortDirection())).toString();
```
banderous: If we made SortDirection.fromOptionalString throw an exception could this cause an issue in production? Where does metaData.getSortDirection() come from? Could it contain null or some other value that would currently just result in an ASC sort order?

andrewfolga: I left this out as I am unsure the answers to those questions.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/SortOrderQueryBuilder.java
```diff
@@ -1,32 +1,50 @@
...
     public String buildSortOrderClause(MetaData metaData) {
         StringBuilder sb = new StringBuilder();
         metaData.getSortOrderFields().forEach(sortOrderField -> {
-            if (sortOrderField.isMetadata()) {
-                sb.append(getMataFieldName(sortOrderField.getCaseFieldId()));
+            Matcher matcher = SPECIAL_CHARS_PATTERN.matcher(sortOrderField.getCaseFieldId());
+            if (matcher.find()) {
+                LOG.error("Illegal sort order field id: {}", sortOrderField.getCaseFieldId());
+                throw new IllegalArgumentException("Illegal sortOrderField.caseFieldId=" + sortOrderField.getCaseFieldId());
             } else {
```
banderous: you could get rid of this else now that we're throwing an exception

banderous: by which I mean the block and braces, not the code inside!

andrewfolga: Done

andrewfolga:
@banderous The solution for whitelisting in `SortOrderQueryBuilder` would restrict the set of potential values that are coming from def store for the metadata sort order fields. We know the character set that the PROD data contains allows the following:
`^[a-zA-Z0-9\xA0\-\'\ \#\%\&\(\)\.\?\[\]\_\£\s]*$`
If you are unable to approve this character set then knowing that we cannot go with the following one either:
`^[A-Za-z0-9_.-]+$`

I suggest we open another ticket to discuss the whitelisting of metadata sort order fields + case data case field ids and leave the current column validation improvement as is.

PR #787, opened 2020-02-02
RDM-7457 - SpringBoot to latest 2.2.4

smathangi:

https://tools.hmcts.net/jira/browse/RDM-7457

### Change description ###

spring_boot_upgrade to latest 2.2.4 Release

```
[ ] Yes
[ ] No
```


src/aat/resources/features/F-044/S-579-Prerequisite_Solicitor_3_Token_For_Update_Case.td.json
```diff
@@ -30,7 +30,7 @@
     "_extends_": "Common_200_Response",
     "headers": {
       "Content-Encoding": "gzip",
-      "Content-Type": "application/json;charset=UTF-8",
+      "Content-Type": "application/json",
```
mario-paniccia: not for this PR but I was thinking whether it's possible to have for the headers a common header or few common headers so that when similar changes happen we don't have to update all these files

MSancaktutar: Yes, this header should have been in a common one that’s already in place.
Probably was missed initially.

PR #792, opened 2020-02-03
Remove downstream call exception details from case type not found responses

MSancaktutar:
https://tools.hmcts.net/jira/browse/RDM-7509

### Change description ###
Added cache to data store for jurisdictions.

```
[ ] Yes
[X] No
```


charts/ccd-data-store-api/values.preview.template.yaml
```diff
@@ -19,6 +19,7 @@ java:
     DATA_STORE_DB_PASSWORD: "{{ .Values.postgresql.postgresqlPassword}}"
     DATA_STORE_DB_OPTIONS: "?stringtype=unspecified"
     DATA_STORE_DB_MAX_POOL_SIZE: 10
+    DEFINITION_STORE_HOST: http://ccd-definition-store-api-pr-469.service.core-compute-preview.internal/
```
smathangi: is it intentional?

MSancaktutar: nope. will remove it.

PR #799, opened 2020-02-06
Rdm 4641 - Persist state last modified dateTime

smathangi:

https://tools.hmcts.net/jira/browse/RDM-4641

### Change description ###

 Persist state last modified dateTime

```
[ ] Yes
[ x] No
```


src/aat/resources/features/F-036/S-199.td.json
```diff
@@ -29,6 +29,7 @@
         "case_type_id": "AAT_AUTH_15",
         "created_date": "[[ANY]]",
         "last_modified": "[[ANY]]",
+        "last_state_modified_date": "[[ANY]]",
```
mario-paniccia: not that great that we have to update so many test data files when we add a property to the json. This number of json files will only grow in the future. I see some serious maintainability problems @MSancaktutar

MSancaktutar: If the outcomes of many functions are changing out of a code change, we should update all the functional verifications. That's integral part of the delivery. However, to reduce the maintenance efforts, we already have data inheritance/re-use mechanisms in place. The current use of the mechanism is already reducing that load a huge amount. If further uses are possible, we can improve the test data, just as with anything else.

MSancaktutar: For example, SB upgrade brought a new Strict-Transport-Security header to almost all our responses, and we added it to only 2 or 3 places among 300+ files, to reflect the change to the verification of hundreds of test scenarios. This is how things should ideally work, and we can achieve it gradually improving the quality of test data files for smarter use of the framework etc.

mario-paniccia: adding a new property to the case data it's something that it's gonna happen again in the future. I'd suggest we apply the inheritance/re-use mechanism for these cases  as well to reduce the number of test files we'll have to update in the future. Too many test files had to be updated in this PR. I'd suggest we raise an Improvement ticket for this

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/CreateCaseEventService.java
```diff
@@ -168,12 +168,15 @@ private CaseDetails getCaseDetails(final String caseReference) {
...
 
-    private CaseDetails saveCaseDetails(final CaseDetails caseDetails,
+    private CaseDetails saveCaseDetails(CaseDetails caseDetailsBefore, final CaseDetails caseDetails,
                                         final CaseEvent eventTrigger,
                                         final Optional<String> state) {
         if (!state.isPresent() && !equalsIgnoreCase(CaseState.ANY, eventTrigger.getPostState())) {
             caseDetails.setState(eventTrigger.getPostState());
         }
+        if (!caseDetails.getState().equalsIgnoreCase(caseDetailsBefore.getState())) {
+            caseDetails.setLastStateModifiedDate(LocalDateTime.now(ZoneOffset.UTC));
```
mario-paniccia: just an improvement suggestion. Feel free to ignore. 
Usually when working with dates, to make testability simpler, it's recommended to have a DateService with methods like now() and others which can then be mocked in Unit tests. We could introduce that

smathangi: Yes, I agree. I'll see if i can introduce "Clock" class.

smathangi: done

src/main/java/uk/gov/hmcts/ccd/CoreCaseDataApplication.java
```diff
@@ -33,4 +37,9 @@ public static void main(String[] args) {
         SpringApplication.run(CoreCaseDataApplication.class, args);
     }
 
+    @Bean
+    public Clock uTCClock() {
```
mario-paniccia: shall we do `utcClock`?

src/main/java/uk/gov/hmcts/ccd/CoreCaseDataApplication.java
```diff
@@ -33,4 +37,9 @@ public static void main(String[] args) {
...
+        return Clock.system(ZoneOffset.UTC);
```
mario-paniccia: could we replace with: 
`Clock.systemUTC()`?

mario-paniccia: can we notify QAs we've made some changes on the way we retrieve current time ? Better doing some regression testing on this

smathangi: sure but no changes tbh. 
LocalDateTime.now(ZoneOffset.UTC) => LocalDateTime.now(Clock.systemUTC())

mario-paniccia: ok then if you are 100% confident ignore my request

PR #811, opened 2020-02-12
Rdm 5691

smathangi:

https://tools.hmcts.net/jira/browse/RDM-5691

### Change description ###

Update Existing cases with new [LAST_STATE_MODIFIED_DATE] metadata

```
[ ] Yes
[ x] No
```


lastStateModifiedMigration.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
smathangi: Throw away endpoint after we run on production once.

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/CreateCaseEventService.java
```diff
@@ -216,7 +226,7 @@ private void saveAuditEventForCaseDetails(final AboutToSubmitCallbackResponse ab
         auditEvent.setUserId(user.getId());
         auditEvent.setUserLastName(user.getSurname());
         auditEvent.setUserFirstName(user.getForename());
-        auditEvent.setCreatedDate(LocalDateTime.now(ZoneOffset.UTC));
+        auditEvent.setCreatedDate(now());
```
MSancaktutar: createdDate in the audit event may potentially be different from newCaseDetails.setLastStateModifiedDate(now).
Would this not be a problem?

src/main/java/uk/gov/hmcts/ccd/endpoint/LastStateModifiedMigrationEndpoint.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Autowired
+    public LastStateModifiedMigrationEndpoint(LastStateModifiedMigrationService lastStateModifiedMigrationService) {
+        this.service = lastStateModifiedMigrationService;
+    }
+
+    @PostMapping(value = "/last-state-modified/migrate")
+    public void migrate(@RequestParam(value = "jurisdiction") final String jurisdiction,
+                                           @RequestParam(value = "batchSize", required = false, defaultValue = "5000") final int batchSize,
+                                           @RequestParam(value = "dryRun", required = false, defaultValue = "false") final boolean dryRun) {
```
MSancaktutar: dryRun default value should be true.

src/main/resources/db/changelog/db.changelog-RDM-4641.xml
```diff
@@ -0,0 +1,13 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+
+    <changeSet id="rdm-4641" author="sateesh.mathangi@hmcts.net">
+        <addColumn tableName="case_data">
+            <column name="last_state_modified_date" type="timestamp">
+                <constraints nullable="true" />
```
MSancaktutar: This should be changed to nullable="false", but maybe after the migration

smathangi: Yes, after migration.

src/main/java/uk/gov/hmcts/ccd/endpoint/LastStateModifiedMigrationEndpoint.java
smathangi: done

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/CreateCaseEventService.java
smathangi: I thought about it. But I guess I don't think it'll give a problem. The difference is in just nano seconds. Similar thing we've already in `setLastModified` in  
https://github.com/hmcts/ccd-data-store-api/blob/master/src/main/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepository.java#L70
Having said I'm happy to fix for `LastStateModified`. Please let me know.

smathangi: Done :)

MSancaktutar: As long as there is no and will be no logic in code or in data analysis/reporting that assumes strict equality of the values in both places, nano seconds wouldn't really matter. But my worry was around potential later issues with codes or data analytics assuming strict equality. Thanks for the improvement.

PR #816, opened 2020-02-19
RDM-7575 Display Context Parameter in models

RebeccaBaker:
https://tools.hmcts.net/jira/browse/RDM-7575

### Change description ###
Extended models in data store to support displayContextParameter

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/WizardPageComplexFieldOverride.java
```diff
@@ -57,4 +58,13 @@ public String getShowCondition() {
...
     }
+
+    @JsonProperty("display_context_parameter")
+    public String getDisplayContextParameter() {
+        return displayContextParameter;
+    }
+
+    public void setDisplayContextParameter(String displayContextParameter) {
+        this.displayContextParameter = displayContextParameter;
+    }
```
mario-paniccia: major: Why has this been introduced? 
according to the requirements, we won't be supporting overriding of dcp for the complex fields. i.e. we are not introducing a DisplayContextParameter column in EventToComplexTypes sheet. This should then be removed. I think as it stands, this will always end-up being null

PR #827, opened 2020-02-26
RDM-7667 fix NPE when hidden field in a collection

mattnayler:

 [RDM-7667](https://tools.hmcts.net/jira/browse/RDM-7667)

### Change description ###

<del>Fix NullPointerException when a complex type is used inside a collection which contains a hidden field that is either a complex type or a collection.</del>

Fix NullPointerException when a nested complex type is used inside a collection and the child complex type data is being hidden.

<del>Solution: Adjust delete permission check on complex type so it verifies delete permission on current node prior to checking children (i.e. so it can safely skip children if data is legitimately null).</del>

Solution: Adjust delete permission check on complex type to filter nulls in the same way it currently filters empty nodes.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/common/CompoundAccessControlService.java
```diff
@@ -133,7 +133,9 @@ private boolean isCurrentNodeOrAnyChildNodeDeletedWithoutAccess(final JsonNode n
         if (correspondingNewNode.isPresent()) {
             JsonNode newNode = correspondingNewNode.get();
             for (CaseField field : caseField.getFieldType().getCollectionFieldType().getComplexFields()) {
-                if (field.isCompoundFieldType() && existingNode.get(VALUE) != null && existingNode.get(VALUE).get(field.getId()) != null
+                if (field.isCompoundFieldType()
+                    && existingNode.get(VALUE) != null && existingNode.get(VALUE).get(field.getId()) != null
+                    && newNode.get(VALUE) != null && newNode.get(VALUE).get(field.getId()) != null
```
mario-paniccia: This if getting too complicated. Let's move these null checks into a separate private method:
```if (field.isCompoundFieldType() && notNulls(existingNode, newNode, field.getId()) && ....)```

File: src/main/java/uk/gov/hmcts/ccd/domain/service/common/CompoundAccessControlService.java [hunk 2]
```diff
@@ -133,7 +133,9 @@ private boolean isCurrentNodeOrAnyChildNodeDeletedWithoutAccess(final JsonNode n
         if (correspondingNewNode.isPresent()) {
             JsonNode newNode = correspondingNewNode.get();
             for (CaseField field : caseField.getFieldType().getCollectionFieldType().getComplexFields()) {
```
mario-paniccia: Unit Tests? 
Integration Tests?

PR #828, opened 2020-02-26
RDM-7455: [Immigration and Asylum]- Date Validation errors not cleare…

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-7455

### Change description ###

Pass only the current wizard page fields to validate

```
[ ] Yes
[ X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java
```diff
@@ -69,13 +70,12 @@ public JsonNode invoke(String caseTypeId,
...
                     caseDetailsBefore = caseService.clone(caseDetails);
-                    currentOrNewCaseDetails = caseService.populateCurrentCaseDetailsWithEventFields(content,
-                        caseDetails);
-
+                    currentOrNewCaseDetails = caseService.populateCurrentCaseDetailsWithEventFields(content, caseDetails);
                 } else {
                     currentOrNewCaseDetails = caseService.createNewCaseDetails(caseTypeId, caseType.getJurisdictionId(),
                         content.getEventData() == null ? content.getData() : content.getEventData());
                 }
+                removeNextPageFieldData(currentOrNewCaseDetails, wizardPageOptional.get().getOrder(), caseTypeId, event.getEventId());
```
mario-paniccia, line 78: If I don't recall wrong, page display order is optional. Even though the ordering is guaranteed only if a display order is defined, if it's missing I think pages are ordered by page id or in the order they are defined on the spreadsheet (can't recall exactly). In this case is it possible that this code will remove fields from a previous page?
Worth doing some testing.

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java
```diff
@@ -90,6 +90,19 @@ public JsonNode invoke(String caseTypeId,
...
+            Set<String> wizardPageFields = uiDefinitionRepository
+                .getWizardPageCollection(caseTypeId, eventId)
+                .stream()
+                .filter(wizardPage -> wizardPage.getOrder() > order)
+                .map(WizardPage::getWizardPageFieldNames)
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
+            currentCaseDetails.getData().entrySet().removeIf(entry -> wizardPageFields.contains(entry.getKey()));
+        }
+    }
```
mario-paniccia, line 104: this method probably does not belong to the MidEventCallback. Could we introduce an utility class if we don't have a suitable one already? suggestions: `WizardPagesUtils`

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java
kiran-yenigala-hmcts: Will check this one...

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java
```diff
@@ -90,6 +90,19 @@ public JsonNode invoke(String caseTypeId,
         return dataJsonNode(content.getData());
     }
 
+    private void removeNextPageFieldData(CaseDetails currentCaseDetails, Integer order, String caseTypeId, String eventId) {
+        if (order != null) {
```
mario-paniccia, line 94: in theory page display order is optional. Even thought we recommend to always set it. The ordering is guaranteed only if a display order is defined. Otherwise I think pages are ordered by page id or in the order they are defined on the spreadsheet (can't recall exactly). In this case we won't be removing the next pages right?

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java
```diff
@@ -90,6 +90,19 @@ public JsonNode invoke(String caseTypeId,
...
+            Set<String> wizardPageFields = uiDefinitionRepository
```
mario-paniccia, line 95: looking at this now, rather than a `WizardPagesUtils` I think the best thing is the UIDefRepo to return the list of pages wrapped inside a new class `WizardPages` that will then contain all the utility methods to process the pages. But this might be too many changes now. you can ignore it

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java
kiran-yenigala-hmcts: I have verified all the prod files, PageDisplayOrder is defined for most of the definitions except for some fields in below definition files.
**CCD_Publiclaw_BulkScanningExceptionRecord_V00.01-PROD
ccd-fpl-prod**

PR #846, opened 2020-03-12
RDM-7630

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-7630

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventFieldComplex.java
```diff
@@ -11,15 +11,25 @@
...
     }
 
     public CaseEventFieldComplex(String reference,
-                                 Integer order) {
+                                  Integer order) {
         this.reference = reference;
         this.order = order;
     }
 
+    public CaseEventFieldComplex(String reference,
```
Thor-tech-of-metal: Cosmetic: Given that a new constructor was added we can make all attribute immutable and remove setters.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseTypeTab.java
```diff
@@ -1,7 +1,7 @@
 package uk.gov.hmcts.ccd.domain.model.definition;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.*;
```
Thor-tech-of-metal: Not sure if checkstyle likes .*  organise imports

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CriteriaField.java
```diff
@@ -69,4 +70,14 @@ public String getRole() {
     public void setRole(final String role) {
         this.role = role;
     }
+
+    @ApiModelProperty(value = "")
+    @JsonProperty("display_context_parameter")
```
Thor-tech-of-metal: Do we need the  @ApiModelProperty(value = "") ?

src/main/java/uk/gov/hmcts/ccd/domain/service/createcase/DefaultCreateCaseOperation.java
```diff
@@ -14,9 +14,7 @@
 import uk.gov.hmcts.ccd.data.user.UserRepository;
 import uk.gov.hmcts.ccd.domain.model.aggregated.IdamUser;
 import uk.gov.hmcts.ccd.domain.model.callbacks.AfterSubmitCallbackResponse;
-import uk.gov.hmcts.ccd.domain.model.definition.CaseDetails;
-import uk.gov.hmcts.ccd.domain.model.definition.CaseEvent;
-import uk.gov.hmcts.ccd.domain.model.definition.CaseType;
+import uk.gov.hmcts.ccd.domain.model.definition.*;
```
Thor-tech-of-metal: Not sure if checkstyle likes .*  organise imports

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/WorkbasketInputDefinition.java
```diff
@@ -37,4 +38,16 @@ public void setCaseTypeId(String caseTypeId) {
     public void setFields(List<WorkbasketInputField> fields) {
         this.fields = fields;
     }
+
+    /**
+     **/
+    @ApiModelProperty(value = "")
```
Thor-tech-of-metal: Do we need the @ApiModelProperty(value = "") as value="" ?

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchInputDefinition.java
```diff
@@ -40,4 +41,16 @@ public void setFields(List<SearchInputField> fields) {
         this.fields = fields;
     }
 
+    /**
+     **/
+    @ApiModelProperty(value = "")
```
Thor-tech-of-metal: @ApiModelProperty(value = "")  do we need the value="" ?

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+            return newNode;
+        }
+
+        return collectionNode;
+    }
+
+    private TextNode createNode(String displayContextParameter, String valueToConvert, BaseType baseType, String fieldPath) {
+        String format = getDisplayContextParameterOfType(displayContextParameter, DisplayContextParameterType.DATETIMEENTRY)
+            .map(DisplayContextParameter::getValue)
+            .orElse("");
```
Thor-tech-of-metal: I am not sure about the  requirements . Can we create a "" format ?  Would that be default ? or it will lead directly in exception.   If we know before it we can treat it  as  throw new DataProcessingException() . in Common  error handling place.

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+import uk.gov.hmcts.ccd.domain.types.BaseType;
+import uk.gov.hmcts.ccd.domain.types.CollectionValidator;
+import uk.gov.hmcts.ccd.endpoint.exceptions.DataProcessingException;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static uk.gov.hmcts.ccd.domain.model.definition.FieldType.*;
+import static uk.gov.hmcts.ccd.domain.service.processor.DisplayContextParameter.*;
+
```
Thor-tech-of-metal: The same .* .

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DisplayContextParameterType.java
```diff
@@ -0,0 +1,38 @@
+package uk.gov.hmcts.ccd.domain.service.processor;
+
+import com.google.common.base.Strings;
+
+import java.util.*;
+import java.util.regex.Matcher;
```
Thor-tech-of-metal: you can organise imports in inteliJ and add the option to  implicit add each import . It is very cosmetic but sometimes checkstyle tends to complaint about " .* " .

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+        });
+
+        return newNode;
+    }
+
+    protected abstract JsonNode executeSimple(JsonNode node, CommonField field, BaseType baseType, String fieldPath);
+
+    protected abstract JsonNode executeCollection(JsonNode collectionNode, CommonField field, String fieldPath);
+
+    protected boolean isNullOrEmpty(final JsonNode node) {
```
Thor-tech-of-metal: protected boolean isNullOrEmpty(final JsonNode node)  is quite duplicated any. any places . We could create a util class .

src/test/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeFormatParserTest.java
```diff
@@ -0,0 +1,30 @@
...
+    void shouldConvertDateTimeToIso8601() {
+        final String dateTimeFormat = "HHmmssSSS dd/MM/yyyy";
+        final String value = "123059000 20/10/2000";
+
+        final String result = dateTimeFormatParser.convertDateTimeToIso8601(dateTimeFormat, value);
+
+        assertAll(
+            () -> assertThat(result, is("2000-10-20T12:30:59.000"))
+        );
+    }
```
kiran-yenigala-hmcts: Please add test cases for convertDateToIso8601 and some failure cases as well for the below exceptions 
IllegalArgumentException and DateTimeParseException

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CriteriaField.java
danlysiak: Purely added for consistency with rest of this class.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/WorkbasketInputDefinition.java
danlysiak: Purely added for consistency with rest of this class.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchInputDefinition.java
danlysiak: Purely added for consistency with rest of this class.

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+@Component
+public class DateTimeEntryProcessor extends CaseDataFieldProcessor {
```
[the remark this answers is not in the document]
danlysiak: This is to cater for the next PR #851 where there are two different types of `FieldProcessor` (`CaseDataFieldProcessor` being one of them).

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultField.java
```diff
@@ -98,4 +100,12 @@ public String buildCaseFieldId() {
         }
         return getCaseFieldId();
     }
+
+    public String getDisplayContextParameter() {
```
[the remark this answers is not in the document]
danlysiak: Another potential improvement ticket maybe? There are various fields which are on different models (DCP being one of them) so this kind of idea could apply to a number of fields. Naturally abstract wouldn't work in that case but I get the general point!

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventFieldComplex.java
mario-paniccia: minor: 
consider introducing a builder for CaseEventFieldComplex rather than telescoping constructors:
https://www.vojtechruzicka.com/avoid-telescoping-constructor-pattern/

src/main/java/uk/gov/hmcts/ccd/domain/service/createcase/DefaultCreateCaseOperation.java
mario-paniccia: agree with you Kiran. I did raise a ticket for this it's been in Kanban for a while but never gets picked up :)

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+
+        final BaseType fieldType = BaseType.get(caseViewField.getFieldType().getType());
+
+        if (BaseType.get(COMPLEX) == fieldType) {
+            return executeComplex(node, caseField.getFieldType().getComplexFields(), caseEventField, wizardPageField, caseField.getId());
+        } else if (BaseType.get(COLLECTION) == fieldType) {
+            return executeCollection(node, caseViewField, caseField.getId());
+        } else {
+            return executeSimple(node, caseViewField, fieldType, caseField.getId());
+        }
```
mario-paniccia: can't we use 
`caseField.isCollectionFieldType()` 
`caseField.isComplexFieldType()`
?

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
```diff
@@ -0,0 +1,87 @@
...
```
mario-paniccia: Important:
we are introducing the concept of a generic FieldProcessor which is good and in line with the ticket implementation suggestion. I've been wanting to implement this for long. 
But because this is meant to be generic so that can be used in the future for other types of processing, we cannot use inheritance but instead we should use composition. Otherwise we'll end up with an exponentially growing number of subclasses when we need to introduce other types of processing. And it will be not possible to combine different types of processing together. For example CRUD processing together with DateTime formatting. 
We can keep this for now. But I will raise a separate redesign ticket to introduced a generic CaseFieldProcessor.

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+    public DateTimeEntryProcessor(CaseViewFieldBuilder caseViewFieldBuilder,
+                                  DateTimeFormatParser dateTimeFormatParser) {
+        super(caseViewFieldBuilder);
+        this.dateTimeFormatParser = dateTimeFormatParser;
+    }
+
+    @Override
+    protected JsonNode executeSimple(JsonNode node, CommonField field, BaseType baseType, String fieldPath) {
+        return !isNullOrEmpty(node)
+            && hasDisplayContextParameterType(field.getDisplayContextParameter(), DisplayContextParameterType.DATETIMEENTRY)
```
mario-paniccia: To improve code readability and reduce duplication is essential to introduce utility method in the domain classes where they belong to. 
`hasDisplayContextParameterType` should be removed. And instead we should introduced  a method in `DisplayContextParameter` class that allows to check the type of a dcp. For example:
`displayContextParameter.hasType(DisplayContextParameterType.DATETIMEENTRY)`
Then we could add a method to CommonField that returns the display context parameter as an instance of `DisplayContextParameter` other than a String. 
If we do that this will simplify in:
`field.getDisplayContextParameter.hasType(DisplayContextParameterType.DATETIMEENTRY)` 
This is much more reusable and will help in reducing duplication

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+            });
+
+            return newNode;
+        }
+
+        return collectionNode;
+    }
+
+    private TextNode createNode(String displayContextParameter, String valueToConvert, BaseType baseType, String fieldPath) {
+        String format = getDisplayContextParameterOfType(displayContextParameter, DisplayContextParameterType.DATETIMEENTRY)
```
mario-paniccia: `getDisplayContextParameterOfType(displayContextParameter, DisplayContextParameterType.DATETIMEENTRY)`
I think the code would be simpler if `CommonField` had the diplay context parameter already converted into a list of `DisplayContextParameter` rather than a raw String. Same as it has a `List<AccessControlList>`.
If that was the case, then `CommonField` could have a method `getDisplayContextParameterOfType(DisplayContextParameterType.DATETIMEENTRY)`

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+            ));
+        try {
+            if (baseType == BaseType.get(DATETIME)) {
+                return new TextNode(dateTimeFormatParser.convertDateTimeToIso8601(format, valueToConvert));
+            } else {
+                return new TextNode(dateTimeFormatParser.convertDateToIso8601(format, valueToConvert));
+            }
+        } catch (Exception e) {
+            throw new DataProcessingException().withDetails(
+                String.format("Unable to process field %s with value %s. Expected format: %s",
```
mario-paniccia: If I'm not wrong here we are failing the case update operation if the date field is not in the format specified in the DisplayContextParameter. This won't be an issue with the UI because it's going to send us the date fields in the correct format. 
But what about Services that are invoking the update event programmatically? They will still be passing unformatted date fields probably? will those calls throw exception? 
In case the date is already in ISO8601 format, does the `dateTimeFormatParser.convertDateTimeToIso8601` throws exception? 
We could think of allowing date fields that are already in ISO8601 format rather than throwing exception. 
We need to speak with BAs to understand the behaviour in this situation and state it clearly in the ticket for future reference

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+            return null;
+        }
+        ObjectNode newNode = MAPPER.createObjectNode();
+        complexCaseFields.stream().forEach(complexCaseField -> {
+            final BaseType complexFieldType = BaseType.get(complexCaseField.getFieldType().getType());
+            final String fieldId = complexCaseField.getId();
+            final JsonNode caseFieldNode = complexNode.get(fieldId);
+            final String fieldPath = fieldPrefix + FIELD_SEPARATOR + fieldId;
+
+            if (complexFieldType == BaseType.get(COLLECTION)) {
```
mario-paniccia:
```suggestion
            if (complexFieldType == BaseType.get(COLLECTION)) {
```
complexCaseField.isCollectionFieldType()

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+                newNode.set(fieldId, executeCollection(caseFieldNode, complexCaseField, fieldPath));
+            } else if (complexFieldType == BaseType.get(COMPLEX)) {
```
mario-paniccia:
```suggestion
            } else if (complexCaseField.isComplexFieldType()
) 
```

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+                                      String fieldPrefix) {
+        if (complexNode == null) {
+            return null;
+        }
+        ObjectNode newNode = MAPPER.createObjectNode();
+        complexCaseFields.stream().forEach(complexCaseField -> {
+            final BaseType complexFieldType = BaseType.get(complexCaseField.getFieldType().getType());
+            final String fieldId = complexCaseField.getId();
+            final JsonNode caseFieldNode = complexNode.get(fieldId);
+            final String fieldPath = fieldPrefix + FIELD_SEPARATOR + fieldId;
```
mario-paniccia: building a field's path it's common logic that could be encapsulated in a common class or utility method. If we don't have something we can reuse already

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+
+            if (complexFieldType == BaseType.get(COLLECTION)) {
+                newNode.set(fieldId, executeCollection(caseFieldNode, complexCaseField, fieldPath));
+            } else if (complexFieldType == BaseType.get(COMPLEX)) {
+                Optional.ofNullable(
+                    executeComplex(caseFieldNode, complexCaseField.getFieldType().getComplexFields(), caseEventField, wizardPageField, fieldPath))
+                    .ifPresent(result -> newNode.set(fieldId, result));
```
mario-paniccia: I wouldn't bother wrapping a null with an optional just so I can call `ifPresent`. In this case a null check is simpler to read IMHO:
```
JsonNode jsonNode = executeComplex(caseFieldNode, complexCaseField.getFieldType().getComplexFields(), caseEventField, wizardPageField,
                    fieldPath);
                if (jsonNode != null) {
                    newNode.set(fieldId, jsonNode);
                }
```

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeFormatParser.java
```diff
@@ -0,0 +1,44 @@
+package uk.gov.hmcts.ccd.domain.service.processor;
```
mario-paniccia: maybe better move this into a `processor.date` package?

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DisplayContextParameter.java
```diff
@@ -0,0 +1,61 @@
+package uk.gov.hmcts.ccd.domain.service.processor;
```
mario-paniccia: DisplayContextParameter it's a core domain concept that should be outside the processor package

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DisplayContextParameterType.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final int TYPE_GROUP = 1;
+    private static final int VALUE_GROUP = 2;
+
+    public static Optional<DisplayContextParameterType> getParameterTypeFor(String displayContextParameter) {
+        Matcher m = PATTERN.matcher(displayContextParameter);
+        if (m.matches()) {
+            try {
+                return Optional.of(valueOf(m.group(TYPE_GROUP)));
+            } catch (IllegalArgumentException e) {
+                return Optional.empty();
```
mario-paniccia: Optional.empty should be used to model missing values, not incorrect or unknown values as we are doing here

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DisplayContextParameterType.java
```diff
@@ -0,0 +1,38 @@
...
+            }
+        }
+        return Optional.empty();
```
mario-paniccia: Optional.empty should be used to model missing values, not incorrect or unknown values as we are doing here

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DisplayContextParameterType.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public static Optional<String> getParameterValueFor(String displayContextParameter) {
+        Matcher m = PATTERN.matcher(displayContextParameter);
+        if (m.matches() && !Strings.isNullOrEmpty(m.group(VALUE_GROUP))) {
+            return Optional.of(m.group(VALUE_GROUP));
+        }
+        return Optional.empty();
```
mario-paniccia: Optional.empty should be used to model missing values, not incorrect or unknown values as we are doing here

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DisplayContextParameterType.java
```diff
@@ -0,0 +1,38 @@
...
+            } catch (IllegalArgumentException e) {
+                return Optional.empty();
+            }
+        }
+        return Optional.empty();
+    }
+
+    public static Optional<String> getParameterValueFor(String displayContextParameter) {
+        Matcher m = PATTERN.matcher(displayContextParameter);
+        if (m.matches() && !Strings.isNullOrEmpty(m.group(VALUE_GROUP))) {
```
mario-paniccia: is the null check required here? if value group of the pattern (.+) is matching that means there is at least on char right?

src/main/java/uk/gov/hmcts/ccd/domain/service/validate/DefaultValidateCaseFieldsOperation.java
```diff
@@ -11,33 +11,38 @@
...
         final CaseType caseType = caseDefinitionRepository.getCaseType(caseTypeId);
         if (caseType == null) {
-            throw new ValidationException("Cannot find case type definition for  " + caseTypeId);
+            throw new ValidationException("Cannot find case type definition for " + caseTypeId);
         }
         if (!hasEventId(caseType, content.getEventId())) {
-            throw new ValidationException("Cannot validate case field because of event" + content.getEventId() + " is not found in case type definition");
+            throw new ValidationException("Cannot validate case field because of event " + content.getEventId() + " is not found in case type definition");
         }
+        content.setData(fieldProcessorService.processData(content.getData(), caseType, content.getEventId()));
```
mario-paniccia: I'm a bit puzzled by this choice. I would have imagined we would extend the DateTimeValidator and DateValidator to do validation of the formatted date. And if successful then convert. 
Instead we are converting to the internal format before doing the validation

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/DataProcessingException.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmcts.ccd.endpoint.exceptions;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.UNPROCESSABLE_ENTITY)
+public class DataProcessingException extends ValidationException {
+
+    public DataProcessingException() {
+        super("Processing of data failed");
```
mario-paniccia: I think error reporting must provide more details about the cause of the failure. This message is too generic and won't help investigating the issue

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/DataProcessingException.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmcts.ccd.endpoint.exceptions;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.UNPROCESSABLE_ENTITY)
+public class DataProcessingException extends ValidationException {
```
mario-paniccia: why DataProcessing exception should be a subclass of ValidationException? in theory we could have processing errors that have nothing to do with validation

src/test/java/uk/gov/hmcts/ccd/domain/service/createcase/DefaultCreateCaseOperationTest.java
```diff
@@ -169,6 +169,33 @@ void shouldThrowValidationException_whenCaseTypeIsNotDefinedForJurisdiction() {
                      "Cannot create case because of cti is not defined as case type for jid");
```
mario-paniccia: not enough unit testing probably

mario-paniccia: and we absolutely must have a good suite of integration tests

src/main/java/uk/gov/hmcts/ccd/domain/service/createcase/DefaultCreateCaseOperation.java
smathangi: resolving conversation as it's nothing particular relate to this PR.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/FieldType.java
```diff
@@ -26,6 +26,8 @@
     public static final String PREDEFINED_COMPLEX_ADDRESS_UK = "AddressUK";
     public static final String PREDEFINED_COMPLEX_ORDER_SUMMARY = "OrderSummary";
     public static final String PREDEFINED_COMPLEX_CASELINK = "CaseLink";
+    public static final String DATETIME = "DateTime";
```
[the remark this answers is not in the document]
smathangi: Agree. But that'll be an improvement on existing code - nothing specific to this PR changes.

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeEntryProcessor.java
smathangi: done

PR #851, opened 2020-03-17
RDM-7692 - DateTime Display Formatting

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-7692

### Change description ###
DateTime Display Formatting

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CommonField.java
```diff
@@ -29,6 +30,10 @@
 
     void setDisplayContextParameter(String displayContextParameter);
```
mario-paniccia: we have a domain object DisplayContext. Can we use that instead of String?

mario-paniccia: we have a model object DisplayContextParameter. Can we use that instead of String? it would reduce the boilerplate in the code

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CommonField.java
```diff
@@ -44,6 +49,13 @@ default boolean isCompoundFieldType() {
         return isCollectionFieldType() || isComplexFieldType();
     }
 
+    default DisplayContext displayContextType() {
+        return Optional.ofNullable(getDisplayContext())
+            .filter(dc -> !dc.equals("HIDDEN"))
+            .map(DisplayContext::valueOf)
+            .orElse(null);
```
mario-paniccia: this method in case the value of display context is "HIDDEN" is going to return `null`. Which is logically wrong.

mario-paniccia: I would also avoid wrapping nulls into Optional just to be able to use it's api. In the end is probably making the code more complicated to read

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/WizardPageComplexFieldOverride.java
```diff
@@ -67,4 +68,11 @@ public String getDisplayContextParameter() {
     public void setDisplayContextParameter(String displayContextParameter) {
         this.displayContextParameter = displayContextParameter;
     }
+
+    public DisplayContext displayContextType() {
+        return Optional.ofNullable(getDisplayContext())
+            .filter(dc -> !dc.equals("HIDDEN"))
+            .map(DisplayContext::valueOf)
+            .orElse(null);
+    }
```
mario-paniccia: same as above

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeValueFormatter.java
```diff
@@ -0,0 +1,133 @@
...
+            ArrayNode newNode = MAPPER.createArrayNode();
+            collectionNode.forEach(item -> {
+                JsonNode newItem = item.deepCopy();
+                ((ObjectNode)newItem).replace(CollectionValidator.VALUE,
+                    createNode(caseViewField.getDisplayContextParameter(), item.get(CollectionValidator.VALUE).asText(), collectionFieldType, fieldPath, displayContext));
+                newNode.add(newItem);
+            });
+
+            return newNode;
+        }
```
mario-paniccia: the logic that loops through collection fields and navigates complex fields is common logic that applies to all processors. So it should be made generic and written only once rather than being inside each single subclass of CaseViewFieldProcessor. But we can ignore this comment for now as I have raised a separate ticket to introduce a generic Processor

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/AbstractDefaultGetCaseViewOperation.java
```diff
@@ -79,6 +83,7 @@ CaseDetails getCaseDetails(String caseReference) {
             CommonField[] caseViewFields = tab.getTabFields().stream()
                 .filter(filterCaseTabFieldsBasedOnSecureData(caseDetails))
                 .map(caseTypeTabField -> CaseViewField.createFrom(caseTypeTabField, data))
+                .map(fieldProcessorService::processCaseViewField)
                 .toArray(CaseViewField[]::new);
```
mario-paniccia: minor:
this method is becoming a bit too long. It's generally not recommended to have long sequences of chained methods because it makes the code difficult to understand. It could be useful breaking this down

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/CaseViewFieldProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+import static uk.gov.hmcts.ccd.domain.model.definition.FieldType.COMPLEX;
+
+public abstract class CaseViewFieldProcessor extends FieldProcessor {
+
+    public CaseViewFieldProcessor(CaseViewFieldBuilder caseViewFieldBuilder) {
+        super(caseViewFieldBuilder);
+    }
+
+    public CaseViewField execute(CaseViewField caseViewField) {
+        final BaseType fieldType = BaseType.get(caseViewField.getFieldType().getType());
```
mario-paniccia: we could reduce boilerplate by adding a 
isCollectionFieldType() and isComplexFieldType() methods to `FieldType`

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessorService.java
```diff
@@ -66,6 +81,27 @@ public FieldProcessorService(final List<CaseDataFieldProcessor> caseDataFieldPro
...
+        return uiDefinitionRepository.getWizardPageCollection(caseTypeId, eventId)
+            .stream()
+            .map(WizardPage::getWizardPageFields)
+            .flatMap(List::stream)
+            .collect(Collectors.toList());
+    }
+
+    private WizardPageField wizardPageField(List<WizardPageField> wizardPageFields, String fieldId) {
+        return wizardPageFields.stream().filter(f -> f.getCaseFieldId().equals(fieldId)).findAny().orElse(null);
+    }
```
mario-paniccia: minor:
I can see more and more often we need to find a wizard field corresponding to a certain field id. Probably we could introduce a method on the UIDefinitionRepository that returns a map of the wizard fields by fieldID

src/main/java/uk/gov/hmcts/ccd/domain/service/processor/DateTimeValueFormatter.java
```diff
@@ -0,0 +1,133 @@
...
+                    return new TextNode(dateTimeFormatParser.convertIso8601ToDate(format, valueToConvert));
+                }
+            } catch (Exception e) {
+                throw new DataProcessingException().withDetails(
+                    String.format("Unable to process field %s with value %s. Expected format: %s",
+                        fieldPath,
+                        valueToConvert,
+                        format)
+                );
+            }
```
mario-paniccia: duplicated

src/test/java/uk/gov/hmcts/ccd/domain/service/processor/FieldProcessorServiceTest.java
```diff
@@ -0,0 +1,238 @@
+package uk.gov.hmcts.ccd.domain.service.processor;
+
```
mario-paniccia: we should add a good bunch of  integration tests to protect ourselves from regressions

PR #852, opened 2020-03-19
Assign Jurisdiction Functional Tests

noronhaa:
### Change description ###
Functional tests for main Assign jurisdiction story (RDM-7162)

```
[ ] Yes
[x] No
```


src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -1,36 +1,54 @@
...
+    "pathVariables": {
+      "uid": "[[DYNAMIC]]"
+    }
+  },
+  "expectedResponse": {
+    "body": {
+      "user": {
+        "idam": {
+          "id": "[[ANYTHING_PRESENT]]",
+          "email": "[[ANYTHING_PRESENT]]",
```
MSancaktutar: This shouldn't be ANYTHING_PRESENT. We can know what to expect for this attribute in the response, and it is the same env variable as what we used for the request.

src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -1,36 +1,54 @@
...
+  "request": {
+    "pathVariables": {
+      "uid": "[[DYNAMIC]]"
+    }
+  },
+  "expectedResponse": {
+    "body": {
+      "user": {
+        "idam": {
+          "id": "[[ANYTHING_PRESENT]]",
```
MSancaktutar: This shouldn't be ANYTHING_PRESENT. We can know what to expect for this attribute in the response, and it is the same uid as what we used for the request.

src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -1,36 +1,54 @@
...
+          "email": "[[ANYTHING_PRESENT]]",
+          "forename": "[[ANYTHING_PRESENT]]",
+          "surname": "[[ANYTHING_PRESENT]]",
```
MSancaktutar: These should also be consistent predefined values we can specifically expect in the response.

src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -1,36 +1,54 @@
...
+          "roles": "[[ANYTHING_PRESENT]]",
```
MSancaktutar: same as above

src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -1,36 +1,54 @@
...
+          "defaultService": null
+        }
+      },
+      "channels": null,
+      "jurisdictions": [
+        {
+          "id": "BEFTA_JURISDICTION_1",
+          "name": "BEFTA_JURISDICTION_1",
+          "description": "Content for the Test Jurisdiction.",
+          "caseTypes": "[[ANYTHING_PRESENT]]"
```
MSancaktutar: casetypes under BEFTA_JURISDICTION_1 are predefined. Please expressly list them as expected.

src/aat/resources/features/F-051/F-051.feature
```diff
@@ -16,13 +16,12 @@ Background:
...
-  Scenario: must return appropriate negative response for a user not having a profile in CCD
+  Scenario: must return appropriate user setting from idam role for a user not having a profile in CCD
     Given a user with [no profile in CCD]
     When a request is prepared with appropriate values
-    And the request [uses a Case ID that doesn’t exist in CCD]
     And it is submitted to call the [Get default settings for user] operation of [CCD Data Store]
-    Then a negative response is received
-    And the response [has the 403 return code]
+    Then a positive response is received
+    And the response [has the 200 return code]
```
MSancaktutar: This is a potential breaking change and should be checked if there is any consumer of this operation relying on the previous behaviour.

MSancaktutar: The line 24 can be removed, since, from a behavioural specification point of view, "Then a positive response is received" would be sufficient.

src/aat/resources/features/F-102/F-102.feature
```diff
@@ -50,3 +50,14 @@ Feature: F-102: Get jurisdictions available to the user
...
     And the response has all other details as expected
+
+  @S-580
+  Scenario: must return a list of jurisdictions for a valid user with no user profile
+    Given a user with [appropriate idam roles but no CCD user profile]
+    When a request is prepared with appropriate values
+    And the request [has CREATE as case access parameter]
+    And it is submitted to call the [Get jurisdictions available to the user] operation of [CCD Data Store]
+    Then a positive response is received
+    And the response [contains HTTP 200 Ok status code]
```
MSancaktutar: line 61 can be removed

src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -1,36 +1,54 @@
...
+  "method": "GET",
+  "uri": "/caseworkers/{uid}/profile",
+  "specs": [
+    "no profile in CCD",
+    "uses the Case ID of the case just created",
+    "has the 200 return code"
+  ],
+  "user": {
+    "username": "befta.caseworker.1.noprofile@gmail.com",
+    "password": "[[$CCD_BEFTA_CASEWORKER_1_PWD]]"
```
MSancaktutar: password used is for another user

src/aat/resources/features/F-102/S-580.td.json
```diff
@@ -0,0 +1,78 @@
...
+  "title": "Get jurisdictions available to the user",
+  "specs": [
+    "appropriate idam roles but no CCD user profile",
+    "has CREATE as case access parameter",
+    "contains HTTP 200 Ok status code",
+    "contains the list of jurisdictions a user has access to"
+  ],
+  "user": {
+    "username": "befta.caseworker.1.noprofile@gmail.com"
+  },
```
MSancaktutar: this will inherit the password of another user specified in the parent object, which is not good. every user should use its own password env variable.

src/aat/resources/features/F-102/S-580.td.json
```diff
@@ -0,0 +1,78 @@
...
+          "description": "Content for the Test Jurisdiction.",
+          "caseTypes": [
+            {
+              "id": "BEFTA_CASETYPE_1_1",
+              "description": "Create a case of type BEFTA_CASETYPE_1_1",
+              "version": null,
+              "name": "BEFTA Case Type 1 1",
+              "jurisdiction": null,
+              "events": "[[ANYTHING_PRESENT]]",
+              "states": "[[ANYTHING_PRESENT]]",
```
MSancaktutar: these can be expected to be specific values, couldn't them?

src/aat/resources/features/F-102/S-580.td.json
```diff
@@ -0,0 +1,78 @@
...
+              ]
+            },
+            {
+              "id": "CASE_TYPE_WITH_NO_CASES",
+              "description": "Create a case of type CASE_TYPE_WITH_NO_CASES",
+              "version": null,
+              "name": "CT With No Cases -Don't Create",
+              "jurisdiction": null,
+              "events": "[[ANYTHING_PRESENT]]",
+              "states": "[[ANYTHING_PRESENT]]",
```
MSancaktutar: these can be expected to be specific values, couldn't them?

src/aat/resources/features/F-102/S-580.td.json
```diff
@@ -0,0 +1,78 @@
...
+              "searchAliasFields": [],
+              "security_classification": null,
+              "case_fields": [],
+              "printable_document_url": null,
+              "acls": [
+                {
+                  "role": "caseworker-befta_jurisdiction_1",
```
MSancaktutar: better make a dynamic reference to the field in the request, to avoid copy-paste duplication of the same value.

src/aat/resources/features/F-051/F-051.feature
noronhaa: will do, someone may need to retrospectively go through all current scenarios as those 2 lines are seen together throughout the test suite

src/aat/resources/features/F-102/S-580.td.json
noronhaa: we are specifically looking for the list of jurisdictions here specifically

MSancaktutar: The response should be checked in full. Returning a wrong list of events and states is a bug in the api even if we return the correct data in the rest of the response.

src/aat/resources/features/F-051/F-051.feature
MSancaktutar: you're right in that.

src/aat/resources/features/F-051/S-109.td.json
MSancaktutar: ${[scenarioContext][theInvokingUser][id]}

src/aat/resources/features/F-102/S-580.td.json
MSancaktutar: forget about it. that is not a data in the scenario context

src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -1,36 +1,481 @@
...
+    "pathVariables": {
+      "uid": "[[DYNAMIC]]"
+    }
+  },
+  "expectedResponse": {
+    "body": {
+      "user": {
+        "idam": {
+          "id": "${[scenarioContext][theInvokingUser][id]}",
+          "email": "befta.caseworker.1.noprofile@gmail.com",
```
MSancaktutar: we could use
`"email": "${[scenarioContext][theInvokingUser][username]}",`

src/aat/resources/features/F-102/F-102.feature
```diff
@@ -50,3 +50,13 @@ Feature: F-102: Get jurisdictions available to the user
     And the response [contains HTTP 400 Bad Request]
     And the response [contains an error message : Access can only be 'create', 'read' or 'update']
     And the response has all other details as expected
+
+  @S-580
+  Scenario: must return a list of jurisdictions for a valid user with no user profile
```
MSancaktutar: must return a list of jurisdictions for a valid user with no **CCD** user profile

PR #862, opened 2020-03-25
RDM-7373 Lower severity of CaseValidationException

mattnayler:

[RDM-7373](https://tools.hmcts.net/jira/browse/RDM-7373): Lower log severity of Case Validation Errors

### Change description ###

**Logger calls**
* Merge validation results log entry with exception log entry.
* Use the lower severity (`Warn`).
* Log field names only to prevent user data being sent to logs.

**Application Insights telemetry**
* Add some validation result information to the exception telemetry.
* Set a lower severity level on the exception (`Warning`).

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/RestExceptionHandler.java
```diff
@@ -50,6 +60,29 @@ public RestExceptionHandler(AppInsights appInsights) {
...
+        List<String> fieldIds = new ArrayList<>();
+        if (exception.getDetails() != null) {
+            String detailsJson = this.objectMapperService.convertObjectToString(exception.getDetails());
+            fieldIds = JsonPath.read(detailsJson, "$..field_errors[*].id");
+        }
+
+        Map<String, String> customProperties = new HashMap<>();
+        customProperties.put("CaseValidationError field IDs", Arrays.toString(fieldIds.toArray()));
+
+        LOG.warn("{}: The following list of fields are in an error state: {}", exception.getMessage(), fieldIds, exception);
```
mario-paniccia, line 77: I was wondering whether it makes sense to not log the stacktrace for the the validation errors. What do you think @mattnayler?

mario-paniccia, line 77:
```suggestion
        LOG.warn("{}: The following list of fields are in an invalid state: {}", exception.getMessage(), fieldIds, exception);
```

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/RestExceptionHandler.java
```diff
@@ -50,6 +60,29 @@ public RestExceptionHandler(AppInsights appInsights) {
...
+    @ExceptionHandler(CaseValidationException.class)
+    @ResponseBody
+    public ResponseEntity<HttpError> handleCaseValidationException(HttpServletRequest request, CaseValidationException exception) {
+
+        // read field IDs from CaseValidationException.details (i.e. avoid using full error details as this may contain user data)
+        List<String> fieldIds = new ArrayList<>();
+        if (exception.getDetails() != null) {
+            String detailsJson = this.objectMapperService.convertObjectToString(exception.getDetails());
+            fieldIds = JsonPath.read(detailsJson, "$..field_errors[*].id");
+        }
```
mario-paniccia, line 72: wonder whether we can enhance CaseValidationException with a property that stores fields with validation error so that here rather than manipulating JSON we could just call: `exception.getFields()` and we get back a list of field ids

src/main/java/uk/gov/hmcts/ccd/AppInsights.java
```diff
@@ -33,6 +37,20 @@ public void trackException(Exception e) {
         telemetry.trackException(e);
     }
 
+    public void trackException(Exception e, Map<String, String> customProperties, SeverityLevel severityLevel) {
```
mario-paniccia, line 40: mind adding a javadoc explaining the usage of custom properties

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/RestExceptionHandler.java
mattnayler: The same field level validation routines may be triggered in three separate places for a single UI call: 
* Initial event data submission
* After callback has returned new data
* After data has been merged with existing unchanged case details (i.e. before case event is saved)

So whilst the data submitted to AppInsights will always contain the full stack details it may still be useful for a dev to see their stack trace is different to the one they are investigating (i.e. it was for me during RDM-7840).  So I am currently tempted to leave it but also happy to be overruled, @mario-paniccia?

mario-paniccia: right, good point. Let's leave it then cheers

mattnayler: Thanks that sounds much cleaner. Updated in next push.

src/main/java/uk/gov/hmcts/ccd/AppInsights.java
mattnayler: Updated to mirror the javadoc comments used in the wrapped Microsoft components.

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/CaseValidationException.java
```diff
@@ -2,11 +2,36 @@
...
 
-    public CaseValidationException() {
+    private final String[] fields;
+
+    public CaseValidationException(List<CaseFieldValidationError> fieldErrors) {
         super("Case data validation failed");
+
+        String[] fieldIds = new String[0];
+
+        if (fieldErrors != null && !fieldErrors.isEmpty()) {
```
mario-paniccia, line 20: don't think this null check and empty check is really needed

src/main/java/uk/gov/hmcts/ccd/endpoint/exceptions/CaseValidationException.java
```diff
@@ -2,11 +2,36 @@
...
 @ResponseStatus(code = HttpStatus.UNPROCESSABLE_ENTITY)
 public class CaseValidationException extends ValidationException {
 
-    public CaseValidationException() {
+    private final String[] fields;
+
+    public CaseValidationException(List<CaseFieldValidationError> fieldErrors) {
         super("Case data validation failed");
+
+        String[] fieldIds = new String[0];
```
mario-paniccia, line 18: I'd recommend to prefer List than Array in Java. The perf gain of using Array is almost 0 but the boilerplate in the code is much higher because working with arrays in Java is very cumbersome

PR #865, opened 2020-03-25
Rdm 8001

Thor-tech-of-metal:
RDM-8001: CCD-data-Store -Fix CVEs general and straight forward changes
    
          [RDM-8001] (https://tools.hmcts.net/jira/browse/RDM-8001).
        CCD-data-Store -Fix CVEs general and straight forward changes

[ X] Yes
[ ] No


build.gradle
```diff
@@ -251,7 +251,7 @@ dependencies {
     compile(group: 'io.searchbox', name: 'jest', version: '6.3.1') {
         exclude group: 'org.apache.logging.log4j', module: 'log4j-core'
     }
-    compile(group: 'org.elasticsearch', name: 'elasticsearch', version: '6.4.2') {
+    compile(group: 'org.elasticsearch', name: 'elasticsearch', version: '7.4.0') {
```
mario-paniccia, line 254: careful here. The version we use of the ElasticSearch server on Prod is 6.4.2
You are updating the version of the client library Data Store uses to talk to the server to a higher version respect to the server. Please make sure this won't cause issues. Can you check the documentation? Not sure if v7 of the client library is 100% compatible with v6 of the server. Most likely yes, but we need to double check

Thor-tech-of-metal: This is one of the things that I wanted to discuss with you tomorrow. It seems that the latest version does not use type any more and we use _doc. I added these comments in the ticket

Thor-tech-of-metal: I am going to remove this change to use elastic '6.4.2'

PR #866, opened 2020-03-25
Rdm 7943 jackson

Thor-tech-of-metal:
RDM-7943-JACKSON: Fix Suppressed CVEs CCD-data-Store

      [RDM-7943] (https://tools.hmcts.net/jira/browse/RDM-7943).
    Jackson changes to fix TESTs.

[ X] Yes
[ ] No


src/main/java/uk/gov/hmcts/ccd/config/JacksonUtils.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+
+import java.util.HashMap;
+
+
+public class JacksonUtils {
```
mario-paniccia, line 14: Great job there in removing all the duplication in the creation of ObjectMappers 👏 
I think we already addressed the issue of duplicated ObjectMappers in DataStore in the past.
The solution we proposed was to configure all the ObjectMappers the application uses in:
https://github.com/hmcts/ccd-data-store-api/blob/16bbdebdc08a8ab631bcf3070d2d06b03cf61111/src/main/java/uk/gov/hmcts/ccd/config/JacksonObjectMapperConfig.java#L12

Then these get injected, for example:
https://github.com/hmcts/ccd-data-store-api/blob/17c6d53b194d7be60277669edff6ffbbaeafe652/src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseSearchOperation.java#L49
Perhaps injecting them is a better approach than using Utils classes with static methods? Not saying we should change the current solution, just want to suggest an alternative.
You could refactor this class to a bean called for example `JsonService` that it's configured in `JacksonObjectMapperConfig` as well and then itself can get injected? This would also make testability simpler.
I appreciate that would be more changes though so I'm not saying we have to do it. Just wanted to highlight the solution we had agreed in the past

mario-paniccia: Actually I've noticed now that we also introduced a service that provides common utility methods:
https://github.com/hmcts/ccd-data-store-api/blob/3310d47146cd9f2f7bf43c9b546378e1176974b6/src/main/java/uk/gov/hmcts/ccd/domain/service/common/DefaultObjectMapperService.java#L19

I think we are mixing up 2 separate approaches here. I'd spent some time thinking of which one is the best and then use that consistenly

PR #869, opened 2020-03-26
RDM-6916 Create indexes for CaseEvent table

mattnayler:

  [RDM-6916](https://tools.hmcts.net/jira/browse/RDM-6916): "Create indexes for CaseEvent table to improve query performance"

### Change description ###

Create indexes for CaseEvent table to improve query performance:
* Conditional DROP followed by CREATE of idx_case_event_created_date.
* Create the following new indexes for:
  * case_type_id
  * event_id, state_id

```
[ ] Yes
[x] No
```


src/main/resources/db/changelog/db.changelog-RDM-6916.xml
```diff
@@ -0,0 +1,43 @@
...
+        </preConditions>
+        <dropIndex indexName="idx_case_event_created_date"
+                   schemaName="public"
+                   tableName="case_event" />
+    </changeSet>
+
+
+    <changeSet id="rdm-6916" author="matt.nayler@hmcts.net">
+
+        <createIndex indexName="idx_case_event_created_date"
```
banderous: Suggest you use Postgres' CONCURRENTLY modifier when creating these indexes; standard index creation will lock the case_event table for writes until the indexes are built, which could take time since it's a big table. You'd probably need to write <sql>, don't think liquibase supports the concurrently flag.

mattnayler: Many thanks for your comment.  Some [timing details from a series of test runs](https://tools.hmcts.net/jira/browse/RDM-6916?focusedCommentId=739903&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-739903) have been added to the Jira ticket associated to this PR and will be used during the review and checkpoint process to decide the way forward.

NB: A separate set of scripts implementing the CONCURRENTLY modifier as you have recommended have been prepared in advance if the change is required.

PR #899, opened 2020-04-09
RDM-8113 handle display without seconds or milliseconds

RebeccaBaker:
https://tools.hmcts.net/jira/browse/RDM-8113

### Change description ###
RDM-8113 handle display without seconds or milliseconds
RDM-8057 split DCP correctly when there is multiple params

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/aat/resources/features/F-051/S-109.td.json
+++ b/src/aat/resources/features/F-051/S-109.td.json
@@ -23,7 +23,7 @@
   "expectedResponse": {
     "headers": {
       "Content-Encoding" : "gzip",
-      "Content-Length" : "857",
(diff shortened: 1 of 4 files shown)


src/aat/resources/features/F-051/S-109.td.json
```diff
@@ -23,7 +23,7 @@
   "expectedResponse": {
     "headers": {
       "Content-Encoding" : "gzip",
-      "Content-Length" : "857",
+      "Content-Length" : "[[ANYTHING_PRESENT]]",
```
MSancaktutar, line 26: ANY_INTEGER_PRESENT would be better

PR #900, opened 2020-04-09
RDM-8166 - Using Formatted Date(Time) Fields on Workbasket & Search Input Screens

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-8166

### Change description ###
Using Formatted Date(Time) Fields on Workbasket & Search Input Screens

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/search/Field.java
```diff
@@ -41,4 +47,13 @@ public boolean isMetadata() {
     public void setMetadata(boolean metadata) {
         this.metadata = metadata;
     }
+
+    public Optional<CommonField> getNestedField(String path) {
+        final List<String> pathElements = Arrays.asList(path.trim().split("\\."));
```
kiran-yenigala-hmcts, line 52: if we can move this to FieldType that will be keep this class simple

src/main/java/uk/gov/hmcts/ccd/domain/model/search/SearchResultViewColumn.java
```diff
@@ -73,32 +70,11 @@ public String getDisplayContextParameter() {
     }
 
     public Optional<CommonField> getNestedField(String path) {
-        if (StringUtils.isBlank(path) || path.trim().split("\\.").length == 1 || this.getCaseFieldType().getChildren().isEmpty()) {
+        final List<String> pathElements = Arrays.asList(path.trim().split("\\."));
```
kiran-yenigala-hmcts, line 73: Move this method to FieldType class

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/FieldType.java
```diff
@@ -129,6 +134,34 @@ public void setCollectionFieldType(FieldType collectionFieldType) {
...
+            CommonField caseField = optionalCaseField.get();
+
+            if (pathElements.size() == 1) {
+                return Optional.of(caseField);
+            } else {
+                List<CaseField> newCaseFields = caseField.getFieldType().getChildren();
+                List<String> tail = pathElements.subList(1, pathElements.size());
+
+                return reduce(newCaseFields, tail);
+            }
```
kiran-yenigala-hmcts, line 159: instead of subList() better to use index like below then it will give some performance benefit as well.
reduce(List<CaseField> caseFields, List<String> pathElements, int startIndex)
........
return reduce(newCaseFields, pathElements, index + 1);

danlysiak: Won't action this as part of this ticket as is part of an existing implementation that was copied over from `CommonField`. In latest commit I've updated `CommonField` to now use this one as well so it's not all duplicated.

PR #909, opened 2020-04-17
RDM-8051 Log and audit: log audit events in Data Store

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-8051

### Change description ###
Log and Audit functionality based on the Spring Aspects and Interceptor

```
[ ] Yes
[X ] No
```


src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
```diff
@@ -0,0 +1,169 @@
...
+            (isNotBlank(idamId) ? ", idamId:" + idamId : "") +
+            (isNotBlank(invokingService) ? ", invokingService:" + invokingService : "") +
+            (isNotBlank(path) ? ", endpointCalled:" + httpMethod + " " + path : "") +
+            ", operationOutcome:" + httpStatus +
+            (isNotBlank(caseType) ? ", caseType:" + caseType : "") +
+            (isNotBlank(jurisdiction) ? ", jurisdiction:" + jurisdiction : "") +
+            (isNotBlank(eventSelected) ? ", eventSelected:" + eventSelected : "") +
+           (isNotBlank(targetIdamId) ? ", idamIdOfTarget:" + targetIdamId : "") +
+//            (isNotBlank(listOfCaseTypes) ? ", listOfCaseTypes:" + listOfCaseTypes : "") +
+            (targetCaseRoles != null && !targetCaseRoles.isEmpty() ? ", targetCaseRoles:" +
```
Thor-tech-of-metal, line 159: Remove the commented lines .

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private final Clock clock;
+    private final SecurityUtils securityUtils;
+    private final UserRepository userRepository;
+    private final AuditRepository auditRepository;
+
+    public AuditService(@Qualifier("utcClock") final Clock clock,
+                        @Qualifier(CachedUserRepository.QUALIFIER) final UserRepository userRepository,
+                        @Lazy final SecurityUtils securityUtils, final AuditRepository auditRepository) {
+        this.clock = clock;
```
Thor-tech-of-metal, line 28: Any particular reason why @Lazy final SecurityUtils securityUtils is lazy. Because it seems that this securityUtils attribute is important for the AuditService functionality.   without securityUtils AuditService will not work.   Maybe it is good if it always present.   If securityUtils were inside an if statement yes .

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
```diff
@@ -0,0 +1,61 @@
...
+            entry.setJurisdiction(auditContext.getJurisdiction());
+            entry.setCaseId(auditContext.getCaseId());
+            entry.setCaseType(auditContext.getCaseType());
+            entry.setEventSelected(auditContext.getEventName());
+            entry.setTargetIdamId(auditContext.getTargetIdamId());
+            entry.setTargetCaseRoles(auditContext.getTargetCaseRoles());
+        }
+
+        entry.setRequestId(request.getHeader("request-id"));
+
```
Thor-tech-of-metal, line 57: I think that there is a constant ion the model for "request-id" value .  If not maybe we can create one . Only if there is a shared common area for test modules and business modules.

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/AuditAspect.java
```diff
@@ -0,0 +1,74 @@
...
+    private ExpressionEvaluator evaluator = new ExpressionEvaluator();
+
+    @Around("@annotation(logAudit)")
+    public Object audit(ProceedingJoinPoint joinPoint, LogAudit logAudit) throws Throwable {
+        Object result = null;
+        try {
+            result = joinPoint.proceed();
+            return result;
+        } catch (Exception ex) {
+            throw ex;
```
Thor-tech-of-metal, line 37: there is a logger "LOG" it would be good if we log the trace and a message.  We know where it is coming from

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/ExpressionEvaluator.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public Object getObject() {
+            return object;
+        }
+
+        public Object[] getArgs() {
+            return args;
+        }
+    }
```
Thor-tech-of-metal, line 65: public Object[] getArgs() {. }.  It is not used anywhere and it is not overwriting .

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditRepository.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts.ccd.auditlog;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AuditRepository {
```
mario-paniccia, line 8: better to have an AuditRepository interface with at the moment one implementation LoggerAuditRepository. 
This will allow in the future to easily change the destination of the log files if needed. For example they might want to ask to store on DB rather than on std out

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
```diff
@@ -0,0 +1,169 @@
...
+        return targetCaseRoles;
+    }
+
+    public void setTargetCaseRoles(List<String> targetCaseRoles) {
+        this.targetCaseRoles = targetCaseRoles;
+    }
+
+    @Override
+    public String toString() {
+        return TAG + " " + dateTime + " operationType:" + operationType +
```
mario-paniccia, line 148: I'm thinking that maybe for consistency we should have a field id for the datetime, like `datetime:` 
This will also simplify the processing of the records.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditInterceptor.java
```diff
@@ -0,0 +1,31 @@
...
+        this.auditService = auditService;
+    }
+
+    @Override
+    public void afterCompletion(HttpServletRequest request, HttpServletResponse response, Object handler,
+                           @Nullable Exception ex) {
+        if (handler instanceof HandlerMethod && ((HandlerMethod) handler).hasMethodAnnotation(LogAudit.class)) {
+            AuditContext auditContext = AuditContextHolder.getAuditContext();
+            // FIXME : suppress 404 and 403/401 requests
+            auditService.audit(request, response.getStatus(), auditContext);
```
Thor-tech-of-metal, line 26: // FIXME : suppress 404 and 403/401 requests. is this going to be done in this implementation ?

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
```diff
@@ -0,0 +1,169 @@
...
+            (targetCaseRoles != null && !targetCaseRoles.isEmpty() ? ", targetCaseRoles:" +
+                targetCaseRoles.stream().map(String::toString).collect(Collectors.joining(",")) : "") +
+            (isNotBlank(requestId) ? ", X-Request-ID:" + requestId : "") +
+            '}';
+    }
+
+    public String formattedAuditData() {
+        return  ReflectionToStringBuilder.toString(this, JSON_STYLE, false, false,
+            true, null).toString();
+    }
```
mario-paniccia, line 168: I would move this logic into a separate AuditEntryFormatter class for better separation of concerns. We could also more easily change the formatting

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
```diff
@@ -0,0 +1,169 @@
...
+           (isNotBlank(targetIdamId) ? ", idamIdOfTarget:" + targetIdamId : "") +
+//            (isNotBlank(listOfCaseTypes) ? ", listOfCaseTypes:" + listOfCaseTypes : "") +
+            (targetCaseRoles != null && !targetCaseRoles.isEmpty() ? ", targetCaseRoles:" +
+                targetCaseRoles.stream().map(String::toString).collect(Collectors.joining(",")) : "") +
+            (isNotBlank(requestId) ? ", X-Request-ID:" + requestId : "") +
+            '}';
+    }
+
+    public String formattedAuditData() {
+        return  ReflectionToStringBuilder.toString(this, JSON_STYLE, false, false,
```
mario-paniccia, line 166: we'd need to evaluate impact of performance if using reflection

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditInterceptor.java
```diff
@@ -0,0 +1,31 @@
...
+    private final AuditService auditService;
+
+    public AuditInterceptor(AuditService auditService) {
+        this.auditService = auditService;
+    }
+
+    @Override
+    public void afterCompletion(HttpServletRequest request, HttpServletResponse response, Object handler,
+                           @Nullable Exception ex) {
+        if (handler instanceof HandlerMethod && ((HandlerMethod) handler).hasMethodAnnotation(LogAudit.class)) {
```
mario-paniccia, line 23: in terms of design I wouldn't pass down web level classes like HttpServletRequest to the AuditService. Ideally AuditService should not be tied to any web class. We could have the interceptor parsing the request, adding all that info to the context, and invoking the AuditService with the context only so the AuditService knows nothing about HttpServletRequest  and HttpServletResponse

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
```diff
@@ -0,0 +1,61 @@
...
+        String formattedDate = LocalDateTime.now(clock).format(ISO_LOCAL_DATE_TIME);
+
+        entry.setDateTime(formattedDate);
+        entry.setHttpStatus(httpResponseStatus);
+        entry.setHttpMethod(request.getMethod());
+        entry.setPath(request.getRequestURI());
+        entry.setIdamId(userRepository.getUser().getEmail());
+        entry.setInvokingService(securityUtils.getServiceName());
+
+        if (auditContext != null) {
```
mario-paniccia, line 46: out of coursiosity when is it that auditContext can be null?

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
```diff
@@ -0,0 +1,169 @@
...
+
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.apache.commons.lang3.builder.ToStringStyle.JSON_STYLE;
+
+public class AuditEntry {
```
mario-paniccia, line 11: with this design it's not possible from the code to distinguish what properties we are logging for each operation type. 
I wonder whether it's possible to have an AuditEntry superclass with one implementation for each operation type that contains properties only relevant to that operation. So for example CaseSearchAudit, CaseUpdateAudit....

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditRepository.java
```diff
@@ -0,0 +1,15 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(AuditRepository.class);
+
+    public void save(final AuditEntry auditEntry) {
+        LOG.info(auditEntry.toString());
```
mario-paniccia, line 13: I think the AuditService could be injected with a AuditLogFormatter which is responsible for formatting the audit string. Then the Repository just logs the formatted entry

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/AuditAspect.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.expression.spel.SpelEvaluationException;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.ccd.auditlog.LogAudit;
+
+import java.lang.reflect.Method;
+import java.util.List;
+
+@Aspect
+@Component
+public class AuditAspect {
```
mario-paniccia, line 22: since we're using AOP I'll request some thorough performance testing on this ticket. Will add a comment on the ticket

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/AuditContextHolder.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.hmcts.ccd.auditlog.aop;
+
+public class AuditContextHolder {
+
+    private AuditContextHolder() {}
+
+    private static final InheritableThreadLocal<AuditContext> threadLocal = new InheritableThreadLocal<>();
```
mario-paniccia, line 7: ThreadLocal are a bit tricky and could cause mem leaks. For my knowledge, why we need to use it here? what problem is this addressing?

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CallbackEndpoint.java
```diff
@@ -30,6 +32,7 @@ public CallbackEndpoint(final PrintableDocumentListOperation printableDocumentLi
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = "Printable documents list retrieved")
     })
+    @LogAudit(operationType = OperationType.SEARCH_CASE)
```
mario-paniccia, line 35: mmm don't understand why this operation is being logged as a search

mario-paniccia: We could probably ignore this endpoint

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseAccessEndpoint.java
```diff
@@ -33,6 +35,7 @@ public CaseAccessEndpoint(CaseAccessOperation caseAccessOperation) {
         @ApiResponse(code = 200, message = "List of cases ids found"),
         @ApiResponse(code = 400, message = "Invalid case ID")
     })
+    @LogAudit(operationType = OperationType.SEARCH_CASE)
     public List<String> findCaseIdsGivenUserIdHasAccessTo(
```
mario-paniccia, line 39: we said to ignore this endpoint. If we want to log it I don't think we should log it as a search. Search case should be only for the search api I'd say

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
```diff
@@ -112,6 +114,7 @@ public CaseDetailsEndpoint(@Qualifier(CreatorGetCaseOperation.QUALIFIER) final G
         @ApiResponse(code = 400, message = "Invalid case ID"),
         @ApiResponse(code = 404, message = "No case found for the given ID")
     })
+    @LogAudit(operationType = OperationType.VIEW_CASE)
```
mario-paniccia, line 117: this endpoint is not used by the UI when the user is viewing a case. So the name of the operation might be a bit misleading. I wonder whether we want to introduce a separate operation name for when the case is being accessed programmatically like in this case. Rather than by the UI. I'll ask the Lee and the BAs

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
```diff
@@ -160,6 +164,7 @@ public CaseDetails findCaseDetailsForCitizen(
         @ApiResponse(code = 404, message = "No case found for the given ID"),
         @ApiResponse(code = 422, message = "Process could not be started")
     })
+    @LogAudit(operationType = OperationType.CREATE_CASE)
```
mario-paniccia, line 167: I think we agreed we wouldn't log these start event operations:
```Hi @Mario We have few questions for Log and Audit. https://tools.hmcts.net/confluence/display/RCCD/CCD+Case+Log+And+Audit+Endpoints+Data+Availability+-+V1
1.) Do we need to log this one: Case access API - GET /caseworkers/{uid}/jurisdictions/{jid}/case-types/{ctid}/cases/ids
2.) Do we need to log middle case creation steps, like Standard case API - GET /caseworkers/{uid}/jurisdictions/{jid}/case-types/{ctid}/event-triggers/{etid}/token  Those are trigerred on starting a new event (for Update or Create case), but not when we persist the case on the Submit.






Mario:house_with_garden:  12:32 PM
Hi @Rafal Kalita 1) no 2) no ```

File: src/main/java/uk/gov/hmcts/ccd/endpoint/std/CallbackEndpoint.java (hunk 17 above)
Comment (rafalkalita) [reply src/main/java/uk/gov/hmcts/ccd/endpoint/std/CallbackEndpoint.java:35]: This is a mistake. Will remove and in another PR with v1 will go as 
@LogAudit(operationType = OperationType.VIEW_CASE, caseId = "#caseDetails.reference", caseType = "#caseTypeId",
        jurisdiction = "#jurisdictionId")

File: src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java [hunk 21]
```diff
@@ -308,6 +318,7 @@ public CaseDetails saveCaseDetailsForCitizen(
         @ApiResponse(code = 422, message = "Field validation failed"),
         @ApiResponse(code = 409, message = "Case reference not unique")
     })
+    @LogAudit(operationType = OperationType.CREATE_CASE)
     public JsonNode validateCaseDetails(
```
mario-paniccia, line 322: we agreed not to log this

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseAccessEndpoint.java
rafalkalita: It is a mistake. V1 will be in another PR. Will remove from this PR.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseAccessEndpoint.java
```diff
@@ -58,6 +61,7 @@ public CaseAccessEndpoint(CaseAccessOperation caseAccessOperation) {
         @ApiResponse(code = 400, message = "Invalid case ID")
     })
     @ResponseStatus(value = HttpStatus.CREATED)
+    @LogAudit(operationType = OperationType.UPDATE_CASE_ACCESS)
```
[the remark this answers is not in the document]
rafalkalita: It is a mistake. V1 will be in another PR. Will remove from this PR.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
rafalkalita: We have initially committed this on v1 as well, and then asked questions. Will go to another PR.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
rafalkalita: Same as other comments. Will clean up and will be in v1 PR.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
```diff
@@ -385,6 +398,7 @@ public CaseDetails createCaseEventForCitizen(
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = "Documents list for the given case id")
     })
+    @LogAudit(operationType = OperationType.SEARCH_CASE)
     public List<Document> getDocumentsForCase(
```
mario-paniccia, line 402: don't think it's correct to classify this as a search. There is not requirement to log this in the design. We could ignore or if we want to log it then we need  a more suitable operation name, like CASE_DOCUMENTS_ACCESS?

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
rafalkalita: Please check and let us know. Same as other comments. Will clean up and will be in v1 PR.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
```diff
@@ -437,6 +453,7 @@ public CaseDetails createCaseEventForCitizen(
     @ApiOperation(value = "Get the pagination metadata for a case data search")
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = "Pagination metadata for the given search criteria")})
+    @LogAudit(operationType = OperationType.SEARCH_CASE)
```
mario-paniccia, line 456: I thing we can ignore this. We are already logging the search operation in searchCasesForCaseWorkers and searchCasesForCitizens. If we log here as well every user search would be logged twice which would be misleading. For one use operation I think it's better to have 1 log audit only.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
mario-paniccia: the reason I think it's best not to log the start of a case update it that it's better to have 1 log audit for each user operation. Because to update a case in CCD you need to call 2 endpoint, each user update would be result with 2 audit logs. Which is not nice

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseDetailsEndpoint.java
```diff
@@ -449,6 +466,7 @@ public PaginatedSearchMetadata searchCasesMetadataForCaseworkers(@PathVariable("
...
     public PaginatedSearchMetadata searchCasesMetadataForCitizens(@PathVariable("uid") final String uid,
```
mario-paniccia, line 470: ignore please

src/main/java/uk/gov/hmcts/ccd/endpoint/std/DraftsEndpoint.java
```diff
@@ -113,6 +117,7 @@ public DraftResponse updateDraftForCaseWorker(
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = "A displayable draft")
     })
+    @LogAudit(operationType = OperationType.CREATE_CASE)
```
mario-paniccia, line 120: UPDATE_CASE?

src/main/java/uk/gov/hmcts/ccd/auditlog/OperationType.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.hmcts.ccd.auditlog;
+
+public enum OperationType {
```
mario-paniccia, line 3: rename to AuditOperationType please

src/main/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpoint.java
```diff
@@ -218,6 +193,7 @@ public SearchResultView searchNew(@PathVariable("jid") final String jurisdiction
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = "A displayable case")
     })
+    @LogAudit(operationType = OperationType.SEARCH_CASE)
```
mario-paniccia, line 196: this is not a search, it's  a  case view. VIEW_CASE?

src/main/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpoint.java
```diff
@@ -236,13 +212,14 @@ public CaseView findCase(@PathVariable("jid") final String jurisdictionId,
         @ApiResponse(code = 200, message = "Empty pre-state conditions"),
         @ApiResponse(code = 422, message = "The case status did not qualify for the event")
     })
+    @LogAudit(operationType = OperationType.CREATE_CASE)
     public CaseEventTrigger getEventTriggerForCaseType(@PathVariable("uid") String userId,
```
mario-paniccia, line 216: better to ignore

src/main/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpoint.java
```diff
@@ -290,6 +269,7 @@ public CaseEventTrigger getEventTriggerForDraft(@PathVariable("uid") String user
         @ApiResponse(code = 200, message = "Displayable case data"),
         @ApiResponse(code = 404, message = "Invalid jurisdiction/case type/case reference or event id")
     })
+    @LogAudit(operationType = OperationType.VIEW_CASE)
```
mario-paniccia, line 272: better to introduce a VIEW_CASE_HISTORY operation

src/main/java/uk/gov/hmcts/ccd/v2/external/controller/DocumentController.java
```diff
@@ -66,6 +68,7 @@ public DocumentController(
             message = V2.Error.PRINTABLE_DOCUMENTS_ENDPOINT_DOWN
         )
     })
+    @LogAudit(operationType = OperationType.VIEW_CASE, caseId = "#caseId")
     public ResponseEntity<DocumentsResource> getDocuments(@PathVariable("caseId") String caseId) {
```
mario-paniccia, line 72: either we ignore, or we create a better operation name I'd suggest

src/main/java/uk/gov/hmcts/ccd/v2/external/controller/StartTriggerController.java
```diff
@@ -61,6 +63,8 @@ public StartTriggerController(
             message = EVENT_TRIGGER_NOT_FOUND
         )
     })
+    @LogAudit(operationType = OperationType.CREATE_CASE, jurisdiction = "#result.body.caseDetails.jurisdiction",
+        caseType = "#caseTypeId", eventName = "#triggerId")
     public ResponseEntity<StartTriggerResource> getStartCaseTrigger(@PathVariable("caseTypeId") String caseTypeId,
```
mario-paniccia, line 68: better to ignore

src/main/java/uk/gov/hmcts/ccd/v2/external/controller/StartTriggerController.java
```diff
@@ -104,6 +108,8 @@ public StartTriggerController(
             message = EVENT_TRIGGER_NOT_FOUND
         )
     })
+    @LogAudit(operationType = OperationType.UPDATE_CASE, jurisdiction = "#result.body.caseDetails.jurisdiction",
+        caseType = "#result.body.caseDetails.caseTypeId", eventName = "#triggerId", caseId = "#caseId")
     public ResponseEntity<StartTriggerResource> getStartEventTrigger(@PathVariable("caseId") String caseId,
```
mario-paniccia, line 113: we can ignore

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseController.java
```diff
@@ -108,6 +113,8 @@ public UICaseController(
             message = "Case event not found"
         )
     })
+    @LogAudit(operationType = VIEW_CASE, caseId = "#caseId", eventName = "#result.body.event.eventId",
```
mario-paniccia, line 116: I'd suggest VIEW_CASE_HISTORY

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
smathangi: That's a placeholder method if we want json format . But requirements want planing text.. so will delete the method.

src/main/java/uk/gov/hmcts/ccd/endpoint/ui/QueryEndpoint.java
rafalkalita: We will ignore this one.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditInterceptor.java
smathangi: yes, probably as part of V1 PR

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
smathangi: Eg: 400 bad request when Spring fails to bind data to controller methods. Also, 403/401

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditRepository.java
smathangi: if you introduce formatter class then it should inject here rather than AuditServcie.  I like passing domain entity to repository

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/AuditAspect.java
smathangi: sure, but that's the cleanest option we came across. Any other design suggestions always welcome.

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/AuditContextHolder.java
smathangi: it depends on how you use it. how do you pass to Interceptor then..?

src/main/java/uk/gov/hmcts/ccd/endpoint/std/DraftsEndpoint.java
rafalkalita: We will ignore drafts totally. We use UPDATE_CASE when we already have the case created (it has a caseId/reference)

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
smathangi: removed.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
smathangi: need to reflect in the jira/confluence first.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
smathangi: you get that from LogAudit() annotation on the controller method.  I feel defining another class hierarchy complicates aspect design.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditInterceptor.java
smathangi: agree.. don't like passing http semantics

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
smathangi: Can't work without that - it's not optional. 
SecurityUtils referring AuthTokenGenerator which is loading lazily at the moment.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
smathangi: moved this to interceptor and extracted to a constant now.

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/AuditAspect.java
smathangi: Controller advice logs the same. Actually no need to catch and throw here.

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/ExpressionEvaluator.java
smathangi: That class similar to VO/DTO.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditRepository.java
smathangi: at any point of time we've one implementation only right? Also this is not a library class. https://martinfowler.com/bliki/InterfaceImplementationPair.html

src/main/java/uk/gov/hmcts/ccd/auditlog/OperationType.java
```diff
@@ -0,0 +1,19 @@
...
+    CREATE_CASE("Create case"),
+    UPDATE_CASE("Update case"),
+    VIEW_CASE("View case"),
+    SEARCH_CASE("Search case"),
+    UPDATE_CASE_ACCESS("Update case access");
+
+    private final String label;
+
+    private OperationType(String label) {
```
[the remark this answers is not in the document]
smathangi: removed

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditInterceptor.java
smathangi: done. removed

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditEntry.java
smathangi: done

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseController.java
rafalkalita: done

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
Thor-tech-of-metal: Fair enough . It means that SecurityUtils is liked to Lazy initialization due to it is using AuthTokenGenerator with lazy loading.  Perfect : 👌

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/ExpressionEvaluator.java
Thor-tech-of-metal: It is ok . there has to be a getter.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class AuditInterceptor extends HandlerInterceptorAdapter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AuditAspect.class);
+
+    public static final String REQUEST_ID = "request-id";
+
+    private final AuditService auditService;
+
+    private static final List<HttpStatus> IGNORED_STATUSES = Lists.newArrayList(NOT_FOUND);
```
mario-paniccia, line 28: this should be a config property rather than hardcoded?

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditOperationType.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.hmcts.ccd.auditlog;
+
+public enum AuditOperationType {
+    CREATE_CASE("Create case"),
+    UPDATE_CASE("Update case"),
+    CASE_ACCESSED("Case Accessed"),
+    SEARCH_CASE("Search case"),
+    UPDATE_CASE_ACCESS("Update case access"),
```
mario-paniccia, line 8: mmm did we say to call this "Update case access permissions"?

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditRepository.java
mario-paniccia: The benefit of introducing an interface is not so that we can support more than one destination at a single time. That is not what I meant. The benefit is that it allows us to change the destination limiting the amount of changes we need to do. We just inject a different implementation class without having to change the client classes using the AuditRepository. Because the client classes use an interface. If you look at any implementation of the repository design pattern on internet it's always defined as an interface, also when there is only 1 implementation class:
https://thoughts-on-java.org/implementing-the-repository-pattern-with-jpa-and-hibernate/
https://medium.com/@krzychukosobudzki/repository-design-pattern-bc490b256006

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
```diff
@@ -0,0 +1,57 @@
...
+public class AuditService {
+
+    private final Clock clock;
+    private final SecurityUtils securityUtils;
+    private final UserRepository userRepository;
+    private final AuditRepository auditRepository;
+
+    public AuditService(@Qualifier("utcClock") final Clock clock,
+                        @Qualifier(CachedUserRepository.QUALIFIER) final UserRepository userRepository,
+                        @Lazy final SecurityUtils securityUtils, final AuditRepository auditRepository) {
```
mario-paniccia, line 26: we should use dependency injection for AuditRepository

src/main/java/uk/gov/hmcts/ccd/auditlog/aop/AuditAspect.java
mario-paniccia: I think AOP is a good choice. We just need to evaluate the performance of the solution.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditRepository.java
smathangi: haha.. we can't generalise every single principle to each scenario. This particular one has just one line of code (log statement), surely there won't be any caching required, dependency injection still possible.  I've seen many well known libraries including Spring has single implementation class without interface.  

Personally I disagree in this particular scenario but pushed as requested.

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditService.java
smathangi: it does

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditInterceptor.java
smathangi: done

src/main/java/uk/gov/hmcts/ccd/auditlog/AuditOperationType.java
rafalkalita: done

PR #956, opened 2020-05-27
RDM-8325 - Internal /searchCases API

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-8325

### Change description ###
Implement internal version of the ES based search API

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/definition/CachedUIDefinitionGateway.java
```diff
@@ -43,6 +44,13 @@ public SearchResult getSearchResult(final int version, final String caseTypeId)
         return httpUiDefinitionGateway.getSearchResult(version, caseTypeId);
     }
 
+    @Override
+    @Cacheable("searchCasesResultCache")
+    public SearchResult getSearchCasesResult(final int version, final String caseTypeId, final UseCase useCase) {
```
mario-paniccia, line 49: please let's suffix definition domain classes and methods with Definition

mario-paniccia, line 49: can you remove `final` please? makes the code heavy to read

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedUIDefinitionGateway.java
```diff
@@ -43,6 +44,13 @@ public SearchResult getSearchResult(final int version, final String caseTypeId)
...
+        LOG.debug("remote retrieving version {} of search cases result for {}", version, caseTypeId);
```
mario-paniccia, line 50:
```suggestion
        LOG.debug("remote retrieving version {} of search cases result definition for {}", version, caseTypeId);
```

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedUIDefinitionGateway.java
mario-paniccia, line 49:
```suggestion
    public SearchResult getSearchCasesResultDefinition(final int version, final String caseTypeId, final UseCase useCase) {
```

mario-paniccia, line 49: temporary: do we need caseTypeId?

src/main/java/uk/gov/hmcts/ccd/data/definition/HttpUIDefinitionGateway.java
```diff
@@ -182,6 +183,30 @@ public SearchResult getWorkBasketResult(int version, String caseTypeId) {
         }
     }
 
+    @Override
+    public SearchResult getSearchCasesResult(int version, String caseTypeId, UseCase useCase) {
+        try {
+            final Instant start = Instant.now();
```
mario-paniccia, line 189: have you got `final` configured in your IDE? I think `final` added everywhere really makes the code heavy to read. I think it's not worth it

src/main/java/uk/gov/hmcts/ccd/data/user/UserService.java
```diff
@@ -58,6 +62,13 @@ public UserProfile getUserProfile() {
         return createUserProfile(idamProperties, userId, jurisdictionDefinitions);
     }
 
+    public List<CaseTypeDefinition> getUserCaseTypes() {
+        return Arrays.stream(getUserProfile().getJurisdictions())
+            .map(JurisdictionDisplayProperties::getCaseTypeDefinitions)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toList());
+    }
```
mario-paniccia, line 70: this is an interesting solution in general. But for ManagaCaseAssignment won't work I think. I've just been recently told that the only IDAM role the client has is `caseworker-caa` and she needs to see all the existing case types. So in this case this will return an empty list

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseTypeDefinition.java
```diff
@@ -168,4 +174,27 @@ public boolean isCaseFieldACollection(String caseFieldId) {
...
+        }
+
+        return Optional.empty();
+    }
+
+    @JsonIgnore
+    public Map<String, TextNode> getLabelsFromCaseFields() {
+        return getCaseFieldDefinitions()
+            .stream()
+            .filter(caseField -> LABEL.equals(caseField.getFieldTypeDefinition().getType()))
```
mario-paniccia, line 197: temporary: why we getting only the field of type label here?

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseTypeDefinition.java
```diff
@@ -168,4 +174,27 @@ public boolean isCaseFieldACollection(String caseFieldId) {
...
+            .collect(Collectors.toMap(CaseFieldDefinition::getId, caseField -> instance.textNode(caseField.getLabel())));
```
mario-paniccia, line 198:
```suggestion
            .collect(Collectors.toMap(CaseFieldDefinition::getId, caseField -> jsonNodeFactory.textNode(caseField.getLabel())));
```

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseTypeDefinition.java
```diff
@@ -168,4 +174,27 @@ public boolean isCaseFieldACollection(String caseFieldId) {
     public Optional<CaseFieldDefinition> getCaseField(String caseFieldId) {
         return caseFieldDefinitions.stream().filter(caseField -> caseField.getId().equalsIgnoreCase(caseFieldId)).findFirst();
     }
+
+    @JsonIgnore
+    public Optional<CommonField> getCommonFieldByPath(String path) {
```
mario-paniccia, line 179:
```suggestion
    public Optional<CommonField> getFieldDefinitionByPath(String path) {
```

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseTypeDefinition.java
```diff
@@ -168,4 +174,27 @@ public boolean isCaseFieldACollection(String caseFieldId) {
...
+            return Optional.empty();
+        }
+        List<String> pathElements = Arrays.stream(path.trim().split("\\.")).collect(toList());
+
+        Optional<CaseFieldDefinition> topLevelCaseField = getCaseField(pathElements.get(0));
+        if (topLevelCaseField.isPresent()) {
+            return topLevelCaseField.get().getComplexFieldNestedField(pathElements.stream().skip(1).collect(Collectors.joining(",")));
+        }
+
+        return Optional.empty();
```
mario-paniccia, line 190:
```suggestion
  return topLevelCaseField.flatMap(f -> 
            f.getComplexFieldNestedField(pathElements.stream().skip(1).collect(Collectors.joining(",")))
        );
```

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseTypeDefinition.java
```diff
@@ -168,4 +174,27 @@ public boolean isCaseFieldACollection(String caseFieldId) {
     public Optional<CaseFieldDefinition> getCaseField(String caseFieldId) {
         return caseFieldDefinitions.stream().filter(caseField -> caseField.getId().equalsIgnoreCase(caseFieldId)).findFirst();
     }
+
+    @JsonIgnore
+    public Optional<CommonField> getCommonFieldByPath(String path) {
+        if (StringUtils.isBlank(path)) {
+            return Optional.empty();
+        }
+        List<String> pathElements = Arrays.stream(path.trim().split("\\.")).collect(toList());
```
mario-paniccia, line 183: lots of code emerging more and more to deal with path manipulation. 2 examples in this class, here another: 
https://github.com/hmcts/ccd-data-store-api/blob/0fc75ab4364e95f78271f62bb04322f94b406a4f/src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultField.java#L61

Can we gather all path related code into a new class `CaseFieldPathUtils` class please

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultField.java
```diff
@@ -108,4 +119,32 @@ public String getDisplayContextParameter() {
...
+        JsonNode caseField = Optional.ofNullable(caseFields.get(firstPathElement))
+            .orElseThrow(() -> new BadRequestException(format(NESTED_ELEMENT_NOT_FOUND_FOR_PATH, path)));
+
+        if (pathElements.size() == 1) {
+            return caseField;
+        } else {
+            List<String> tail = pathElements.subList(1, pathElements.size());
+            return reduce(caseField, tail, path);
+        }
+    }
```
mario-paniccia, line 149: temporary: not sure this is the best place for this code. Maybe better moving this to a new method in `CaseFieldPathUtils`?

src/main/java/uk/gov/hmcts/ccd/domain/model/search/UseCase.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.common.base.Strings;
+
+import java.util.Arrays;
+
+public enum UseCase {
+
+    WORKBASKET("WORKBASKET"),
+    SEARCH("SEARCH"),
+    ORG_CASES("ORGCASES"),
+    DEFAULT("");
```
mario-paniccia, line 14: temporary: does this mean every time a service introduces a new use case we need to do a new release of Data Store?

src/main/java/uk/gov/hmcts/ccd/domain/model/search/elasticsearch/UICaseSearchHeader.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.ccd.domain.model.search.elasticsearch;
+
+import lombok.Data;
+import lombok.NonNull;
+
+import java.util.List;
+
+@Data
+public class UICaseSearchHeader {
```
mario-paniccia, line 12:
```suggestion
public class SearchResultViewHeaderGroup {
```

src/main/java/uk/gov/hmcts/ccd/domain/model/search/elasticsearch/UICaseSearchHeader.java
```diff
@@ -0,0 +1,17 @@
...
+
+    @NonNull
+    private UICaseSearchHeaderMetadata metadata;
```
mario-paniccia, line 16:
```suggestion
    private HeaderGroupMetadata metadata;
```

src/main/java/uk/gov/hmcts/ccd/domain/model/search/elasticsearch/UICaseSearchResult.java
```diff
@@ -0,0 +1,39 @@
...
+import lombok.NonNull;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.emptyList;
+
+@Data
+public class UICaseSearchResult {
```
mario-paniccia, line 13: we are trying to follow the convention that all internal api result objects are denoted by `View`

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchCasesOperation.java
```diff
@@ -0,0 +1,170 @@
...
+import uk.gov.hmcts.ccd.endpoint.exceptions.BadRequestException;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static uk.gov.hmcts.ccd.domain.service.common.AccessControlService.CAN_READ;
+
+@Service
+public class MergeDataToSearchCasesOperation {
```
mario-paniccia, line 24: temporary: I would not model this as an operation with an execute method. Because by convention in Data Store Operation classes correspond to business operations like create case, view case. Even thought I can see we are introducing them for other other things recently

src/main/java/uk/gov/hmcts/ccd/domain/model/search/elasticsearch/UICaseSearchResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public static final UICaseSearchResult EMPTY = new UICaseSearchResult(emptyList(), emptyList(), 0L);
+
+    @NonNull
+    private List<UICaseSearchHeader> headers;
+    @NonNull
+    private List<SearchResultViewItem> cases;
+    @NonNull
+    private Long total;
+
```
mario-paniccia, line 23: was thinking it might provide useful to also add the use case here

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,101 @@
...
+        @Qualifier(AuthorisedCaseSearchOperation.QUALIFIER) CaseSearchOperation caseSearchOperation,
+        ElasticsearchQueryHelper elasticsearchQueryHelper) {
+        this.caseSearchOperation = caseSearchOperation;
+        this.elasticsearchQueryHelper = elasticsearchQueryHelper;
+    }
+
+    @PostMapping(
+        path = "",
+        headers = {
+            V2.EXPERIMENTAL_HEADER
```
mario-paniccia, line 48: not needed. Let's removed it please

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,101 @@
...
+        },
+        produces = {
+            V2.MediaType.UI_CASE_SEARCH
```
mario-paniccia, line 51: I know other v2 apis have versioned custom vendor types, but I'd like to follow HMCTS guidelines from now on which says API versioning should be avoided. We can then use a generic media type `application/json`. See:
https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1392411337#RDM-8325:InternalSearchAPI-3.1-Vendor-SpecificMediaTypes

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static uk.gov.hmcts.ccd.auditlog.aop.AuditContext.CASE_ID_SEPARATOR;
+import static uk.gov.hmcts.ccd.auditlog.aop.AuditContext.MAX_CASE_IDS_LIST;
+
+@RestController
+@RequestMapping(path = "/internal/searchCases")
+@Slf4j
+public class UICaseSearchController {
```
mario-paniccia, line 32: minor: 
we could call it `UISearchController` so that we can add other types of search related endpoints in the future, for example if we introduced searching events.

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,101 @@
...
+        headers = {
+            V2.EXPERIMENTAL_HEADER
+        },
+        produces = {
+            V2.MediaType.UI_CASE_SEARCH
+        }
+    )
+    @ApiOperation(
+        value = "Elastic search for cases returning paginated data",
+        notes = V2.EXPERIMENTAL_WARNING
```
mario-paniccia, line 56: `notes = V2.EXPERIMENTAL_WARNING` 
can you remove please

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,101 @@
...
+        @ApiResponse(
+            code = 400,
+            message = ERROR_CASE_ID_INVALID
+        ),
+        @ApiResponse(
+            code = 404,
+            message = "Case not found"
+        )
+    })
+    // TODO: Docs
```
mario-paniccia: do we need to document other error codes?

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,124 @@
...
+            message = "An unexpected situation that is not attributable to the user or API Client; or request is invalid. "
+                      + "For some other types HTTP code 400 is returned instead.\n"
+                      + "Invalid request examples include:\n"
+                      + "- Malformed JSON request."
+        )
+    })
+    @LogAudit(operationType = AuditOperationType.SEARCH_CASE, caseTypeIds = "#caseTypeIds",
+        caseId = "T(uk.gov.hmcts.ccd.v2.internal.controller.UICaseSearchController).buildCaseIds(#result)")
+    public ResponseEntity<CaseSearchResultViewResource> searchCases(
+                                     @ApiParam(value = "Comma-separated list of case type ID(s).")
```
mario-paniccia, line 99:
```suggestion
                                     @ApiParam(value = "Comma-separated list of case type ID(s). If omitted, the search is across all existing case types. For better performance, prefer specifying the case types when possible")
```

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,124 @@
...
+                                     @RequestParam(value = "ctid", required = false) List<String> caseTypeIds,
+                                     @ApiParam(value = "Use case for search. Examples include `WORKBASKET`, `SEARCH` or `ORGCASES`.")
```
mario-paniccia, line 101:
```suggestion
                                     @ApiParam(value = "Use case for search. Examples include `WORKBASKET`, `SEARCH` or `ORGCASES`. Used when the list of fields to return is configured in the CCD definition. If omitted, all the case fields are returned")
```

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,124 @@
...
+                                     @RequestParam(value = "usecase", required = false) final String useCase,
+                                     @ApiParam(value = "Native ElasticSearch Search API request. Please refer to the ElasticSearch official "
+                                         + "documentation.",
```
mario-paniccia, line 104: can you also point them to the elastic search API LLD please: https://tools.hmcts.net/confluence/display/RCCD/CCD+ElasticSearch+and+new+search+API+Design

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public CrossCaseTypeSearchRequest prepareRequest(List<String> caseTypeIds, String useCaseString, String jsonSearchRequest) {
+        UseCase useCase;
+        try {
+            useCase = UseCase.valueOfReference(useCaseString);
+        } catch (IllegalArgumentException ex) {
+            throw new BadSearchRequest(String.format("The provided use case '%s' is unsupported.", useCaseString));
+        }
+
+        rejectBlackListedQuery(jsonSearchRequest);
```
mario-paniccia, line 77: can't we delegate this to the external search?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,166 @@
...
+            })
+            .findFirst();
+        blackListedQueryOpt.ifPresent(blacklisted -> {
+            throw new BadSearchRequest(String.format("Query of type '%s' is not allowed", blacklisted));
+        });
+    }
+
+    private List<String> buildCaseTypeIds(List<String> caseTypeIds) {
+        return CollectionUtils.isEmpty(caseTypeIds)
+            ? userService.getUserCaseTypes().stream().map(CaseTypeDefinition::getId).collect(Collectors.toList())
```
mario-paniccia, line 163: temporary: now that we are making ctid optional in the external search API to allow searching on all existing case types, is there any benefit in doing this?

mario-paniccia: furthermore, this can negatively impact search performance. If you consider a just redeployed DataStore, where all caches are empty. For each user jurisdiction we need to load its definition from def store before we can make the search. This might be too inefficient. It heavily relies on caching of the definitions.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,166 @@
...
+        } catch (IllegalArgumentException ex) {
+            throw new BadSearchRequest(String.format("The provided use case '%s' is unsupported.", useCaseString));
+        }
+
+        rejectBlackListedQuery(jsonSearchRequest);
+
+        final List<String> updatedCaseTypeIds = buildCaseTypeIds(caseTypeIds);
+
+        JsonNode searchRequest = stringToJsonNode(jsonSearchRequest);
+        if (useCase != UseCase.DEFAULT) {
```
mario-paniccia, line 82: I wouldn't introduce a concept of Default use case. It's a bit misleading we don't really have a default use case. If use case is not passed then is missing. 
Can't we just check whether the string is empty or null?

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
```diff
@@ -67,45 +59,25 @@ public CaseSearchResult searchCases(
...
-                  required = true)
+            required = true)
         @RequestBody String jsonSearchRequest) {
 
         Instant start = Instant.now();
 
-        rejectBlackListedQuery(jsonSearchRequest);
+        CrossCaseTypeSearchRequest request = elasticsearchQueryHelper.prepareRequest(
+            caseTypeIds,
+            UseCase.DEFAULT.getReference(),
```
mario-paniccia, line 69: use case is a concept of the internal API only. It should not appear here.

mario-paniccia: also, having a common helper to prepare both internal and external request will make our life more complicated in the future because it's likely the requests will have different requirements so you'll have code that doesn't make sense for one of the 2

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
```diff
@@ -67,45 +59,25 @@ public CaseSearchResult searchCases(
...
             + "search results, please state the alias fields to be returned in the _source property for e.g."
             + " \"_source\":[\"alias.customer\",\"alias.postcode\"]",
-                  required = true)
+            required = true)
         @RequestBody String jsonSearchRequest) {
 
         Instant start = Instant.now();
 
-        rejectBlackListedQuery(jsonSearchRequest);
+        CrossCaseTypeSearchRequest request = elasticsearchQueryHelper.prepareRequest(
```
mario-paniccia, line 67: potential regression risk here ⚠️ . We have changed the external API. We introduced:
`buildCaseTypeIds(caseTypeIds)`
on the external API which definitely would require regression testing.

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
```diff
@@ -1,55 +1,47 @@
...
 @RestController
 @RequestMapping(path = "/",
-                consumes = MediaType.APPLICATION_JSON_VALUE,
-                produces = MediaType.APPLICATION_JSON_VALUE)
+    consumes = MediaType.APPLICATION_JSON_VALUE,
+    produces = MediaType.APPLICATION_JSON_VALUE)
 @Api(tags = {"Elastic Based Search API"})
 @SwaggerDefinition(tags = {
-    @Tag(name = "Elastic Based Search API", description = "New ElasticSearch based search API")
+    @Tag(name = "Elastic Based Search API", description = "External ElasticSearch based search API")
```
mario-paniccia, line 32:
```suggestion
    @Tag(name = "Elastic Based Search API", description = "ElasticSearch based search API")
```

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
```diff
@@ -67,45 +59,25 @@ public CaseSearchResult searchCases(
...
+            jsonSearchRequest
+        );
 
-        CrossCaseTypeSearchRequest request = new CrossCaseTypeSearchRequest.Builder()
-            .withCaseTypes(caseTypeIds)
-            .withSearchRequest(stringToJsonNode(jsonSearchRequest))
-            .build();
-
-        CaseSearchResult result = caseSearchOperation.execute(request);
+        CaseSearchResult result = caseSearchOperation.executeExternal(request);
```
mario-paniccia, line 73: in the code base methods we don't really mention external vs internal. I'd keep it agnostic as simply `execute`.

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,126 @@
...
+                                     example = "{\"_source\":[\"data.TextField\"],\"query\":{\"match_all\":{}},\"size\":20,\"from\":1}")
+                                     @RequestBody String jsonSearchRequest) {
+        Instant start = Instant.now();
+
+        String useCaseTransformed = Strings.isNullOrEmpty(useCase) ? useCase : useCase.toUpperCase();
```
mario-paniccia, line 110:
```suggestion
        String useCaseUppercase = Strings.isNullOrEmpty(useCase) ? useCase : useCase.toUpperCase();
```

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchCaseSearchOperation.java
```diff
@@ -43,22 +45,25 @@
...
         this.objectMapper = objectMapper;
         this.caseDetailsMapper = caseDetailsMapper;
         this.applicationParams = applicationParams;
         this.caseSearchRequestSecurity = caseSearchRequestSecurity;
+        this.mergeDataToSearchCasesOperation = mergeDataToSearchCasesOperation;
     }
 
     @Override
-    public CaseSearchResult execute(CrossCaseTypeSearchRequest request) {
+    public CaseSearchResult executeExternal(CrossCaseTypeSearchRequest request) {
```
mario-paniccia, line 66: I would keep the code method and classes agnostic to the external vs internal classification

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,159 @@
...
+        this.objectMapper = objectMapper;
+        this.applicationParams = applicationParams;
+        this.objectMapperService = objectMapperService;
+        this.searchQueryOperation = searchQueryOperation;
+        this.getCaseTypeOperation = getCaseTypeOperation;
+        this.userService = userService;
+        this.elasticsearchMappings = elasticsearchMappings;
+    }
+
+    public CrossCaseTypeSearchRequest prepareRequest(List<String> caseTypeIds, String useCase, String jsonSearchRequest) {
```
mario-paniccia, line 69: some elements are not suitable for the external API. for example the concept of use case, and the adding of the default sorting. If we want to keep this common class shared between the external and internal api, then let's keep only the things strictly in common

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private ArrayNode buildSortNode(List<String> caseTypeIds, String useCase) {
+        ArrayNode sortNode = objectMapper.createArrayNode();
+        caseTypeIds.forEach(caseTypeId -> addCaseTypeSorts(caseTypeId, useCase, sortNode));
+        return sortNode;
+    }
+
+    private void addCaseTypeSorts(String caseTypeId, String useCase, ArrayNode sortNode) {
+        Optional<CaseTypeDefinition> caseTypeOpt = getCaseTypeOperation.execute(caseTypeId, CAN_READ);
```
mario-paniccia, line 102: temporary: wonder whether calling the getCaseTypeOperation is too "heavy". Perhaps would be ok to just call: 
    `CaseTypeDefinition getCaseType(String caseTypeId);` ?

src/main/resources/application.properties
```diff
@@ -149,6 +149,19 @@ spring.elasticsearch.jest.read-timeout=10000ms
 management.health.elasticsearch.enabled=false
 management.health.status.order=DOWN, OUT_OF_SERVICE, UNKNOWN, UP, OUT_OF_SYNC, PROBLEM, COULD_NOT_CONNECT
 
+# Elasticsearch mappings
+elasticsearch.type-mappings.defaultText=Text, TextArea, FixedList, FixedListEdit, MultiSelectList, FixedRadioList, DynamicList
+elasticsearch.type-mappings.defaultDouble=Number, MoneyGBP
+elasticsearch.type-mappings.defaultDate=Date, Time, DateTime
+elasticsearch.type-mappings.ccdPhoneUK=PhoneUK
+elasticsearch.type-mappings.defaultKeyword=YesOrNo, Email, Postcode
+elasticsearch.type-mappings.ccdDocument=Document
```
mario-paniccia, line 158: why are these needed? I thought we only needed to know whether a CCD type is a defaultText

src/main/resources/application.properties
```diff
@@ -149,6 +149,19 @@ spring.elasticsearch.jest.read-timeout=10000ms
...
+elasticsearch.case-predefined-mappings.defaultText=reference, jurisdiction, state, case_type_id
+elasticsearch.case-predefined-mappings.defaultLong=id
+elasticsearch.case-predefined-mappings.defaultDate=created_date, last_modified
+elasticsearch.case-predefined-mappings.defaultKeyword=security_classification
+elasticsearch.case-predefined-mappings.disabled=@timestamp, @version, index_id
```
mario-paniccia: why are these needed? I thought we only needed to know whether a CCD metadata is a defaultText

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseTypeDefinition.java
danlysiak: Just to not break existing behaviour with generating the case response data (this was originally in `MergeDataToSearchResultOperation`). Unsure of the original reasoning behind it, however..

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
```diff
@@ -1,55 +1,46 @@
...
 @RestController
 @RequestMapping(path = "/",
-                consumes = MediaType.APPLICATION_JSON_VALUE,
-                produces = MediaType.APPLICATION_JSON_VALUE)
+    consumes = MediaType.APPLICATION_JSON_VALUE,
+    produces = MediaType.APPLICATION_JSON_VALUE)
 @Api(tags = {"Elastic Based Search API"})
 @SwaggerDefinition(tags = {
-    @Tag(name = "Elastic Based Search API", description = "New ElasticSearch based search API")
+    @Tag(name = "Elastic Based Search API", description = "ElasticSearch based search API")
```
mario-paniccia, line 31:
```suggestion
    @Tag(name = "Elastic Based Search API", description = "ElasticSearch based cases search API")
```

mario-paniccia: Or something along those lines :)

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,139 @@
...
+                      + "\n\t\"size\": 20,\n\t\"from\": 1\n}",
+            required = true
+        )
+    )
+    @LogAudit(operationType = AuditOperationType.SEARCH_CASE, caseTypeIds = "#caseTypeIds",
+        caseId = "T(uk.gov.hmcts.ccd.v2.internal.controller.UICaseSearchController).buildCaseIds(#result)")
+    public ResponseEntity<CaseSearchResultViewResource> searchCases(
+                                     @ApiParam(value = "Case type ID for search.", required = true)
+                                     @RequestParam(value = "ctid") String caseTypeId,
+                                     @ApiParam(value = "Use case for search. Examples include `WORKBASKET`, `SEARCH` or `ORGCASES`. "
```
mario-paniccia, line 116:
```suggestion
                                     @ApiParam(value = "Use case for search. Examples include `WORKBASKET`, `SEARCH` or `orgcases`. "
                                     
```

mario-paniccia: So that we don't give the impression that it has to be uppercase

src/main/java/uk/gov/hmcts/ccd/v2/internal/controller/UICaseSearchController.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.stream.Collectors;
+
+import static uk.gov.hmcts.ccd.auditlog.aop.AuditContext.CASE_ID_SEPARATOR;
+import static uk.gov.hmcts.ccd.auditlog.aop.AuditContext.MAX_CASE_IDS_LIST;
+
+@RestController
+@RequestMapping(path = "/internal/searchCases")
+@Api(tags = {"Elastic Based Search API"})
+@SwaggerDefinition(tags = {
+    @Tag(name = "Elastic Based Search API", description = "Internal ElasticSearch based search API")
```
mario-paniccia, line 30:
```suggestion
    @Tag(name = "Elastic Based Search API", description = "Internal ElasticSearch based cases search API. It returns extra information on case fields required by the UI for display purposes ")
```

mario-paniccia: or something similar :)

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
```diff
@@ -67,17 +58,12 @@ public CaseSearchResult searchCases(
...
 
         Instant start = Instant.now();
 
-        rejectBlackListedQuery(jsonSearchRequest);
-
-        CrossCaseTypeSearchRequest request = new CrossCaseTypeSearchRequest.Builder()
-            .withCaseTypes(caseTypeIds)
-            .withSearchRequest(stringToJsonNode(jsonSearchRequest))
-            .build();
+        CrossCaseTypeSearchRequest request = elasticsearchQueryHelper.prepareRequest(caseTypeIds, jsonSearchRequest);
```
mario-paniccia, line 66: minor: I have the feeling having this class in common between the 2 endpoints internal and eternal will complicate our lives. We need the 2 endpoints controller methods to be able to evolve independently. They might have different requirements. For example we might decide to allow not passing any case type for the external searches in the future. But only for the external not the internal. If that happens, since the code is in common, you risk regressions on the internal version as well if we don't notice the code is in common.
If the goal is to remove duplication, there are better ways to do it. You can have the common utility methods in `elasticsearchQueryHelper`. But call only the ones needed from the 2 controllers

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
```diff
@@ -87,25 +73,6 @@ public CaseSearchResult searchCases(
...
-                Matcher m = p.matcher(jsonSearchRequest);
-                return m.find();
-            })
-            .findFirst();
-        blackListedQueryOpt.ifPresent(blacklisted -> {
-            throw new BadSearchRequest(String.format("Query of type '%s' is not allowed", blacklisted));
-        });
-    }
-
     public static String buildCaseIds(CaseSearchResult caseSearchResult) {
```
mario-paniccia, line 76: my IDE flag this as unused. Can we remove?

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
mario-paniccia, line 66: I like the idea of having common code in ElasticsearchQueryHelper but because we are moving everything in there the controllers are almost empty and ElasticsearchQueryHelper is quite complex. We could improve things by moving the code in `prepareRequest` back to the controllers. And leave the utility methods common to both in ElasticsearchQueryHelper. I prefer to see in the controller method itself the steps it's taking.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,161 @@
...
+        if (CollectionUtils.isEmpty(caseTypeIds)) {
+            throw new BadSearchRequest("At least one case type ID is required.");
+        }
+
+        JsonNode searchRequest = validateAndConvertRequest(jsonSearchRequest);
+
+        return new CrossCaseTypeSearchRequest.Builder()
+            .withCaseTypes(caseTypeIds)
+            .withSearchRequest(searchRequest)
+            .build();
```
mario-paniccia, line 75: minor: if we moved this code inside the `CaseSearchOperation.execute` we could have the external search API be completely responsible to create the CrossCaseTypeSearchRequest. We just delegate to the external. What do you think? too much stuff to move around?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private void applyConfiguredSort(JsonNode searchRequest, String caseTypeId, String useCase) {
+        JsonNode sortNode = searchRequest.get(SORT);
+        if (sortNode == null || useCase == null) {
+            ArrayNode appliedSortsNode = buildSortNode(caseTypeId, useCase);
+            if (appliedSortsNode.size() > 0) {
+                ((ObjectNode)searchRequest).set(SORT, appliedSortsNode);
+            }
+        }
```
mario-paniccia, line 99: I would strongly suggest we introduce a separated class that deals only with manipulation of the JSON response. Nothing else. 
For example a ElasticSearchRequest class. With methods like:
new ElasticSearchRequest(jsonRequest)
esr.toJson()
esr.isSorted()
esr.addSorting(fields)
For now we can use it here only and then pass down to the external API the json request by calling .toJson(). So we don't have to change much code

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static uk.gov.hmcts.ccd.domain.service.common.AccessControlService.CAN_READ;
+
+@Service
+@Slf4j
+public class ElasticsearchQueryHelper {
```
mario-paniccia, line 36: this class is too big, because is doing too many unrelated things. Like applying sorting, validation, Json manipulation. We need to apply separation of concerns and move some stuff out to other helper classes. Some thing can be moved back to the controller.

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedUIDefinitionGateway.java
```diff
@@ -43,6 +43,13 @@ public SearchResult getSearchResult(final int version, final String caseTypeId)
         return httpUiDefinitionGateway.getSearchResult(version, caseTypeId);
     }
 
+    @Override
+    @Cacheable("searchCasesResultCache")
+    public SearchResult getSearchCasesResultDefinition(int version, String caseTypeId, String useCase) {
```
mario-paniccia, line 48: minor: probably we could have considered removing the use case so we cache the whole SearchCasesResultFields tab to decrease the network calls when requesting different use cases

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseFieldPathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+                                                                 boolean pathIncludesParent) {
+        if (StringUtils.isBlank(path) || fieldTypeDefinition.getChildren().isEmpty() || (pathIncludesParent && splitPath(path).length == 1)) {
+            return Optional.empty();
+        }
+        List<String> pathElements = getPathElements(path);
+
+        return reduce(fieldTypeDefinition.getChildren(), pathIncludesParent ? getPathElementsTail(pathElements) : pathElements);
+    }
+
+    public static JsonNode getCaseFieldNodeByPath(JsonNode node, String path) {
```
mario-paniccia, line 52: can we find a better name? maybe getNestedCaseFieldByPath?
can you also add a little javadoc with description here and on the other methods too please

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+import uk.gov.hmcts.ccd.endpoint.exceptions.BadSearchRequest;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static uk.gov.hmcts.ccd.domain.service.common.AccessControlService.CAN_READ;
+
+@Service
+public class CaseSearchResultGenerator {
```
mario-paniccia, line 27:
```suggestion
public class CaseSearchResultViewGenerator {
```

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public UICaseSearchResult execute(final String caseTypeId,
+                                      final CaseSearchResult caseSearchResult,
+                                      final String useCase) {
+        // TODO: Filter out fields from result that haven't been requested before returning (RDM-8556)
+        return new UICaseSearchResult(
+            buildHeaders(caseTypeId, useCase, caseSearchResult),
+            buildItems(useCase, caseSearchResult),
+            caseSearchResult.getTotal(),
+            useCase
```
mario-paniccia, line 49: minor: sorry Dan I changed my mind only if it's not a big change can we remove use case? I should have applied YAGNI

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+        );
+    }
+
+    private List<SearchResultViewItem> buildItems(String useCase, CaseSearchResult caseSearchResult) {
+        List<SearchResultViewItem> items = new ArrayList<>();
+        caseSearchResult.getCases().forEach(caseDetails -> {
+            getCaseTypeDefinition(caseDetails.getCaseTypeId()).ifPresent(caseType -> {
+                final SearchResult searchResult = searchQueryOperation.getSearchResultDefinition(caseType, useCase);
```
mario-paniccia, line 57: I find it very confusing that once there results are already returned by the searchQueryOperation then we are then calling it again during the
generation of the final response

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+        getCaseTypeDefinition(caseTypeId).ifPresent(caseType -> {
+            SearchResultViewHeaderGroup caseSearchHeader = buildHeader(useCase, caseSearchResult, caseTypeId, caseType);
+            headers.add(caseSearchHeader);
+        });
+
+        return headers;
+    }
+
+    private Optional<CaseTypeDefinition> getCaseTypeDefinition(String caseTypeId) {
+        return getCaseTypeOperation.execute(caseTypeId, CAN_READ);
```
mario-paniccia, line 76: I wouldn't call the whole operation. I think we can simplify things by just getting it from the DefRepo? The external API will take care of access control. Important thing that we test this.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+        return searchResultField.getDisplayContextParameter() == null
+            ? commonField.getDisplayContextParameter()
+            : searchResultField.getDisplayContextParameter();
+    }
+
+    private SearchResultViewItem buildSearchResultViewItem(final CaseDetails caseDetails,
+                                                           final CaseTypeDefinition caseTypeDefinition,
+                                                           final SearchResult searchResult) {
+
+        Map<String, JsonNode> caseData = new HashMap<>(caseDetails.getData());
```
mario-paniccia, line 148: how come we are building new hashmaps?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+        Map<String, Object> caseMetadata = new HashMap<>(caseDetails.getMetadata());
+        Map<String, TextNode> labels = caseTypeDefinition.getLabelsFromCaseFields();
```
mario-paniccia, line 150: not sure I understand why we are getting the fields of type label for here. Furthermore, shall we rename the method to 
`getLabelFieldTypesFromCaseFields()` or sim? unless I'm misunderstanding what this is doing

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+        Map<String, Object> caseFields = prepareData(searchResult, caseData, caseMetadata, labels);
+
+        String caseId = caseDetails.hasCaseReference() ? caseDetails.getReferenceAsString() : caseDetails.getId();
```
mario-paniccia, line 153: we can just call caseDetails.getId(). We will never be searching on draft cases that are the ones with  no id

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private Map<String, Object> prepareData(SearchResult searchResult,
+                                            Map<String, JsonNode> caseData,
+                                            Map<String, Object> metadata,
+                                            Map<String, TextNode> labels) {
+
+        Map<String, Object> newResults = new HashMap<>();
+
+        searchResult.getFieldsWithPaths().forEach(searchResultField -> {
+            JsonNode jsonNode = caseData.get(searchResultField.getCaseFieldId());
```
mario-paniccia, line 165:
```suggestion
            JsonNode topLevelCaseFieldNode = caseData.get(searchResultField.getCaseFieldId());
```

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+            if (jsonNode != null) {
+                newResults.put(searchResultField.getCaseFieldId() + "." + searchResultField.getCaseFieldPath(),
```
mario-paniccia, line 167: I can see this is duplicated in a couple of places. Can you add a method to the searchResultField class called getIdIncludingPath() or sim?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,161 @@
...
+    private void addCaseTypeSorts(String caseTypeId, String useCase, ArrayNode sortNode) {
+        Optional<CaseTypeDefinition> caseTypeOpt = getCaseTypeOperation.execute(caseTypeId, CAN_READ);
+        caseTypeOpt.ifPresent(caseType -> searchQueryOperation.getSortOrders(caseType, useCase)
+            .forEach(field -> sortNode.add(buildSortOrderFieldNode(caseType, field))));
+    }
+
+    private ObjectNode buildSortOrderFieldNode(CaseTypeDefinition caseTypeDefinition, SortOrderField sortOrderField) {
+        ObjectNode objectNode = objectMapper.createObjectNode();
+        CommonField commonField = caseTypeDefinition.getComplexSubfieldDefinitionByPath(sortOrderField.getCaseFieldId()).orElseThrow(() ->
+            new NullPointerException(String.format("Case field '%s' does not exist in configuration for case type '%s'.",
```
mario-paniccia, line 117: minor: NPE doesn't fit very well here? what about just a RuntimeException?

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
```diff
@@ -0,0 +1,161 @@
...
+        } else {
+            sb.append(CASE_DATA_PREFIX).append(sortOrderField.getCaseFieldId());
+            if (fieldType.getType().equals(FieldTypeDefinition.COLLECTION)) {
+                sb.append(COLLECTION_VALUE_SUFFIX);
+            }
+        }
+
+        if ((sortOrderField.isMetadata() && elasticsearchMappings.isDefaultTextMetadata(sb.toString()))
+             || (!sortOrderField.isMetadata() && elasticsearchMappings.isDefaultTextCaseData(fieldType))) {
+            sb.append(KEYWORD_SUFFIX);
```
mario-paniccia, line 135: it would be good adding some explanatory comment here since it's not possible to deduct from the code why we are doing this.

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
```diff
@@ -87,27 +90,57 @@ public SearchResultView execute(final String view,
         return mergeDataToSearchResultOperation.execute(caseType.get(), searchResult, draftsAndCases, draftResultError);
     }
 
-    private SearchResult getSearchResultDefinition(final CaseTypeDefinition caseTypeDefinition, final String view) {
-        if (WORKBASKET.equalsIgnoreCase(view)) {
-            return uiDefinitionRepository.getWorkBasketResult(caseTypeDefinition.getId());
+    public SearchResult getSearchResultDefinition(final CaseTypeDefinition caseTypeDefinition, final String useCase) {
```
mario-paniccia, line 93: I know it was like this already. But I don't think this method belongs here. The SearchQueryOperation should only concern with the actual execution of the external search. And be invoked only in that case. It's instead called before and after the actual search as well now which indicates it's doing too many unrelated things

mario-paniccia: minor: I would be great if we can rename SearchResult to SearchResultDefinition

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
```diff
@@ -87,27 +90,57 @@ public SearchResultView execute(final String view,
...
+        final String caseTypeId = caseTypeDefinition.getId();
+        if (Strings.isNullOrEmpty(useCase)) {
+            return buildSearchResultFromCaseFields(caseTypeDefinition);
```
mario-paniccia, line 96:
```suggestion
            return buildSearchResultDefinitionFromCaseFields(caseTypeDefinition);
```

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
```diff
@@ -87,27 +90,57 @@ public SearchResultView execute(final String view,
...
     }
 
-    private void addSortOrderFields(MetaData metadata,SearchResult searchResult) {
-        List<SortOrderField> sortOrders = getSortOrders(searchResult);
-        metadata.setSortOrderFields(sortOrders);
+    public List<SortOrderField> getSortOrders(CaseTypeDefinition caseType, String useCase) {
+        return getSortOrders(getSearchResultDefinition(caseType, useCase));
     }
 
     private List<SortOrderField> getSortOrders(SearchResult searchResult) {
```
mario-paniccia, line 113: perfect candidate method to be moved inside SearchResult itself

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
```diff
@@ -87,27 +90,57 @@ public SearchResultView execute(final String view,
...
-            .filter(searchResultField -> hasSortField(searchResultField))
-            .filter(searchResultField -> filterByRole(searchResultField))
+            .filter(this::hasSortField)
+            .filter(this::filterByRole)
             .sorted(Comparator.comparing(srf -> srf.getSortOrder().getPriority()))
             .map(this::toSortOrderField)
             .collect(Collectors.toList());
     }
 
+    private SearchResult buildSearchResultFromCaseFields(final CaseTypeDefinition caseTypeDefinition) {
```
mario-paniccia, line 122: minor: 
I've noticed that the "getSearchResultDefinition" is called multiple times, which will call this code multiple
times. Given that the performance of the search operation is important and this is potentially slow for big case types, we could consider refactoring to have it called only once if not a big change

src/main/java/uk/gov/hmcts/ccd/v2/internal/resource/CaseSearchResultViewResource.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @ApiModelProperty(value = "Headers for each case type")
+    private List<SearchResultViewHeaderGroup> headers;
+    @ApiModelProperty(value = "All cases across case types")
+    private List<SearchResultViewItem> cases;
+    @ApiModelProperty(value = "Total number of search results (including results not returned due to pagination)")
+    private Long total;
+    @ApiModelProperty(value = "The use case that the response represents; null for standard requests")
+    @JsonProperty("use_case")
+    private String useCase;
```
mario-paniccia, line 31: any particular reason for adding the use case here? If not I'd suggest removing it

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultField.java
```diff
@@ -108,4 +106,16 @@ public String getDisplayContextParameter() {
...
+    public String getUseCase() {
+        return useCase;
+    }
+
+    public void setUseCase(String useCase) {
+        this.useCase = useCase;
+    }
+
+    public JsonNode getCaseFieldNode(JsonNode topLevelCaseFieldNode) {
+        return CaseFieldPathUtils.getCaseFieldNodeByPath(topLevelCaseFieldNode, getCaseFieldPath());
```
mario-paniccia, line 119: can you inline this method? i.e. remove it and call directly the CaseFieldPathUtils in the current invoker of this method

src/main/java/uk/gov/hmcts/ccd/domain/model/search/elasticsearch/UICaseSearchResult.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.emptyList;
+
+@Data
+@AllArgsConstructor
+public class UICaseSearchResult {
```
mario-paniccia, line 15:
```suggestion
public class CaseSearchResultView {
```

src/main/java/uk/gov/hmcts/ccd/endpoint/std/CaseSearchEndpoint.java
danlysiak: This is still required (reference by string) for log and audit.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
danlysiak: Inherited behaviour from `MergeDataToSearchResultOperation`. It does look redundant here so removing 👍

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultGenerator.java
danlysiak: Inherited behaviour from `MergeDataToSearchResultOperation`. I'm a bit hesitant to remove in case of breaking anything the UI is depending on, so will leave unless we see at a later date that definitely not required.

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/SearchQueryOperation.java
danlysiak: Made some changes in the generator to reduce the number since we can assume it'll always be a single case type for all items currently 👍

src/main/java/uk/gov/hmcts/ccd/domain/service/search/elasticsearch/ElasticsearchQueryHelper.java
danlysiak: Did look into this potential but it did require more than expected so I have just put back to the controllers.

src/aat/resources/features/F-108/F-108_Test_Data_Base.td.json
```diff
@@ -0,0 +1,22 @@
+{
+  "_guid_": "F-108_Test_Data_Base",
+  "title": "External Elastic Search Endpoint",
+  "productName": "CCD Data Store Elastic Search API",
+  "operationName": "internal search query",
```
MSancaktutar, line 5: should be external 
@tomxelliott

src/aat/resources/features/F-108/F-108.feature
```diff
@@ -0,0 +1,269 @@
...
+  + "as case type
+    Given a case that has just been created as in [S-910_Create_Case_Private_Autotest1]
+    And a case that has just been created as in [S-910_Create_Case_Private_Autotest2]
+    And a wait time of 5 seconds [to allow for Logstash to index the case just created]
+    And a user with [private access to AUTOTEST1 jurisdiction only]
+    When the request [is configured to search for both the previously created cases]
+    And a request is prepared with appropriate values
+    And it is submitted to call the [internal search query] operation of [CCD Data Store Elastic Search API]
+    Then a positive response is received
+    And the response [contains only S-910_Create_Case_Private_Autotest1]
```
MSancaktutar, line 84: And the response [contains only the first case created above]
@tomxelliott

src/aat/java/uk/gov/hmcts/ccd/datastore/befta/DataStoreBeftaRunner.java
```diff
@@ -8,11 +8,10 @@
 import io.cucumber.junit.CucumberOptions;
 import uk.gov.hmcts.befta.BeftaMain;
 
-
 @RunWith(Cucumber.class)
 @CucumberOptions(plugin = "json:target/cucumber.json",
-    glue = "uk.gov.hmcts.befta.player",
-        features = { "classpath:features" }, tags = { "not @Ignore" })
+    glue = {"uk.gov.hmcts.befta.player", "uk.gov.hmcts.ccd.datastore.befta"},
```
MSancaktutar, line 13: a seperate step impl class won't be needed. @tomxelliott

PR #969, opened 2020-06-09
Moved converters to RestTemplateConfiguration

sumansreenivasan:
As mentioned in RDM-4449 moved converters to RestTemplateConfiguration to prevent the converters being set on RestTemplate every time a document is retrieved and have updated the associated tests.

If this converters is needed only for Documents, should there be a different RestTemplateConfiguration with converts only for document retrieval or is it ok to be added in RestTemplateConfiguration?

https://tools.hmcts.net/jira/browse/RDM-4449

### Change description ###
As mentioned in RDM-4449 moved converters to RestTemplateConfiguration to prevent the converters being set on RestTemplate every time a document is retrieved and have updated the associated tests.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
@@ -9,10 +9,16 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.http.MediaType;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
 import org.springframework.web.client.RestTemplate;
 
 import javax.annotation.PreDestroy;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 @Configuration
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/ccd/RestTemplateConfiguration.java
```diff
@@ -48,11 +54,21 @@
...
         restTemplate.setRequestFactory(requestFactory);
+
+        List<HttpMessageConverter<?>> messageConverters = new ArrayList<>();
+        //Add the Jackson Message converter
+        MappingJackson2HttpMessageConverter converter = new MappingJackson2HttpMessageConverter();
+        // Note: here we are making this converter to process any kind of response,
+        // not only application/*json, which is the default behaviour
+        converter.setSupportedMediaTypes(Collections.singletonList(MediaType.ALL));
+        messageConverters.add(converter);
+        restTemplate.setMessageConverters(messageConverters);
```
mario-paniccia, line 71: We are changing the configuration of the default rest template used by all http calls in Data Store. Before we do this, please triple check this configuration you are adding is suitable for all the calls and won't cause problems/regressions. 
If there are doubts about it, as an alternative to minimise impact, we can introduce a new method here that creates a rest template dedicated to the `DocumentManagementRestClient`. And apply this configuration only there. And inject that in the `DocumentManagementRestClient`

src/main/java/uk/gov/hmcts/ccd/domain/types/sanitiser/client/DocumentManagementRestClient.java
```diff
@@ -43,18 +38,6 @@ public Document getDocument(FieldTypeDefinition fieldTypeDefinition, String url)
...
-        //Add the Jackson Message converter
-        MappingJackson2HttpMessageConverter converter = new MappingJackson2HttpMessageConverter();
-        // Note: here we are making this converter to process any kind of response,
-        // not only application/*json, which is the default behaviour
-        converter.setSupportedMediaTypes(Collections.singletonList(MediaType.ALL));
-        messageConverters.add(converter);
-        restTemplate.setMessageConverters(messageConverters);
-
         Document document = null;
         try {
```
mario-paniccia, line 42: I liked the suggestion in the ticket to change the response code from 400 to 500. Shall we do that?

sumansreenivasan:
Mario has suggested to add a different separate RestTemplate for Document.

PR #1018, opened 2020-07-20
RDM-9045 - Add configurable logging of roles for requests

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-9045

### Change description ###
Add configurable logging of roles for requests

```
[ ] Yes
[x] No
```


charts/ccd-data-store-api/values.yaml
```diff
@@ -60,3 +60,6 @@ java:
     HTTP_CLIENT_SECONDS_IDLE_CONNECTION: '120'
     HTTP_CLIENT_MAX_CLIENT_PER_ROUTE: '20'
     HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY: '2000'
+
+    DATA_STORE_ROLE_LOGGING_ENABLED: 'false'
+    DATA_STORE_ROLE_LOGGING_PATH_REGEX: '.*'
```
danlysiak, line 65: These can be overridden in Flux config when role logging is required.

src/main/resources/application.properties
```diff
@@ -116,6 +116,7 @@ logging.level.uk.gov.hmcts.ccd=${DATA_STORE_DEFAULT_LOG_LEVEL:INFO}
 # logging.level.org.springframework.web=DEBUG
 #logging.level.org.hibernate=INFO
 #logging.level.org.hibernate.type.descriptor.sql=trace
+#logging.level.uk.gov.hmcts.ccd.security.filters.RoleLoggingFilter=DEBUG
```
danlysiak, line 119: Can be uncommented if desired for local dev.

src/main/java/uk/gov/hmcts/ccd/security/filters/RoleLoggingFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    public RoleLoggingFilter(AppInsights appInsights, SecurityUtils securityUtils, @Value("${role.logging.path.regex}") String pathRegex) {
+        this.appInsights = appInsights;
+        this.securityUtils = securityUtils;
+        this.pathPattern = Pattern.compile(pathRegex);
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+        throws ServletException, IOException {
+        String message = buildLogMessage(request, securityUtils.getUserInfo());
```
smathangi, line 42: i guess it won't log in case of auth failures. Is that OK?

danlysiak: That's right, on the 3A was decided we only want it for authorised requests 👍

charts/ccd-data-store-api/values.yaml
```diff
@@ -60,3 +60,6 @@ java:
     HTTP_CLIENT_SECONDS_IDLE_CONNECTION: '120'
     HTTP_CLIENT_MAX_CLIENT_PER_ROUTE: '20'
     HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY: '2000'
+
+    DATA_STORE_ROLE_LOGGING_ENABLED: 'false'
```
MSancaktutar, line 64: I think instead of switching the logging on and off, we should just change the logging level per package/class.

src/main/java/uk/gov/hmcts/ccd/security/filters/RoleLoggingFilter.java
MSancaktutar: Well, I would suggest to log all requests no matter authorisation passes or not. We had some cases to diagnose with role-based authorisation failures and we may have more.

src/main/resources/application.properties
```diff
@@ -162,3 +163,7 @@ ccd.am.read.from_am=${CCD_AM_READ_FROM_AM:}
 # http statuses with comma separated to ignore audit logs
 audit.log.ignore.statues=404
 audit.log.enabled=true
+
+# Role logging
+role.logging.enabled=${DATA_STORE_ROLE_LOGGING_ENABLED:false}
```
MSancaktutar, line 168: I was thinking to have only a log-level alignment, rather than a logical switch to enable/disable the related logging at all.

src/main/java/uk/gov/hmcts/ccd/security/filters/RoleLoggingFilter.java
danlysiak: Apologies @MSancaktutar, in my haste I put authorised when I meant authenticated.. If a request is unauthorised (e.g. user requests case they don't have access to) it DOES still log!

MSancaktutar: Well, in fact, what I meant was... A user's roles not being sufficient to access and operation would be an authorisation failure. And even in that case, we should log the roles.

src/main/java/uk/gov/hmcts/ccd/security/filters/LoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.regex.Pattern;
+
+@Slf4j
+public class LoggingFilter extends OncePerRequestFilter {
```
mario-paniccia, line 16: I think this name is a bit too generic. We can't log everything with this filter. Maybe a bit more specific name would help

src/main/java/uk/gov/hmcts/ccd/security/filters/LoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+import uk.gov.hmcts.reform.idam.client.models.UserInfo;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.regex.Pattern;
+
+@Slf4j
```
mario-paniccia, line 15: can you add a short Javadoc description please

charts/ccd-data-store-api/values.yaml
```diff
@@ -60,3 +60,5 @@ java:
     HTTP_CLIENT_SECONDS_IDLE_CONNECTION: '120'
     HTTP_CLIENT_MAX_CLIENT_PER_ROUTE: '20'
     HTTP_CLIENT_VALIDATE_AFTER_INACTIVITY: '2000'
+
+    DATA_STORE_LOGGING_FILTER_PATH_REGEX: '.*'
```
mario-paniccia, line 64: this name is a bit too generic. Maybe a more specific name would help

lib/AI-Agent.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <ApplicationInsightsAgent>
-    <AgentLogger>INFO</AgentLogger>
+    <AgentLogger>DEBUG</AgentLogger>
```
mario-paniccia, line 3: can we check with Platform Engineering to see that they are ok with this? not sure what the conventions are

src/main/java/uk/gov/hmcts/ccd/security/filters/LoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public LoggingFilter(SecurityUtils securityUtils, String pathRegex) {
+        this.securityUtils = securityUtils;
+        this.pathPattern = Pattern.compile(pathRegex);
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+        throws ServletException, IOException {
+        log.debug(buildLogMessage(request, securityUtils.getUserInfo()));
```
mario-paniccia, line 31: isn't it simpler to just log on INFO? it's definitely an important type of information that can be logged as INFO. 
This would be a safer approach as we wouldn't need to change the AppInsights log level

src/main/java/uk/gov/hmcts/ccd/security/filters/LoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.regex.Pattern;
+
+@Slf4j
+public class LoggingFilter extends OncePerRequestFilter {
+
+    private static final String LOG_MESSAGE_TEMPLATE = "[LOG FILTER] Attempting to serve request %s %s for user with IDAM roles %s";
```
mario-paniccia, line 18: doesn't LOG4J add the class name (`[LOG FILTER]`) already by default?

lib/AI-Agent.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <ApplicationInsightsAgent>
-    <AgentLogger>INFO</AgentLogger>
```
mario-paniccia, line 3: I know the AppInsights client library is performant but that would be much more logging we are sending so maybe we should verify the impact on DataStore performance?

lib/AI-Agent.xml
```diff
@@ -1,6 +1,6 @@
...
+    <AgentLogger>DEBUG</AgentLogger>
     <Instrumentation>
         <BuiltIn enabled="true">
```
mario-paniccia, line 5: was just wondering if it's simpler to just log the user info at INFO level. Less changes would be required.

src/main/java/uk/gov/hmcts/ccd/security/filters/LoggingFilter.java
danlysiak: Initially added to enable searching by a consistent term on AI (as L&A does), but good point, probably not much point as can still search by logger name. Removing.

lib/AI-Agent.xml
MSancaktutar: In fact, by having the root level logging at INFO level, we are writing no debug logs at all, by default.
By configuring only a specific class to log at DEBUG level, we are writing logs in debug level only in that class, if it contains some.

The AI agent is sending anything it receives at or above DEBUG, but the resultant data pushed to Azure is only INFO Logs + any specific debug log. This is actually a nicer way of using the log levels, as you'd also appreciate.

Re sensitive info exposures, you're right. But that concern already applies to all logs.
One can write sensitive data in a log relying on that it's not visible anywhere, but then later config changes may expose such data. So, we are already assuring no logging of any sensitive data at all.

If we use only Info Logs, it also prevents us from use of debug logs in many many libraries that come with built-in logging at debug level. In the case of a need for, let's say, a debugging with help of a  Spring library logs, we will have to set the log level for a specific Spring class or package, but the current config will not allow the specific Spring debug log to appear on AI Resources.

Obviously, pushing to Azure debug logs is not something we can do arbitrarily, as it does come with its costs, but when we enable a debug log, we do so by expressly knowing and managing the fact that this is a temporary and very limited scope of debug logging for a good reason - which happens very rarely, like in this situation where we need some genuinely 'debug' logs.

Well, in short,
The idea is, actually, to be able to preserve the debug logging capability in the full stack of the app at a class-level granularity if and only when needed - and disabled by default by having root log level at INFO.

timja: this is just the agent logger, unless you are debugging something in the app insights agent I would be very surprised that you would need to change this.

MSancaktutar: That's even better. So, it means the actual log level pushed to Azure is defined solely at application properties - nothing managed at AI Agent?

timja: Its configured here:
https://github.com/hmcts/java-logging/blob/master/src/main/resources/logback.xml#L5

PR #1023, opened 2020-07-21
RDM-8829  - flag to enable/disable case-assignments rest api

smathangi:


src/main/resources/application.properties
```diff
@@ -173,6 +173,9 @@ audit.log.enabled=true
 ccd.access-control.cross-jurisdictional-roles=${DATA_STORE_CROSS_JURISDICTIONAL_ROLES:caseworker-caa}
 ccd.access-control.citizen-roles=${DATA_STORE_CITIZEN_ROLES:citizen,letter-holder}
 
+api.case-assigned.enabled=${API_CASE_ASSIGNED_ENABLED:true}
```
MSancaktutar, line 176: Could this be changed like below, please? (along with implied env var to come from flux)
`ccd.conditional-apis.case-assigned-users-and-roles.enabled=...`

smathangi: Ok, will do. Seems to be very long variable 😄

PR #1026, opened 2020-07-22
RDM-9095 - Filter out cross-jurisdiction roles when determining user jurisdictions

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-9095

### Change description ###
Filter out cross-jurisdiction roles when determining user jurisdictions

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -179,7 +180,9 @@ public String getUserId() {
     public List<String> getUserRolesJurisdictions() {
         String[] roles = this.getUserDetails().getRoles();
 
-        return Arrays.stream(roles).map(role ->  role.split("-"))
+        return Arrays.stream(roles)
```
mario-paniccia, line 183: I was thinking whether we can take this opportunity to also make the method a bit more precise. 
With the current implementation any role of the form part1-part2-....partN will be considered a "jurisidictional" role for a jurisdiction "part2".
In reality this (not written) rule applies only for caseworkers (as far as I know).
So was wondering if it would be better applying this method only to caseworkers roles (maybe using a regex?), and rename it to `getCaseworkerUserRolesJurisdictions`?

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -234,4 +223,12 @@ private IdamUser toIdamUser(UserInfo userInfo) {
         return idamUser;
     }
 
+    private boolean isCrossJurisdictionRole(String role) {
+        return applicationParams.getCcdAccessControlCrossJurisdictionRoles().contains(role);
+    }
```
mario-paniccia, line 228: this could be a good candidate to be moved to UserRepository

src/main/resources/application.properties
```diff
@@ -179,3 +179,6 @@ ccd.conditional-apis.case-assigned-users-and-roles.enabled=${CCD_CONDITIONAL_API
 # (Note SecurityLoggingFilter logs at debug)
 security.logging.filter.path.regex=${DATA_STORE_SECURITY_LOGGING_FILTER_PATH_REGEX:.*}
 
+caseworker.role.regex=${CASEWORKER_ROLE_REGEX:caseworker.+}
+caseworker.role.delimiter=${CASEWORKER_ROLE_DELIMITER:-}
+caseworker.role.jurisdiction.element.index=${CASEWORKER_ROLE_JURISDICTION_ELEMENT_INDEX:1}
```
mario-paniccia, line 184: I appreciate the intent to make this configurable. 
But I think the rules around the format of the caseworker IDAM role format are very unlikely to change now. I doubt we'll ever need to use a delimiter different that `-` and a jur position different than `1`. Also if that was the case CCD would break in many other places already. We can therefore simplify things a bit removing these 2 config options if we wanted.

src/main/resources/application.properties
```diff
@@ -179,3 +179,6 @@ ccd.conditional-apis.case-assigned-users-and-roles.enabled=${CCD_CONDITIONAL_API
 # (Note SecurityLoggingFilter logs at debug)
 security.logging.filter.path.regex=${DATA_STORE_SECURITY_LOGGING_FILTER_PATH_REGEX:.*}
 
+caseworker.role.regex=${CASEWORKER_ROLE_REGEX:caseworker.+}
```
mario-paniccia, line 182: can you add these under the section: 
`# List of roles which are comma separated`
and prefix them with:
`ccd.access-control`
please

src/main/resources/application.properties
mario-paniccia: or, to simplify, couldn't we combine these 3 properties into one property called:
ccd.access-control.caseworker-jurisdiction.regex-group:caseworker-(.+)-.*
?

PR #1036, opened 2020-07-30
Disable es ft as on feature branches

MSancaktutar:
none

### Change description ###
Disable ES FTAs on feature branches.

```
[ ] Yes
[X] No
```


Jenkinsfile_CNP
```diff
@@ -79,7 +79,7 @@ env.DM_STORE_BASE_URL = "http://dm-store-aat.service.core-compute-aat.internal"
 env.DEFINITION_STORE_HOST = "http://ccd-definition-store-api-pr-575.service.core-compute-preview.internal"
 env.BEFTA_RESPONSE_HEADER_CHECK_POLICY="JUST_WARN"
 env.ELASTIC_SEARCH_HOSTS = "http://es-ccd-data-store-api-pr-${CHANGE_ID}.service.core-compute-preview.internal"
-env.ELASTIC_SEARCH_ENABLED = "true"
+env.ELASTIC_SEARCH_ENABLED = "false"
```
smathangi, line 82: LGTM. I think we should rename variable specific to tests otherwise it confuses with actual app property.

MSancaktutar: Good point, and then this made me notice a lot more to cleanup from legacy ES tests. Cleaned up by occasion.

PR #1121, opened 2020-09-17
RDM 9505

tomxelliott:

https://tools.hmcts.net/jira/browse/RDM-9505

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/resources/db/changelog/db.changelog-RDM-9505.xml
+++ b/src/main/resources/db/changelog/db.changelog-RDM-9505.xml
@@ -10,16 +10,16 @@
              stripComments="true">
             DROP INDEX IF EXISTS idx_case_data__generated_dob;
             DROP INDEX IF EXISTS idx_case_data_generatedemail;
-            CREATE INDEX IF NOT EXISTS idx_case_data_appeal_appellant_identity_lastName_region ON public.case_data USING BTREE (BTRIM(UPPER((data #>> '{appeal,appellant,name,lastName}'::text[]))), btrim(upper((data #>> '{region}'::text[]))));
-            CREATE INDEX IF NOT EXISTS idx_case_data_documentSentToDwp ON public.case_data USING BTREE (BTRIM(UPPER((data #>> '{documentSentToDwp}'::text[]))));
-            CREATE INDEX IF NOT EXISTS idx_case_data_isscottishcase_documentSentToDwp ON public.case_data USING BTREE (BTRIM(UPPER((data #>> '{isScottishCase}'::text[]))), btrim(upper((data #>> '{documentSentToDwp}'::text[]))));
-            CREATE INDEX IF NOT EXISTS idx_case_data_appeal_appellant_address_postcode ON public.case_data USING BTREE (BTRIM(UPPER((data #>> '{appeal,appellant,address,postcode}'::text[]))));
-            CREATE INDEX IF NOT EXISTS idx_case_data_createdingapsfrom_dwpRegionalCentre ON public.case_data USING BTREE (BTRIM(UPPER((data #>> '{createdInGapsFrom}'::text[]))), btrim(upper((data #>> '{dwpRegionalCentre}'::text[]))));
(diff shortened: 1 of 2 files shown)


src/main/resources/db/changelog/db.changelog-RDM-9505.xml
```diff
@@ -0,0 +1,25 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+
+    <changeSet id="rdm-9505" author="thomas.elliot@hmcts.net">
+        <sql dbms="postgresql"
+             endDelimiter="\nGO"
+             splitStatements="true"
+             stripComments="true">
+            DROP INDEX idx_case_data__generated_dob;
+            DROP INDEX idx_case_data_generatedemail;
```
MSancaktutar, line 12: Dropping these indexes with liquibase may later on cause an issue should the same indexes are re-created for some reason and liquibase scripts are not aligned yet. I'd suggest removing these lines and manually dropping these and all other unused indexes, say, not scanned for the last 3 months. We need to also make sure these indexes are not getting created by other liquibase change sets. If they are, we need to remove basically any references to these indexes.

MSancaktutar (changes requested):
Pls see my comment on the indexes to remove.

src/main/resources/db/changelog/db.changelog-RDM-9505.xml
danlysiak: @MSancaktutar We can't remove references from previous Liquibase migrations (and these indexes are in previous ones) otherwise the Liquibase checksum validation will error. These have already been dropped in prod so this will correctly align us, and if they need to be created again in future then a future migration that does it wouldn't cause an issue.

tomxelliott: Since the scripts are executed sequentially if someone adds them in a later script the drop being here won't cause an issue.

tomxelliott: Although that being said if you would prefer I remove the 2 drop statements, happy to do so 👍

MSancaktutar: Well, of course later scripts to create these are fully safe from a drop statement like above.
But they can still drop indexes that may be manually created in order to, say, intervene a production performance issue.
@danlysiak  - I understand Liquibase will not let us arbitrarily edit past scripts, as they are supposed to create or update the database structure always to a consistent and safe state. However, when a series of changes contain a zero-net-structural-effect type of do/undo things, and if such do/undo statements brings unacceptible side effect of, say, re-constructing an index everytime after a deployment (causing full table scans, dropping database performance beyond just a few seconds during releases), then I'm pretty sure Liquibase also provides a due process and technique to eliminate such no effect changes. On the other hand, Liquibase may also be smart enough to first process the mathematical equivalent of the full series of changes and only after that execute the net mathematical equivalent, then things may be a lot easier to manage. Let's discuss on an interactive call, please.
@danlysiak @tomxelliott

danlysiak (approved):
👍

src/main/resources/db/changelog/db.changelog-RDM-9505.xml
MSancaktutar: Considering the fact that only the incremental delta will be executed in any next round, I'm in full peace of mind with the way index objects are removed. Many thanks Dan & Tom.

MSancaktutar (approved):
LGTM

mario-paniccia (changes requested):
last time we've created this large amount of indexes all in one go we locked the DataStore DB for 15 minutes creating Prod issues with users not able to use CCD. A P1 was raised.
One solution could be to create indexes concurrently: 
https://www.postgresql.org/docs/9.1/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY
Can you have a look please

mario-paniccia (approved):
👍

VamshiRudrab:
I am happy for @tomxelliott to merge this PR.

PR #1167, opened 2020-10-15
RDM-9868 DefaultValue for Complex fields

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-9868

### Change description ###
If defaultValue is present in the EventToComplexTypes tab we add it to all event triggers. 

```
[ ] Yes
[ X] No
```


src/main/java/uk/gov/hmcts/ccd/config/JacksonUtils.java
```diff
@@ -29,7 +36,7 @@ private JacksonUtils(){
         .build();
 
 
-    public static HashMap<String, JsonNode> convertValue(Object from) {
+    public static Map<String, JsonNode> convertValue(Object from) {
```
mario-paniccia, line 39: 👍

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventDefinition.java
```diff
@@ -205,4 +214,29 @@ public void setCanSaveDraft(Boolean canSaveDraft) {
...
+                    String next = e.fieldNames().next();
+                    objectNode.set(next, e.findValue(next));
+                });
+                if (!collect.isEmpty()) { // to prevent construct like "FieldA": {}
+                    data.put(caseField.getCaseFieldId(), objectNode);
+                }
+            });
+
+        return data;
+    }
```
mario-paniccia, line 241: mmm not sure whether it's the best thing in terms of separation of concerns on having a definition object having logic to build JSON...

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
```diff
@@ -1,10 +1,11 @@
 package uk.gov.hmcts.ccd.domain.service.startevent;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.Maps;
```
mario-paniccia, line 4: mmm isn't it a bit too late to set the default values in the operation classes? For example, when invoking a MidEvent callback CCD sends to the midevent callback the case data, here: 

https://github.com/hmcts/ccd-data-store-api/blob/80adc76067063ba3c3600fb3e0674b41bfe5426f/src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java#L69

With this implementation the callback is not receiving the default value right?

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
```diff
@@ -76,12 +80,15 @@ public StartEventResult triggerStartForCaseType(final String caseTypeId,
 
         final CaseTypeDefinition caseTypeDefinition = getCaseType(caseTypeId);
 
+        final Map<String, JsonNode> data = Maps.newHashMap();
+
+        CaseDetails newCaseDetails = caseService
+            .createNewCaseDetails(caseTypeId, caseTypeDefinition.getJurisdictionId(), data);
```
mario-paniccia, line 86: probably better having the CaseService.createNewCaseDetails setting the defaults? we could pass the whole `caseTypeDefinition`.
The idea is that all the operations use the CaseService class, and it's the service class the only one responsible for setting the defaults, rather than the various operations. This would eliminate potential bugs where we get the case from the service but forget to set the defaults. And would also reduce duplication

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
```diff
@@ -99,6 +106,9 @@ public StartEventResult triggerStartForCase(final String caseReference,
```
mario-paniccia, line 106: mmm perhaps line 99 
`final CaseDetails caseDetails = getCaseDetails(caseReference);`
should already return the case with the default values? We should change `getCaseDetails` to use the Service rather than the Repository directly, and have the CaseService dealing with defaulting

mario-paniccia (commented):
some comments

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventDefinition.java
rafalkalita: Not a problem. Can move it to the Service layer. Was just thinking of 'anemic domain model'

kiran-yenigala-hmcts: Yes, Good to move this logic to service class

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
kiran-yenigala-hmcts: @mario-paniccia 
Do we need to worry about the mid event callback, as we default values will be populated in the about to start event callback already....

src/main/java/uk/gov/hmcts/ccd/domain/service/validate/DefaultValidateCaseFieldsOperation.java
```diff
@@ -98,12 +96,10 @@ private void validateContent(final CaseDataContent content,
             int length = referenceArray.length;
             String nodeReference = length > 1 ? referenceArray[length - 2] : referenceArray[length - 1];
             List<JsonNode> parentNodes = caseFieldNode.findParents(nodeReference);
-            parentNodes.stream().forEach(parentNode -> {
+            parentNodes.forEach(parentNode -> {
                 JsonNode orgPolicyNode = findOrgPolicyNode(nodeReference, parentNode, length);
                 if (orgPolicyNode.isNull()) {
                     errorList.add(caseFieldId + " cannot have an empty value.");
-                } else if (!defaultValue.equalsIgnoreCase(orgPolicyNode.textValue())) {
```
kiran-yenigala-hmcts, line 105: We are just validating null value, what if user changes the OrgPolicyRole, that means if user enters any other value in this field, we need to check whether that case-role exists or not

kiran-yenigala-hmcts (commented):
Couple of comments

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventDefinition.java
```diff
@@ -205,4 +214,29 @@ public void setCanSaveDraft(Boolean canSaveDraft) {
...
+                    .filter(e -> !e.getReference().isBlank())
+                    .map(caseEventFieldComplex -> JacksonUtils.buildFromDottedPath(caseEventFieldComplex.getReference(),
+                                                                                   caseEventFieldComplex.getDefaultValue())).collect(toList());
+
+                ObjectNode objectNode = MAPPER.getNodeFactory().objectNode();
+                collect.forEach(e -> {
+                    String next = e.fieldNames().next();
+                    objectNode.set(next, e.findValue(next));
+                });
+                if (!collect.isEmpty()) { // to prevent construct like "FieldA": {}
```
kiran-yenigala-hmcts, line 235: can we do this check before line 230

kiran-yenigala-hmcts (commented):
Minor Comment

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
mario-paniccia: This thing is that the MidEven also sends the caseDetailsBefore fields, which should contain the case data before the modifications done in the event. Now, should that case data contain default values? In my opinion yes. If we say this fields is defaulted to a value, then in CCD that field has got that initial value.

mario-paniccia: @rafalkalita let's ask Nigel please. It would also be ok for now to ignore this comment.

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
mario-paniccia: We can ignore this comment.
Rafal correctly pointed out that since we consider the defaulting a concept only related to events, it's ok to leave to the Operation the responsibility to do the defaulting, rather than the create methods and get case methods of the `CaseService` class. Because we don't want defaulting to apply every time a case is retrieved.

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
mario-paniccia: we can ignore this

src/main/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperation.java
rafalkalita: Nigel confirmed we do not need this: "Defaults only work for start event so not for midevent"

src/main/java/uk/gov/hmcts/ccd/domain/service/validate/DefaultValidateCaseFieldsOperation.java
rafalkalita: Confirmed with Nigel we need it. Implementing.

src/main/java/uk/gov/hmcts/ccd/domain/service/validate/DefaultValidateCaseFieldsOperation.java
```diff
@@ -62,15 +69,15 @@ private void validateOrganisationPolicy(String caseTypeId, CaseDataContent conte
...
-                .stream()
                 .forEach(eventFieldDefinition -> eventFieldDefinition.getCaseEventFieldComplexDefinitions().stream()
                     .filter(cefcDefinition -> isOrgPolicyCaseAssignedRole(cefcDefinition.getReference()))
                     .forEach(cefcDefinition -> {
                         String reference = cefcDefinition.getReference();
-                        validateContent(content, eventFieldDefinition.getCaseFieldId(),
-                            reference,
-                            cefcDefinition.getDefaultValue(),
-                            errorList);
+                        validateContent(content,
```
mario-paniccia, line 76: Probably nothing to do with your PR @rafalkalita, but I just noticed that we are adding custom validation rules (for the organisation) in this class. But this class is a generic one! Why is a method like:
`private void validateOrganisationPolicy(String caseTypeId, CaseDataContent content) {` 
here?

mario-paniccia, line 76: this is getting very complex....

mario-paniccia (approved):
👍

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventDefinition.java
rafalkalita: done

src/main/java/uk/gov/hmcts/ccd/domain/service/validate/DefaultValidateCaseFieldsOperation.java
rafalkalita: Done

hemantt (approved):

PR #1171, opened 2020-10-19
RDM-8898: Removed credentials

rajaram-nagarajan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/RDM-8898

### Change description ###

Eliminate all Credentials in CCD Repos Exposed for Lower Environments

```
[ ] Yes
[x] No
```


rafalkalita:
No need to remove credentials from application.properties. It is used only for local deployment. We use the secrets from the Vault using https://github.com/hmcts/properties-volume-spring-boot-starter.
https://github.com/hmcts/ccd-data-store-api/blob/master/src/main/resources/bootstrap.yaml#L12 will set the DATA_STORE_DB_PASSWORD
https://github.com/hmcts/ccd-data-store-api/blob/master/charts/ccd-data-store-api/values.yaml#L29 will set the DATA_STORE_DB_USERNAME

Secrets are defined in https://github.com/hmcts/ccd-data-store-api/blob/master/charts/ccd-data-store-api/values.yaml#L16 and we might want to add DATA_STORE_DB_USERNAME to the secrets.
Interesting is the data-store-api-POSTGRES-USER is defined in https://github.com/hmcts/ccd-data-store-api/blob/master/infrastructure/main.tf#L72, but not used ... I think.

rajaram-nagarajan:
> No need to remove credentials from application.properties. It is used only for local deployment. We use the secrets from the Vault using https://github.com/hmcts/properties-volume-spring-boot-starter.
> https://github.com/hmcts/ccd-data-store-api/blob/master/src/main/resources/bootstrap.yaml#L12 will set the DATA_STORE_DB_PASSWORD
> https://github.com/hmcts/ccd-data-store-api/blob/master/charts/ccd-data-store-api/values.yaml#L29 will set the DATA_STORE_DB_USERNAME
> 
> Secrets are defined in https://github.com/hmcts/ccd-data-store-api/blob/master/charts/ccd-data-store-api/values.yaml#L16 and we might want to add DATA_STORE_DB_USERNAME to the secrets.
> Interesting is the data-store-api-POSTGRES-USER is defined in https://github.com/hmcts/ccd-data-store-api/blob/master/infrastructure/main.tf#L72, but not used ... I think.

Resolved

rafalkalita (approved):
LGTM

PR #1184, opened 2020-10-27
Rdm 9052

henrykcgi:
### Change description ###

```
[ ] Yes
[X] No
```


liquibase.gradle
```diff
@@ -3,7 +3,7 @@ configurations {
 }
```
mario-paniccia, line 4: in the previous PR we introduced db.changelog-RDM-9052.xml
Are we able to get rid of it now after the update to the newest liquibase version or we still need that?

henrykcgi: We got rid of it.

mario-paniccia (approved):
👍

src/main/resources/db/changelog/db.changelog-master.xml
```diff
@@ -39,6 +39,5 @@
     <include file="db/changelog/db.changelog-RDM-8463.xml"/>
     <include file="db/changelog/db.changelog-RDM-9287.xml"/>
     <include file="db/changelog/db.changelog-RDM-9505.xml"/>
-    <include file="db/changelog/db.changelog-RDM-9052.xml"/>
```
mario-paniccia, line 42: just removing the file won't work on environments were the changelog has already been applied, for example Demo. For that case you'd need to explicitly introduce a new changelog that undo the changed introduced by `changelog-RDM-9052`. And which runs only if `changelog-RDM-9052` has run already.

noronhaa (approved):

PR #1279, opened 2020-12-09
DTSPO-266 Enable self service read only access

timja:
https://tools.hmcts.net/confluence/display/DTSPO/%5BSelf-Service%5D+Database


MSancaktutar (approved):
LGTM

PR #1292, opened 2020-12-17
RDM-6873: Event Enabling Condition - Data Store Changes

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-6873

### Change description ###

Event enabling condition added to CaseEvent

```
[ ] Yes
[X ] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
```diff
@@ -0,0 +1,27 @@
...
+public class CaseEventEnablingService {
+
+    private final EnablingConditionParser enablingConditionParser;
+
+    @Inject
+    public CaseEventEnablingService(EnablingConditionParser enablingConditionParser) {
+        this.enablingConditionParser = enablingConditionParser;
+    }
+
+    public Boolean evaluate(String enablingCondition,
```
mario-paniccia, line 20:
```suggestion
    public Boolean isEventEnabled(String enablingCondition,
```

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
```diff
@@ -0,0 +1,27 @@
...
+                           Map<String, JsonNode> caseEventData) {
```
mario-paniccia, line 21: something weird here. Show condition can have metadata fields in them, i.e. [STATE]=someState. 
I would have thought you needed to pass the whole case so that metadata can be evaluated as well. Instead we are passing only the case fields in the data field

mario-paniccia: If this issue is confirmed, can we also add some integration tests for event enabling conditions with metadata in show conditions to avoid regressions

mario-paniccia: I'm thinking, do we have the same issue with case post states? Do we support metadata fields in show conditions there? we should...

mario-paniccia (commented):
some comments

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
kiran-yenigala-hmcts: We are not supporting metadata fields in case post states for now...

Thor-tech-of-metal (approved):
LGTM

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
kiran-yenigala-hmcts: done

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
kiran-yenigala-hmcts: > something weird here. Show condition can have metadata fields in them, i.e. [STATE]=someState.
> I would have thought you needed to pass the whole case so that metadata can be evaluated as well. Instead we are passing only the case fields in the data field

changed to include CaseDetails and now supporting meta data as well in this..

src/aat/resources/features/F-127 - Even Enabling Condition/S-127.2.td.json
```diff
@@ -0,0 +1,49 @@
+{
+	"_guid_": "S-127.2",
+  "_extends_": "F-127_Get_Case_View_Base",
+  "title": "Get Case event view to validate the events returned in the response when event enabling conditions are valid",
```
rafalkalita, line 4: Should it be "are not valid" at the end? S-127.2 is for a non matching condition.

src/aat/resources/features/F-127 - Even Enabling Condition/S-127.1.td.json
```diff
@@ -0,0 +1,56 @@
+{
+	"_guid_": "S-127.1",
+  "_extends_": "F-127_Get_Case_View_Base",
+  "title": "Get Case event view to validate the events returned in the response when event enabling conditions are not valid",
```
rafalkalita, line 4: Should it be "are valid" at the end? S-127.1 is for a matching condition.

rafalkalita (approved):
Looks good. Just two minor comments.

mario-paniccia (approved):
👍

PR #1295, opened 2020-12-30
RDM-10671: increased HTTP max header size to address issue reported i…

mario-paniccia:
…n RDM-10274

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -4,7 +4,7 @@ server.servlet.contextPath=
 server.compression.enabled=true
 server.compression.mime-types=application/json,application/xml,text/html,text/xml,text/plain
 server.error.include-message=always
-server.max-http-header-size=${MAX_HTTP_HEADER_SIZE:24KB}
+server.max-http-header-size=100KB
 
 spring.cloud.discovery.client.composite-indicator.enabled=false
 spring.jpa.database=postgresql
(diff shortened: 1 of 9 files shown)


kiran-yenigala-hmcts (approved):
LGTM

src/main/resources/application.properties
```diff
@@ -4,6 +4,7 @@ server.servlet.contextPath=
 server.compression.enabled=true
 server.compression.mime-types=application/json,application/xml,text/html,text/xml,text/plain
 server.error.include-message=always
+server.max-http-header-size=${MAX_HTTP_HEADER_SIZE:24KB}
```
MSancaktutar, line 7: A flux change has set this to 100KB on demo, already, for fixing an issue on case sharing. (under RDM-10571)
I think, it's better to either change the default value here to 100KB or raise an additional flux change like that one below.
https://github.com/hmcts/cnp-flux-config/pull/7414/files

MSancaktutar: Also, since "SERVER_MAX_HTTP_HEADER_SIZE" itself is also as good as "MAX_HTTP_HEADER_SIZE", we may refrain from introducing a new env var like this. Setting "SERVER_MAX_HTTP_HEADER_SIZE" in any environment will make it anyway... If we still want to introduce "MAX_HTTP_HEADER_SIZE", we will need to change the demo flux var for consistency.

MSancaktutar (changes requested):
Please consider 2 comments for consistency with a prior similar change.

src/main/resources/application.properties
mario-paniccia: ok it makes sense. I'll just set it to 100KB, get rid of the custom env var, since we can use `SERVER_MAX_HTTP_HEADER_SIZE` in the future if we need to increase more

MSancaktutar (approved):
LGTM.

PR #1369, opened 2021-02-18
RDM-10775: Implement Filtering of Role Assignments

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-10775

### Change description ###

Filter role assignments

```
[ ] Yes
[X ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/MatchingResults.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/MatchingResults.java
@@ -1,0 +1,13 @@
+package uk.gov.hmcts.ccd.domain.model.casedataaccesscontrol;
+
+import lombok.Data;
+
+@Data
+public class MatchingResults {
+    boolean validDate;
+    boolean validCaseId;
+    boolean validJurisdiction;
+    boolean validRegion;
+    boolean validLocation;
+    boolean validClassification;
+}
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignments.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignments.java
@@ -1,0 +1,15 @@
+package uk.gov.hmcts.ccd.domain.service.accessprofile.filter;
+
+import uk.gov.hmcts.ccd.domain.model.casedataaccesscontrol.RoleAssignments;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseDetails;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseTypeDefinition;
+import uk.gov.hmcts.ccd.domain.model.std.CaseDataContent;
+
+public interface FilterRoleAssignments {
+
+    RoleAssignments filter(RoleAssignments roleAssignments, CaseDetails caseDetails);
+
+    RoleAssignments filter(RoleAssignments roleAssignments, CaseDataContent caseDataContent);
+
+    RoleAssignments filter(RoleAssignments roleAssignments, CaseTypeDefinition caseTypeDefinition);
+}
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
@@ -1,0 +1,139 @@
+package uk.gov.hmcts.ccd.domain.service.accessprofile.filter;
+
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
+import uk.gov.hmcts.ccd.data.casedetails.SecurityClassification;
+import uk.gov.hmcts.ccd.domain.model.casedataaccesscontrol.RoleAssignment;
+import uk.gov.hmcts.ccd.domain.model.casedataaccesscontrol.RoleAssignments;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseDetails;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseTypeDefinition;
+import uk.gov.hmcts.ccd.domain.model.std.CaseDataContent;
+import uk.gov.hmcts.ccd.domain.service.AccessControl;
+import uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationUtils;
+
+import static uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationUtils.caseHasClassificationEqualOrLowerThan;
+
+@Slf4j
+@Component
+public class FilterRoleAssignmentsImpl implements FilterRoleAssignments, AccessControl {
+
+    @Override
+    public RoleAssignments filter(RoleAssignments roleAssignments, CaseDetails caseDetails) {
+        log.info("Filter role assignments for case {}", caseDetails.getReference());
+
+        List<RoleAssignment> roleAssignmentList = roleAssignments.getRoleAssignments().stream()
+            .filter(roleAssignment -> validateStartAndEndDate(roleAssignment))
+            .filter(roleAssignment -> validateSecurityValidation(caseDetails, roleAssignment))
+            .filter(roleAssignment -> validateCaseId(caseDetails, roleAssignment))
+            .filter(roleAssignment -> validateJurisdiction(caseDetails, roleAssignment))
+            .filter(roleAssignment -> validateLocation(roleAssignment))
+            .filter(roleAssignment -> validateRegion(roleAssignment))
+            .collect(Collectors.toList());
+
+        return RoleAssignments.builder().roleAssignments(roleAssignmentList).build();
+    }
+
+    @Override
+    public RoleAssignments filter(RoleAssignments roleAssignments, CaseDataContent caseDataContent) {
+        log.info("Filter role assignments for case event {}", caseDataContent.getEvent().getEventId());
+
+        List<RoleAssignment> roleAssignmentList = roleAssignments.getRoleAssignments().stream()
+            .filter(roleAssignment -> validateStartAndEndDate(roleAssignment))
+            .collect(Collectors.toList());
+        return RoleAssignments.builder().roleAssignments(roleAssignmentList).build();
+    }
+
+    @Override
+    public RoleAssignments filter(RoleAssignments roleAssignments, CaseTypeDefinition caseTypeDefinition) {
+        log.info("Filter role assignments for case type {}", caseTypeDefinition.getName());
+
+        List<RoleAssignment> roleAssignmentList = roleAssignments.getRoleAssignments().stream()
+            .filter(roleAssignment -> validateStartAndEndDate(roleAssignment))
+            .collect(Collectors.toList());
+        return RoleAssignments.builder().roleAssignments(roleAssignmentList).build();
+    }
+
+    private boolean validateJurisdiction(CaseDetails caseDetails, RoleAssignment roleAssignment) {
+        log.debug("Apply filter on jurisdiction {} for role assignment {}",
+            caseDetails.getJurisdiction(),
+            roleAssignment.getId());
+        roleAssignment.getMatchingResults().setValidJurisdiction(
+            roleAssignment.getAttributes().getJurisdiction()
+                .equals(caseDetails.getJurisdiction()));
+        return roleAssignment.getMatchingResults().isValidJurisdiction();
+    }
+
+    private boolean validateSecurityValidation(CaseDetails caseDetails, RoleAssignment roleAssignment) {
+        log.debug("Apply filter on security classification {} for role assignment {}",
+            roleAssignment.getClassification(),
+            roleAssignment.getId());
+        Optional<SecurityClassification> securityClassification = SecurityClassificationUtils
+            .getSecurityClassification(roleAssignment
+                .getClassification());
+        if (securityClassification.isPresent()) {
+            boolean value = caseHasClassificationEqualOrLowerThan(SecurityClassificationUtils
+                .getSecurityClassification(roleAssignment
+                    .getClassification())
+                .get())
+                .test(caseDetails);
+            roleAssignment.getMatchingResults().setValidClassification(value);
+        }
+        return roleAssignment.getMatchingResults().isValidClassification();
+    }
+
+    private boolean validateCaseId(CaseDetails caseDetails, RoleAssignment roleAssignment) {
+        log.debug("Apply filter on case id {} for role assignment {}",
+            roleAssignment.getAttributes().getCaseId(),
+            roleAssignment.getId());
+        roleAssignment.getMatchingResults().setValidCaseId(
+            roleAssignment.getAttributes().getCaseId().equals(caseDetails.getReferenceAsString()));
+        return roleAssignment.getMatchingResults().isValidCaseId();
+    }
+
+    private boolean validateLocation(RoleAssignment roleAssignment) {
+        log.debug("Apply filter on location {} for role assignment {}",
+            roleAssignment.getAttributes().getLocation(),
+            roleAssignment.getId());
+        boolean isEmpty = StringUtils.isEmpty(roleAssignment.getAttributes().getLocation());
+        if (isEmpty) {
+            roleAssignment.getMatchingResults().setValidLocation(true);
+        } else {
+            roleAssignment.getMatchingResults().setValidLocation(
+                roleAssignment.getAttributes().getLocation()
+                    .equals("" /* location */));
+        }
+        return roleAssignment.getMatchingResults().isValidLocation();
+    }
+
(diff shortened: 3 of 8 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private boolean validateJurisdiction(CaseDetails caseDetails, RoleAssignment roleAssignment) {
+        log.debug("Apply filter on jurisdiction {} for role assignment {}",
+            caseDetails.getJurisdiction(),
+            roleAssignment.getId());
+        roleAssignment.getMatchingResults().setValidJurisdiction(
+            roleAssignment.getAttributes().getJurisdiction()
+                .equals(caseDetails.getJurisdiction()));
+        return roleAssignment.getMatchingResults().isValidJurisdiction();
```
rafalkalita, line 68: So what is the rationale behind storing the information what did we matched on? It is not needed at the moment. Is this for the future?

rafalkalita (approved):
One comment. Otherwise looks good

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignments.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts.ccd.domain.service.accessprofile.filter;
+
+import uk.gov.hmcts.ccd.domain.model.casedataaccesscontrol.RoleAssignments;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseDetails;
+import uk.gov.hmcts.ccd.domain.model.definition.CaseTypeDefinition;
+import uk.gov.hmcts.ccd.domain.model.std.CaseDataContent;
+
+public interface FilterRoleAssignments {
```
mario-paniccia, line 8:
```suggestion
public interface RoleAssignmentsFilteringService {
```

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+import static uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationUtils.caseHasClassificationEqualOrLowerThan;
+
+@Slf4j
+@Component
+public class FilterRoleAssignmentsImpl implements FilterRoleAssignments, AccessControl {
+
+    @Override
+    public RoleAssignments filter(RoleAssignments roleAssignments, CaseDetails caseDetails) {
+        log.info("Filter role assignments for case {}", caseDetails.getReference());
```
mario-paniccia, line 27:
```suggestion
        log.info("Filter role assignments that apply to case {}", caseDetails.getReference());
```

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+        List<RoleAssignment> roleAssignmentList = roleAssignments.getRoleAssignments().stream()
+            .filter(roleAssignment -> validateStartAndEndDate(roleAssignment))
+            .filter(roleAssignment -> validateSecurityValidation(caseDetails, roleAssignment))
```
mario-paniccia, line 31:
```suggestion
            .filter(roleAssignment -> validateSecurityClassification(caseDetails, roleAssignment))
```

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+@Slf4j
+@Component
+public class FilterRoleAssignmentsImpl implements FilterRoleAssignments, AccessControl {
+
+    @Override
+    public RoleAssignments filter(RoleAssignments roleAssignments, CaseDetails caseDetails) {
+        log.info("Filter role assignments for case {}", caseDetails.getReference());
+
+        List<RoleAssignment> roleAssignmentList = roleAssignments.getRoleAssignments().stream()
+            .filter(roleAssignment -> validateStartAndEndDate(roleAssignment))
```
mario-paniccia, line 30: maybe `matchStartAndEndDate`, `matchSecurityClassification`....?

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+                .get())
+                .test(caseDetails);
+            roleAssignment.getMatchingResults().setValidClassification(value);
+        }
+        return roleAssignment.getMatchingResults().isValidClassification();
+    }
+
+    private boolean validateCaseId(CaseDetails caseDetails, RoleAssignment roleAssignment) {
+        log.debug("Apply filter on case id {} for role assignment {}",
+            roleAssignment.getAttributes().getCaseId(),
```
mario-paniccia, line 91: Different types of role assignments can have different lists of attributes. Organisational role assignments for example won't have a `caseId` attribute
`roleAssignment.getAttributes().getCaseId().equals(caseDetails.getReferenceAsString()))` is going to throw a NullPointerException

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+            roleAssignment.getId());
+        roleAssignment.getMatchingResults().setValidCaseId(
+            roleAssignment.getAttributes().getCaseId().equals(caseDetails.getReferenceAsString()));
```
mario-paniccia, line 94: Different types of role assignments can have different lists of attributes. Organisational role assignments for example won't have a `caseId` attribute
`roleAssignment.getAttributes().getCaseId().equals(caseDetails.getReferenceAsString()))` is going to throw a NullPointerException

src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/MatchingResults.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Data;
+
+@Data
+public class MatchingResults {
+    boolean validDate;
+    boolean validCaseId;
+    boolean validJurisdiction;
+    boolean validRegion;
+    boolean validLocation;
+    boolean validClassification;
```
mario-paniccia, line 12: Consider that different types of Role Assignements can have different lists of attributes. For example `case id` does not apply for an organisational type role assignment, only for case role type role assignement.
Wondering whether modelling this as a boolean we wont' be able to distinguish whether the attribute was present but didn't match, or was not present at all
Maybe it won't matter. Let's give it a tought

mario-paniccia: also, I'd rename to `caseIdMatched`, `jurisdictionMatched`.... since we are not doing a validation here

src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/MatchingResults.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.ccd.domain.model.casedataaccesscontrol;
+
+import lombok.Data;
+
+@Data
+public class MatchingResults {
```
mario-paniccia, line 6:
```suggestion
public class RoleMatchingResults {
```

kiran-yenigala-hmcts:
This PR is closed as we have created new branch for the access-control (attribute-based-access-control) related changes and moved this change to the new branch based on access-control branch (attribute-based-access-control)

src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/MatchingResults.java
mario-paniccia: update: confirmed that this won't work. The list of attributes is dynamic, so we need to either model them as Optional, or as a Map

mario-paniccia: update: confirmed that this won't work. The list of attributes is dynamic, so we are going to model them as Optional, or as a Map. 
Based on those changes, do we also have to apply a similar change to this class?

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+import uk.gov.hmcts.ccd.domain.model.definition.CaseTypeDefinition;
+import uk.gov.hmcts.ccd.domain.model.std.CaseDataContent;
+import uk.gov.hmcts.ccd.domain.service.AccessControl;
+import uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationUtils;
+
+import static uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationUtils.caseHasClassificationEqualOrLowerThan;
+
+@Slf4j
+@Component
+public class FilterRoleAssignmentsImpl implements FilterRoleAssignments, AccessControl {
```
mario-paniccia, line 23:
```suggestion
public class DefaultRoleAssignmentsFilteringService implements RoleAssignmentsFilteringService, AccessControl {
```

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
mario-paniccia, line 30: it would have been nice to come up with a solution where we had a sort of "general loop logic" based on interfaces, that didn't need to be updated every time we introduced a new attribute. Sort of what we do with case fields validation. We add new validator classes, but we don't need to update the validation framework logic

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        return roleAssignment.getMatchingResults().isValidRegion();
+    }
+
+    private boolean validateStartAndEndDate(RoleAssignment roleAssignment) {
+        log.debug("Apply filter on start {} and end time {} for role assignment {}",
+            roleAssignment.getBeginTime(),
+            roleAssignment.getEndTime(),
+            roleAssignment.getId());
+        LocalDateTime now = LocalDateTime.now();
```
mario-paniccia, line 133: can we inject a Clock as done here please: https://github.com/hmcts/ccd-data-store-api/blob/7463f215c020c2c8d9c607c8b856b9de5a55a362/src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/CreateCaseEventService.java#L100
it will simplify testing

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        List<RoleAssignment> roleAssignmentList = roleAssignments.getRoleAssignments().stream()
+            .filter(roleAssignment -> validateStartAndEndDate(roleAssignment))
+            .filter(roleAssignment -> validateSecurityValidation(caseDetails, roleAssignment))
+            .filter(roleAssignment -> validateCaseId(caseDetails, roleAssignment))
+            .filter(roleAssignment -> validateJurisdiction(caseDetails, roleAssignment))
+            .filter(roleAssignment -> validateLocation(roleAssignment))
+            .filter(roleAssignment -> validateRegion(roleAssignment))
+            .collect(Collectors.toList());
+
+        return RoleAssignments.builder().roleAssignments(roleAssignmentList).build();
```
mario-paniccia, line 38: perhaps simpler just to add a constructor
`return new RoleAssignments(roleAssignmentList)`
?

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+import uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationUtils;
+
+import static uk.gov.hmcts.ccd.domain.service.common.SecurityClassificationUtils.caseHasClassificationEqualOrLowerThan;
+
+@Slf4j
+@Component
+public class FilterRoleAssignmentsImpl implements FilterRoleAssignments, AccessControl {
+
+    @Override
+    public RoleAssignments filter(RoleAssignments roleAssignments, CaseDetails caseDetails) {
```
mario-paniccia, line 26: this logic is basically duplicated among the 3 overloaded methods. It would be nice if we could remove this duplication

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+            .filter(roleAssignment -> validateStartAndEndDate(roleAssignment))
+            .collect(Collectors.toList());
+        return RoleAssignments.builder().roleAssignments(roleAssignmentList).build();
+    }
+
+    private boolean validateJurisdiction(CaseDetails caseDetails, RoleAssignment roleAssignment) {
+        log.debug("Apply filter on jurisdiction {} for role assignment {}",
+            caseDetails.getJurisdiction(),
+            roleAssignment.getId());
+        roleAssignment.getMatchingResults().setValidJurisdiction(
```
mario-paniccia, line 65: I wouldn't add the matching results inside the RoleAssignment. Don't see the reason. I would have them as a separate entities returned by the filtering service.
i.e. the `filter` methods return a `RoleAssignmentFilteringResult` object, which contains the new list of RoleAssigment, and a list of `RoleMatchingResult`s.
One benefit is that in the future if needed we can add additional properties on this `RoleAssignmentFilteringResult`

src/test/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImplTest.java
```diff
@@ -0,0 +1,170 @@
+package uk.gov.hmcts.ccd.domain.service.accessprofile.filter;
+
```
mario-paniccia, line 2: we need more testing for the filtering stage. It's something well defined where I don't expect changes, so we can add tests now rather than later on

mario-paniccia (commented):
some comments

src/main/java/uk/gov/hmcts/ccd/domain/service/accessprofile/filter/FilterRoleAssignmentsImpl.java
kiran-yenigala-hmcts: will discuss this with Rafal

rafalkalita: https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html
Looks like there is nothing about the format of the Clock when comes to serialization. We need ISO-8601. Digging further looks like Jackson jsr-310 extension does not support it:
https://fasterxml.github.io/jackson-modules-java8/javadoc/datetime/2.12/com/fasterxml/jackson/datatype/jsr310/JavaTimeModule.html
https://stackoverflow.com/questions/27952472/serialize-deserialize-java-8-java-time-with-jackson-json-mapper

We could provide a @Qualifier("utcLocalDateTime") the same way you are suggesting a Clock. Or we can use the Clock and pass it to the LocalDateTime, because all now() overrides use a Clock at the end.

// LocalDateTime.java
    public static LocalDateTime now(Clock clock) {
        Objects.requireNonNull(clock, "clock");
        final Instant now = clock.instant();  // called once
        ZoneOffset offset = clock.getZone().getRules().getOffset(now);
        return ofEpochSecond(now.getEpochSecond(), now.getNano(), offset);
    }

I am afraid LocalDateTime.now(); will get the default zone in the LocalDateTime.java
    public static LocalDateTime now() {
        return now(Clock.systemDefaultZone());
    }
    but we could use
    public static LocalDateTime now(ZoneId zone) {
        return now(Clock.system(zone));
    }
    
    Because we know the roleAssignments are in UTC I would use:
    LocalDateTime now = LocalDateTime.now(Clock.systemUTC());
    
    https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html#systemUTC--

PR #1382, opened 2021-02-23
Rdm 10937

Thor-tech-of-metal:
RDM-10937: Change [CASE REFERENCE] metadata field type from 'Number' to 'Text'
               [RDM-10937] (https://tools.hmcts.net/jira/browse/RDM-10937).


Jenkinsfile_CNP
```diff
@@ -116,7 +116,7 @@ withPipeline(type, product, component) {
             environmentOfDependencies = "aat"
         }
         env.DEFINITION_STORE_HOST      = "http://ccd-definition-store-api-${environmentOfDependencies}.service.core-compute-${environmentOfDependencies}.internal"
-        env.ELASTIC_SEARCH_FTA_ENABLED = "true"
```
Thor-tech-of-metal, line 119: Elastic search befta has been enable for prs,  in dev branch.  The correct values has to be fixed in another pr, or this one . ☝️

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -11,4 +13,20 @@ public static String formatCaseReference(String caseReference) {
         }
         return caseReference;
     }
+
+    public static boolean checkRegex(final String value) {
+        final String regularExpression = "(?:^[0-9]{16}$|^\\d{4}-\\d{4}-\\d{4}-\\d{4}$)";
+        return value.matches(regularExpression);
+    }
+
+    public static Optional<String> getFormatCaseReference(Optional<String> caseReference) {
```
mario-paniccia, line 22: why you passing the Optional? in terms of API it's better if you pass the String, and it's the code that calls the that deals with the Optional.
Otherwise every code that calls this method will need to wrap the string in an Optional

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -11,4 +13,20 @@ public static String formatCaseReference(String caseReference) {
         }
         return caseReference;
     }
+
+    public static boolean checkRegex(final String value) {
+        final String regularExpression = "(?:^[0-9]{16}$|^\\d{4}-\\d{4}-\\d{4}-\\d{4}$)";
```
mario-paniccia, line 18: can we have this as a constant as a class field please. 
Also, can you give it a better name that helps understand what the regex is ?

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -11,4 +13,20 @@ public static String formatCaseReference(String caseReference) {
...
+
+    public static Optional<String> getFormatCaseReference(Optional<String> caseReference) {
+
+        if (caseReference.isPresent()) {
+            if (!CaseReferenceUtils.checkRegex(caseReference.get())) {
+                return Optional.of("0000000000000000");
+            }
+            return Optional.of(CaseReferenceUtils.formatCaseReference(caseReference.get()));
+        }
+        return caseReference;
```
mario-paniccia, line 30: I think all this logic cold be done by the calling class and this method be completely removed

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -11,4 +13,20 @@ public static String formatCaseReference(String caseReference) {
...
+        return value.matches(regularExpression);
+    }
+
+    public static Optional<String> getFormatCaseReference(Optional<String> caseReference) {
+
+        if (caseReference.isPresent()) {
+            if (!CaseReferenceUtils.checkRegex(caseReference.get())) {
+                return Optional.of("0000000000000000");
+            }
+            return Optional.of(CaseReferenceUtils.formatCaseReference(caseReference.get()));
```
mario-paniccia, line 28: can you rename `formatCaseReference` to something more understandable? like for example: `removeHyphens`?

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -11,4 +13,20 @@ public static String formatCaseReference(String caseReference) {
...
+    public static boolean checkRegex(final String value) {
+        final String regularExpression = "(?:^[0-9]{16}$|^\\d{4}-\\d{4}-\\d{4}-\\d{4}$)";
+        return value.matches(regularExpression);
+    }
+
+    public static Optional<String> getFormatCaseReference(Optional<String> caseReference) {
+
+        if (caseReference.isPresent()) {
+            if (!CaseReferenceUtils.checkRegex(caseReference.get())) {
+                return Optional.of("0000000000000000");
```
mario-paniccia, line 26: is this safe?

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -1,5 +1,7 @@
 package uk.gov.hmcts.ccd.domain.model.common;
 
+import java.util.Optional;
+
 public class CaseReferenceUtils {
```
mario-paniccia, line 5: some unit tests for this class would be useful

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
```diff
@@ -120,7 +120,7 @@ private SearchResultViewItem buildSearchResultViewItem(final CaseDetails caseDet
                                                            final CaseTypeDefinition caseTypeDefinition,
                                                            final SearchResultDefinition searchResult) {
         Map<String, JsonNode> caseData = new HashMap<>(caseDetails.getData());
-        Map<String, Object> caseMetadata = new HashMap<>(caseDetails.getMetadata());
+        Map<String, Object> caseMetadata = new HashMap<>(caseDetails.getFormattedMetadata());
```
rafalkalita, line 123: Change not reflected in unit test.

rafalkalita (commented):
Few comments

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
```diff
@@ -330,6 +330,21 @@ private void setAfterSubmitCallbackResponseEntity(final AfterSubmitCallbackRespo
         return metadata;
     }
 
+    @JsonIgnore
+    public Map<String, Object> getFormattedMetadata() {
```
kiran-yenigala-hmcts, line 334: do we need completely new method for this, can't we change this in getMetaData method

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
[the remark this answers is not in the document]
Thor-tech-of-metal: haha good point , we want the implementation to be like that ,  this represent the null in case_reference value. :)

mario-paniccia: I also wonder whether there's a better way to do this, so someone searches for a non valid reference, we don't execute the query and immediately return empty results. Your solution executes the query.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
mario-paniccia, line 334: probably not the best name. You are not formatting here. You are changing the type of one of the map entries

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
```diff
@@ -330,6 +330,21 @@ private void setAfterSubmitCallbackResponseEntity(final AfterSubmitCallbackRespo
...
+        if (metadata.isEmpty()) {
+            metadata.put(JURISDICTION.getReference(), getJurisdiction());
+            metadata.put(CASE_TYPE.getReference(), getCaseTypeId());
+            metadata.put(STATE.getReference(), getState());
+            metadata.put(CASE_REFERENCE.getReference(), getReference() != null ? getReference().toString() : getId());
```
mario-paniccia, line 339: for my understanding, can you explain where the code breaks if you don't do this conversion? maybe we can find a better way to do this

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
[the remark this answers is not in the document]
mario-paniccia: we don't want an error @rafalkalita , what we want is that when the user searches with an invalid case reference, we want to just return no results

mario-paniccia: I think it would be better to add case reference validation here:
https://github.com/hmcts/ccd-data-store-api/blob/80adc76067063ba3c3600fb3e0674b41bfe5426f/src/main/java/uk/gov/hmcts/ccd/domain/service/search/DefaultSearchOperation.java#L26

so that if the user is searching by case reference and is an invalid one, we just return empty results without executing the query on the DB

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
kiran-yenigala-hmcts: When getFormattedMetaData is called before getMetaData then it will populate metadata map already.
it may break other functionalities.

if you are confident that functionality is not going to break, then in both methods, only below line is different, 
            metadata.put(CASE_REFERENCE.getReference(), getReference() != null ? getReference() : getId());
            
            Which can be handled by boolean parameter, Think about this

kiran-yenigala-hmcts (commented):
couple of comments

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
Thor-tech-of-metal: Toolkit SearchResultViewItemComparatorFactory comparators. Browser errors

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
Thor-tech-of-metal: What would be recommended name ?

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/MergeDataToSearchResultOperation.java
Thor-tech-of-metal: There is not test in caseDetails  but I added a test  to evaluate the case reference type in MergeDataToSearchResultOperation.java

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseDetails.java
Thor-tech-of-metal: I ended up using getConvertedMetadata

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
Thor-tech-of-metal: That one it is important !

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -2,13 +2,19 @@
 
 public class CaseReferenceUtils {
 
+    public static final String CASE_REFERENCE_EXPRESSION = "(?:^[0-9]{16}$|^\\d{4}-\\d{4}-\\d{4}-\\d{4}$)";
```
mario-paniccia, line 5:
```suggestion
    public static final String CASE_REFERENCE_REGEX = "(?:^[0-9]{16}$|^\\d{4}-\\d{4}-\\d{4}-\\d{4}$)";
```

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
```diff
@@ -2,13 +2,19 @@
...
 
-    public static String formatCaseReference(String caseReference) {
+    public static String removeHyphens(String caseReference) {
         if (caseReference.contains("-")) {
             return String.join("", caseReference.split("-"));
         }
         return caseReference;
     }
+
+    public static boolean isAValidCaseReferenceFormat(final String value) {
```
mario-paniccia, line 17:
```suggestion
    public static boolean isAValidCaseReference(final String reference) {
```

src/main/java/uk/gov/hmcts/ccd/domain/service/search/PaginatedSearchMetaDataOperation.java
```diff
@@ -21,6 +20,10 @@ public PaginatedSearchMetaDataOperation(@Qualifier(CachedCaseDetailsRepository.Q
     }
 
     public PaginatedSearchMetadata execute(MetaData metaData, Map<String, String> criteria) {
+
+        if (!metaData.validateAndsConvertReference()) {
```
mario-paniccia, line 24:
```suggestion
        if (!metaData.validateAndConvertReference()) {
```

src/main/java/uk/gov/hmcts/ccd/domain/service/search/PaginatedSearchMetaDataOperation.java
```diff
@@ -21,6 +20,10 @@ public PaginatedSearchMetaDataOperation(@Qualifier(CachedCaseDetailsRepository.Q
...
+            return new PaginatedSearchMetadata();
```
mario-paniccia, line 25:
```suggestion
            return PaginatedSearchMetadata.EMPTY;
```
where EMPTY is a constant in PaginatedSearchMetadata:
`public static final EMPTY = new PaginatedSearchMetadata();`

mario-paniccia (approved):
approved with some minor comments

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
Thor-tech-of-metal: DOne

src/main/java/uk/gov/hmcts/ccd/domain/model/common/CaseReferenceUtils.java
Thor-tech-of-metal: DONE

src/main/java/uk/gov/hmcts/ccd/domain/service/search/PaginatedSearchMetaDataOperation.java
Thor-tech-of-metal: DONE

src/main/java/uk/gov/hmcts/ccd/domain/service/search/PaginatedSearchMetaDataOperation.java
Thor-tech-of-metal: DONE

kiran-yenigala-hmcts (approved):

rafalkalita (approved):

PR #1384, opened 2021-02-25
RDM-11029: Alter the approach to checking work basket result field roles

tomxelliott:

https://tools.hmcts.net/jira/browse/RDM-11029

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultDefinition.java
```diff
@@ -28,25 +29,34 @@ public void setFields(SearchResultField[] fields) {
             .collect(Collectors.toList());
     }
 
-    public Map<String, String> getFieldsUserRoles() {
-        Map<String, String> fields = new HashMap<>();
+    public Map<String, List<String>> getFieldsUserRoles() {
+        Map<String, List<String>> fields = new HashMap<>();
```
sullivangareth, line 33: Would it be worth naming this differently so that you don't need to use this.fields to differentiate from the other instance ?

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultDefinition.java
```diff
@@ -28,25 +29,34 @@ public void setFields(SearchResultField[] fields) {
...
         for (SearchResultField srf : this.fields) {
-            fields.put(srf.getCaseFieldId(), srf.getRole());
+            if (fields.containsKey(srf.getCaseFieldId())) {
+                fields.get(srf.getCaseFieldId()).add(srf.getRole());
+            } else {
+                List<String> roles = new ArrayList<>();
+                roles.add(srf.getRole());
+                fields.put(srf.getCaseFieldId(), roles);
```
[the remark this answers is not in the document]
tomxelliott: I was really keen to use that method of List.of(srf.getRole()). However, these methods create an immutable list which means that when I subsequently try to add values, I get an UnsupportedOperationException.

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultDefinition.java
tomxelliott: Will refactor this now. Thanks for the review

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultDefinition.java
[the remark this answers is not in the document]
sullivangareth: ah right, of course - fair enough then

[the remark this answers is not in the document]
tomxelliott: Will try Arrays.asList() as I believe that creates a mutable list

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/SearchResultDefinition.java
```diff
@@ -28,25 +30,32 @@ public void setFields(SearchResultField[] fields) {
...
+        Map<String, List<String>> fields = getFieldsUserRoles();
+        List<String> userRoles = fields.get(caseFieldId);
+        if (userRoles != null) {
+            long count = userRoles.stream()
+                .filter(userRole -> userRole == null || roles.contains(userRole))
+                .count();
+            return count != 0;
         }
-        return true;
+        return false;
```
mario-paniccia, line 59: there is change in behaviour here when the field has no user roles. In the previous version of the code `fieldHasRole` would return `true`, now it returns `false`. Is that what correct?

tomxelliott: Not quite. Before we would return false if there was a non null role that was not contained in the roles set - we would return true for null roles and those that were contained in the roles set.

With my change, we return false if there is no mapping to the caseFieldId (therefore no roles not even a null one) or if there is a mapping to the caseFieldId but none of the roles in that mapping exist in the roles set or none of them are null (the count will therefore be 0 and the condition will be false and return false).

mario-paniccia (approved):
👍

PR #1416, opened 2021-03-18
Create dummy.yml

nathan-clark:
RDM-11603

```
[ ] Yes
[x] No
```

### DIFF
--- a/.github/workflows/dummy.yml
+++ b/.github/workflows/dummy.yml
@@ -1,4 +1,4 @@
-name: Add a Dummy commit to trigger build
+name: Add a Dummy commit to trigger dev build
 
 on:
   push:
@@ -6,7 +6,7 @@ on:
     paths-ignore:
       - .github/workflows/dummy.yml
   schedule:
-    - cron: "0 0 * * *"
+    - cron: "30 7 * * *"
   workflow_dispatch:
   
 jobs:
@@ -16,15 +16,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: develop
-      - name: Modify Dummy file
-        run: |
-          date=`date '+%Y-%m-%dT%H:%M:%SZ'`
-          echo $date > Dummy_File
       
-      - name: Commit Dummy_File and push to develop branch
+      - name: Push Dummy Commit to develop branch
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
-          commit_message: 'Update Dummy_File'
+          skip_dirty_check: true
+          commit_message: "Trigger build"
+          commit_options: "--allow-empty"
           branch: 'develop'
-          file_pattern: Dummy_File
-          commit_author: Dummy-Commit <actions@github.com>
--- a/.github/workflows/dummy.yml
+++ b/.github/workflows/dummy.yml
@@ -1,0 +1,30 @@
+name: Add a Dummy commit to trigger build
+
+on:
+  push:
+    branches: master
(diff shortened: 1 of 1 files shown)


.github/workflows/dummy.yml
```diff
@@ -0,0 +1,30 @@
+name: Add a Dummy commit to trigger build
+
+on:
+  push:
+    branches: master
+    paths-ignore:
+      - .github/workflows/dummy.yml
+  schedule:
+    - cron: "0 0 * * *"
```
MSancaktutar, line 9: Perhaps better at 07:30am every day, just so that if builds fail we start the day by fixing the issue first?

.github/workflows/dummy.yml
```diff
@@ -0,0 +1,30 @@
+name: Add a Dummy commit to trigger build
+
+on:
+  push:
+    branches: master
```
MSancaktutar, line 5: change this temporarily to this branch name and schedule a dummy commit frequency for test purpose. You can push the dummy commits to a dummy branch for testing.

.github/workflows/dummy.yml
```diff
@@ -0,0 +1,30 @@
...
+    paths-ignore:
+      - .github/workflows/dummy.yml
+  schedule:
```
MSancaktutar, line 8: This workflow will be executed upon every merge into master. Would that re-schedule the dummy commits? What is the effect? Will we be scheduling many redundantly scheduled ones?

.github/workflows/dummy.yml
nathan-clark: it only works when workflow is committed to master

.github/workflows/dummy.yml
nathan-clark: this just ignores push on workflow def file

.github/workflows/dummy.yml
nathan-clark: ok

.github/workflows/dummy.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: develop
+      - name: Modify Dummy file
+        run: |
+          date=`date '+%Y-%m-%dT%H:%M:%SZ'`
+          echo $date > Dummy_File
+      
+      - name: Commit Dummy_File and push to develop branch
```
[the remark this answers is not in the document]
nathan-clark: np

.github/workflows/dummy.yml
MSancaktutar: So, would it reschedule a new cron every time?

nathan-clark: only on schedule, removed the push path as was having conflicts

MSancaktutar (approved):
LGTM

.github/workflows/dummy.yml
```diff
@@ -0,0 +1,22 @@
+name: Add a Dummy commit to trigger dev build
```
smathangi, line 1: this shows as job name under "Actions" . Can you change to "Dummy commit Scheduler" or something more meaningful please.

MSancaktutar: Well, the action this is doing is to push a dummy commit. Scheduled execution is it's feature. So, perhaps better name could be like "Scheduled Dummy Commit to develop Branch". (which can actually be triggered manually as well, though.)

PR #1427, opened 2021-03-30
RDM-11850: [CASE REFERENCE] metadata field type still returned as 'Number' and not as a String via ES api

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-11850

### Change description ###

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultViewGenerator.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultViewGenerator.java
@@ -224,8 +224,10 @@ public class CaseSearchResultViewGenerator {
     }
 
     private Map<String, Object> convertReferenceToString(Map<String, Object> caseMetadata) {
-        final String convertedCaseReference = String.valueOf(caseMetadata.get(CASE_REFERENCE.getReference()));
-        caseMetadata.put(CASE_REFERENCE.getReference(), convertedCaseReference);
-        return caseMetadata;
+        Map<String, Object> caseMetaDataMap = new HashMap<>();
+        caseMetaDataMap.putAll(caseMetadata);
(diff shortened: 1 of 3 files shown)


danlysiak (approved):

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultViewGenerator.java
```diff
@@ -220,7 +219,13 @@ private SearchResultViewItem buildSearchResultViewItem(CaseDetails caseDetails,
...
 
         newResults.putAll(caseData);
-        newResults.putAll(metadata);
+        newResults.putAll(convertReferenceToString(metadata));
         return newResults;
     }
+
+    private Map<String, Object> convertReferenceToString(Map<String, Object> caseMetadata) {
+        final String convertedCaseReference = String.valueOf(caseMetadata.get(CASE_REFERENCE.getReference()));
+        caseMetadata.put(CASE_REFERENCE.getReference(), convertedCaseReference);
```
mario-paniccia, line 228: a minor, not sure if it could cause any issue without a deeper investigation:
you are modifying the original map in CaseDetails here. Perhaps safer to make a copy of the map `caseMetadata` and return the copy with the modified case reference.
Or, not modify the Map, but directly modifying `newResults` in the caller.

mario-paniccia (commented):
one minor comment

src/main/java/uk/gov/hmcts/ccd/domain/service/search/CaseSearchResultViewGenerator.java
kiran-yenigala-hmcts: created copy of the map

mario-paniccia (approved):
👍

PR #1429, opened 2021-03-30
RDM-11771

rajaram-nagarajan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/RDM-11771

### Change description ###

Added display_context for case event complex field definition

```
[ ] Yes
[x] No
```


rajaram-nagarajan:
CCD Backend will only validate the data (min, max, regex etc) and is not meant to be validating for mandatory field. This is a new enhancement requirement which would be addressed in a separate ticket.

PR #1431, opened 2021-03-31
RDM-9354 Case Sensitive CaseRole on Add Case-Assigned Users and Roles endpoint

smathangi:

https://tools.hmcts.net/jira/browse/RDM-9354

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/caseaccess/CaseAccessOperation.java
@@ -112,8 +112,7 @@ public class CaseAccessOperation {
 
         cauRolesByCaseId.forEach((caseId, requestedAssignments) ->
             requestedAssignments.forEach(requestedAssignment ->
-                caseUserRepository.grantAccess(caseId, requestedAssignment.getUserId(),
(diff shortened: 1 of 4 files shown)


rajaram-nagarajan (approved):
LGTM

MSancaktutar (approved):
Looks good, but could we also double check there is no duplicated persistence logic for this to applied to further, around the case_users table?

smathangi:
> Looks good, but could we also double check there is no duplicated persistence logic for this to applied to further, around the case_users table?

at the moment grant api is passing role implicitly (CREATOR).. so no issues there. We could even push this check into repository level so that we can guard for all future usages. But repository should throw a validation error rather than suppressing silently. (callers can suppress if they wish to..). Please let me know your thoughts @MSancaktutar

MSancaktutar:
> > Looks good, but could we also double check there is no duplicated persistence logic for this to applied to further, around the case_users table?
> 
> at the moment grant api is passing role implicitly (CREATOR).. so no issues there. We could even push this check into repository level so that we can guard for all future usages. But repository should throw a validation error rather than suppressing silently. (callers can suppress if they wish to..). Please let me know your thoughts @MSancaktutar

Such a change will go too much beyond the purpose of this change. So, this improvement should now be sufficient.

PR #1450, opened 2021-04-15
RDM-11956 Extend allowed list of document domains to include em-hrs-api

skandala-cgi:
https://tools.hmcts.net/jira/browse/RDM-11956

### Change description ###
Extend allowed list of document domains to include em-hrs-api

```
[ ] Yes
[X] No
```


MSancaktutar (changes requested):
Let's remove CCD_DM_DOMAIN and manage the entire document with just 1 env var so that it contains a regular expression to accept both and only DM Store and HRS documents.

PR #1479, opened 2021-05-13
RDM-10800-Update the V2 internal APIs that return Access metadata information for Read and Update/Delete operations

sullivangareth:

https://tools.hmcts.net/jira/browse/RDM-10800

### Change description ###

Additional information must be provided to ExUI so that ExUI can inform the user of the type of access that they have to the case and/or the process that they should go through to get access to that case.  To support this, CCD will return two metadata fields:

access_granted, which is a CSV of all of the grants that passed the role filtering process.  E.g. STANDARD, SPECIFIC
access_process, which indicates the process to be followed to gain access to the case.  One of CHALLENGED, SPECIFIC, NONE
This information will be included when returning case information to ExUI for any purpose including search (via the internal V2 APIs only)

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/CaseAccessMetadata.java
```diff
@@ -0,0 +1,15 @@
...
+
+@Data
+public class CaseAccessMetadata {
+
+    public static final String ACCESS_GRANTED = "[ACCESS_GRANTED]";
+    public static final String ACCESS_PROCESS = "[ACCESS_PROCESS]";
+
+    public static final String ACCESS_GRANTED_LABEL = "Access Granted";
+
+    private String accessGrants;
```
smathangi, line 13: Can we use enums as they are available already? 
how about keep the property as `private List<GrantType> accessGrants` and expose a additional method for the comma-separated string String getAccessGrantsString() ?

src/main/java/uk/gov/hmcts/ccd/domain/service/aggregated/AuthorisedGetCaseViewOperation.java
```diff
@@ -110,4 +118,21 @@ private CaseView filterUpsertAccess(String caseReference,
...
+        CaseAccessMetadata caseAccessMetadata = caseDataAccessControl.generateAccessMetadata(caseView.getCaseId());
+
+        CaseViewField caseViewField = new CaseViewField();
+        caseViewField.setId(ACCESS_GRANTED);
+        caseViewField.setLabel(ACCESS_GRANTED_LABEL);
+        caseViewField.setSecurityLabel(SecurityClassification.PUBLIC.name());
+        caseViewField.setValue(caseAccessMetadata.getAccessGrants());
+
+        metadataFields.add(caseViewField);
+        caseView.setMetadataFields(metadataFields);
```
smathangi, line 135: Calling `addMetadataField(..)` is better ? 
It looks like it doesn't exist. Do you mind to add that method please?

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/AbstractCaseView.java
```diff
@@ -46,4 +47,11 @@ public void setTabs(CaseViewTab[] tabs) {
     public void setMetadataFields(List<CaseViewField> metadataFields) {
         this.metadataFields = metadataFields;
     }
+
+    public void addMetadataFields(List<CaseViewField> metadataFields) {
```
smathangi, line 51: May be with single item rather than list? 
void addMetadataFields(CaseViewField metadataField)

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/CaseUpdateViewEvent.java
```diff
@@ -25,6 +28,13 @@
     private String endButtonLabel;
     @JsonProperty("can_save_draft")
     private Boolean canSaveDraft;
+    @JsonInclude(NON_NULL)
```
smathangi, line 31: 👍

smathangi (commented):
LGTM.  (Will approve once IT tests are added)

src/test/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/CaseAccessMetadataTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+class CaseAccessMetadataTest {
+
+    private CaseAccessMetadata caseAccessMetadata = new CaseAccessMetadata();
+
+    @Test
+    void testGetAccessGrantsString() {
+        caseAccessMetadata.setAccessGrants(List.of(CHALLENGED, BASIC, STANDARD, EXCLUDED));
+
+        String expectedSortedValue = BASIC + "," + CHALLENGED + "," + EXCLUDED + "," + STANDARD;
```
KleoG, line 21: Could add a comma delimiter final field if used frequently here to join these up

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/AbstractCaseView.java
KleoG: Could probably do what Sateesh suggested, but instead of replacing this method, add a new one to support adding a singe CaseViewField to the list, as the current one you have could also be useful

KleoG (approved):
Left a few comments - nothing major from me, so approved

src/main/java/uk/gov/hmcts/ccd/domain/model/aggregated/AbstractCaseView.java
sullivangareth: 👍

src/test/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/CaseAccessMetadataTest.java
sullivangareth: 👍

sullivangareth: Fixed in https://github.com/hmcts/ccd-data-store-api/pull/1479/commits/528c646b730946ed64922e15dd898b8b992e5f6d

smathangi (approved):
LGTM 👍

PR #1541, opened 2021-06-23
CCD-913: Refactor caching for CaseDefinitionRepository

rajaram-nagarajan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CCD-913

### Change description ###

Refactored caching for CaseDefinitionRepository

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
```diff
@@ -90,7 +103,13 @@ public CaseTypeDefinitionVersion getLatestVersion(String caseTypeId) {
...
     public JurisdictionDefinition getJurisdiction(String jurisdictionId) {
         LOGGER.debug("Will get jurisdiction '{}' from repository.", jurisdictionId);
-        return caseDefinitionRepository.getJurisdiction(jurisdictionId);
+        return jurisdictions.computeIfAbsent(jurisdictionId, caseDefinitionRepository::getJurisdiction);
+    }
+
+    @Override
+    public List<JurisdictionDefinition> getJurisdictions(List<String> jurisdictionIds) {
+        return jurisdictionsList.computeIfAbsent(prepareJurisdictionsKey(jurisdictionIds),
+            e -> caseDefinitionRepository.getJurisdictions(jurisdictionIds));
```
MSancaktutar, line 112: This will fetch every list of jurisdictions separately even if they contain individual jurisdictions that were already fetched and cached. Not only performance will suffer, but also consistency of cached jurisdiction objects will be impacted. So, I believe we should maintain a single source of jurisdiction collection as a cache and serve any individual or list of jurisdictions from there.

MSancaktutar (changes requested):
One comment to be addressed, and also, could we please add more debug logs to see the cache hits and misses?

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
```diff
@@ -49,12 +56,18 @@ public CachedCaseDefinitionRepository(@Qualifier(DefaultCaseDefinitionRepository
...
-        return caseDefinitionRepository.getCaseType(latestVersion.getVersion(), caseTypeId);
+        return this.getCaseType(latestVersion.getVersion(), caseTypeId);
     }
 
     @Override
     public CaseTypeDefinition getCaseType(int version, String caseTypeId) {
-        return caseDefinitionRepository.getCaseType(version, caseTypeId);
+        try {
+            return caseTypes.computeIfAbsent(caseTypeId + version,
+                e -> caseDefinitionRepository.getCaseType(version, caseTypeId)).shallowClone();
```
smathangi, line 66: any particular reason for clone?

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
```diff
@@ -87,7 +86,6 @@ public DefaultCaseDefinitionRepository(final ApplicationParams applicationParams
     }
 
     @Override
-    @Cacheable("caseTypeDefinitionsCache")
     public CaseTypeDefinition getCaseType(int version, String caseTypeId) {
```
smathangi, line 89: what's the use of version here? if we remove cache annotation?

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
smathangi: where do we use this method actually? since it's a request scope cache i don't think we actually gain anything unless same request flow calls this repeatedly. If we really want to cache then please follow Mutlu's suggestion of single source.

rajaram-nagarajan: Sure, I will make the respective changes

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
rajaram-nagarajan: Yes, the caller is mutating the cached object. I thought to clone the object as we don't need to worry about anyone changing the cached object in the future.

src/main/java/uk/gov/hmcts/ccd/data/definition/DefaultCaseDefinitionRepository.java
rajaram-nagarajan: We actually use the version in the cached class, this is just a overridden method

rajaram-nagarajan:
Outdated

PR #1593, opened 2021-07-27
RDM-12701: Remove Embedded Postgres

sullivangareth:

https://tools.hmcts.net/jira/browse/RDM-12701

### Change description ###

Removing the embedded postgres from CCD-Data-Store

```
[ ] Yes
[X] No
```


src/test/java/uk/gov/hmcts/ccd/endpoint/std/CallbackTest.java
```diff
@@ -1405,7 +1403,6 @@ public void shouldReturn404WhenPostCreateEventWithInvalidEventTokenForCitizen()
 
     @Test
     @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = {"classpath:sql/insert_callback_cases.sql"})
-    @Transactional
```
smathangi, line 1408: are we sure @Transactional is not required here?

smathangi (commented):
Just one comment otherwise LGTM.

src/test/java/uk/gov/hmcts/ccd/endpoint/std/CallbackTest.java
sullivangareth: Not completely sure, but with it left in the test consistently fails with the following postgres error

`ERROR: canceling statement due to user request`

I have tried playing with timeouts, but without much luck

smathangi: Ok, if the test passes repeatedly without that annotation then we're good.

smathangi (approved):

src/test/resources/integration_tests.properties
```diff
@@ -2,3 +2,6 @@
 idam.api.url=http://localhost:${wiremock.server.port}/idam
 
 idam.s2s-auth.url=http://localhost:${wiremock.server.port}/s2s
+
```
mattnayler, line 5: Is this file needed now that `src/test/java/uk/gov/hmcts/ccd/integrations/IntegrationTest.java` has been removed?

mattnayler (approved):
Just one question otherwise LGTM 👍 .

src/test/resources/integration_tests.properties
sullivangareth: Good point, will check and remove accordingly

sullivangareth: Fixed in b357d60

build.gradle
```diff
@@ -214,13 +214,13 @@ dependencies {
 
     // FIXME 0.6 doesn't support jsonb; 0.7 doesn't work on Windows
 //    runtime group: 'com.impossibl.pgjdbc-ng', name: 'pgjdbc-ng', version: '0.6'
-    runtime group: 'org.postgresql', name: 'postgresql', version: '42.2.16'
+    runtime group: 'org.postgresql', name: 'postgresql', version: '42.2.23'
```
smathangi, line 217: if this bump is not required for test container, could you remove please. We can do that in another PR. So that we can release this PR as test only changes.

sullivangareth: 👍

sullivangareth: Fixed in 640c299

PR #1625, opened 2021-08-16
CCD-913: Add request level caching for jurisdiction cache

rajaram-nagarajan:
### JIRA link  ###

https://tools.hmcts.net/jira/browse/CCD-913

### Change description ###

Added request level caching for jurisdiction cache and fixed other unit/integration test issues with caching

```
[ ] Yes
[x] No
```


MSancaktutar (approved):
LGTM

src/main/java/uk/gov/hmcts/ccd/data/definition/CachedCaseDefinitionRepository.java
```diff
@@ -109,7 +110,7 @@ public CaseTypeDefinitionVersion getLatestVersion(String caseTypeId) {
     @Override
     public JurisdictionDefinition getJurisdiction(String jurisdictionId) {
         LOGGER.debug("Will get jurisdiction '{}' from repository.", jurisdictionId);
-        return caseDefinitionRepository.getJurisdiction(jurisdictionId);
+        return jurisdictions.computeIfAbsent(jurisdictionId, caseDefinitionRepository::getJurisdiction);
```
MSancaktutar, line 113: @rajaram-nagarajan - after the prod incident in trying to cache the case type objects, should we think again about this change here? Could you please do a careful check on where the `JurisdictionDefinition` objects returned from this method are used, and how they are used? Essentially, are they used to assume a separate instance returned everytime for modification within request scope?

rajaram-nagarajan: > @rajaram-nagarajan - after the prod incident in trying to cache the case type objects, should we think again about this change here? Could you please do a careful check on where the `JurisdictionDefinition` objects returned from this method are used, and how they are used? Essentially, are they used to assume a separate instance returned everytime for modification within request scope?

@MSancaktutar - I can confirm that there is no modification on this returned cached objects. We are good here.

PR #1675, opened 2021-09-29
Cfts 1020 master

skandala-cgi:
https://tools.hmcts.net/jira/browse/CFTS-1020

### Change description ###
Respondent Sols not seeing their unassigned case list

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
@@ -245,19 +245,26 @@ public class AccessControlService {
                                                     final Predicate<AccessControlList> access,
                                                     final boolean isClassification,
                                                     final CaseFieldDefinition childField) {
+        LOG.info("Child Field "+ childField);
+
         if (caseField.isCollectionFieldType() && jsonNode.isArray()) {
             jsonNode.forEach(caseFieldValueJsonNode -> {
                 if (caseFieldValueJsonNode.get(VALUE).get(childField.getId()) != null) {
                     filterChildrenUsingJsonNode(childField, caseFieldValueJsonNode.get(VALUE).get(childField.getId()),
-                                                userRoles, access, isClassification);
+                        userRoles, access, isClassification);
                 }
             });
         } else {
-            filterChildrenUsingJsonNode(childField, jsonNode.path(childField.getId()), userRoles, access,
-                                        isClassification);
+            if (jsonNode.path(childField.getId()) != null) {
+                filterChildrenUsingJsonNode(childField, jsonNode.path(childField.getId()), userRoles, access,
+                    isClassification);
+            }
+            else {
+                LOG.info(
+                    "Can not found Id value for  childField {} with path {} with user roles={}", childField, jsonNode.path(childField.getId()), userRoles);
+            }
         }
     }
-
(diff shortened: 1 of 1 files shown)


MSancaktutar (changes requested):
I think it would be better to apply this guard inside the `filterChildrenUsingJsonNode` method, as there are multiple places calling the method.

MSancaktutar (approved):
LGTM, thank you. We may need some additional tests, perhaps at unit level, but happy for it to go for demo testing, and then we can add additional tests in the meantime.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -221,7 +222,15 @@ private JsonNode filterChildrenUsingJsonNode(final CaseFieldDefinition caseField
...
                                                  final Predicate<AccessControlList> access,
                                                  boolean isClassification) {
-        if (caseField.isCompoundFieldType()) {
+        // We are getting class com.fasterxml.jackson.databind.node.MissingNode cannot be cast to class
+        // com.fasterxml.jackson.databind.node.ObjectNode exception when childField.getId() is null, so this
+        // is added to check if the value exists.
+        if (jsonNode instanceof MissingNode) {
+            LOG.info("Can not found value for  jsonNode={}, jsonNodeParent={}, user roles={}, caseType={}, "
+                    + "accessControlList={}",
+                 jsonNode, jsonNode.findParent("id"), userRoles, caseField.getCaseTypeId(),
```
smathangi, line 231: jsonNode.findParent("id") always return a null for MissingNode. Is there any value to log this?

MSancaktutar: We're thinking to replace it with the full path to the missing node. That part will follow later subject to other priorities and in the worst case, we can remove parent log as we already have enough on this log line combined with other telemetries.

smathangi (approved):

MSancaktutar (approved):
Much nicer, thank you!

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlService.java
```diff
@@ -283,7 +283,17 @@ private void locateAndRemoveChildNode(final CaseFieldDefinition caseField, final
...
-            ((ObjectNode) jsonNode).remove(childField.getId());
+            // We are getting class com.fasterxml.jackson.databind.node.MissingNode cannot be cast to class
+            // com.fasterxml.jackson.databind.node.ObjectNode exception when childField.getId() is null, so this
+            // is added to check if the value exists.
+            if (jsonNode instanceof ObjectNode) {
+                ((ObjectNode) jsonNode).remove(childField.getId());
+            } else {
+                LOG.debug("Logging below details as we are getting com.fasterxml.jackson.databind.node.MissingNode "
+                    + " cannot be cast to class com.fasterxml.jackson.databind.node.ObjectNode");
+                LOG.info("Can not find value for caseFieldId={}, accessControlList={}",
```
danlysiak, line 294:
```suggestion
                LOG.info("Can not find field with caseFieldId={}, accessControlList={}",
```

danlysiak: And perhaps log the `caseTypeId` too?

PR #1712, opened 2021-11-01
RDM-12857

KleoG:
https://tools.hmcts.net/jira/browse/RDM-12857

### Change description ###
Update Case - Validate Event - Update Code for TTL
FTAs to check whether TTL data has changed on mid-event callback during validation 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/casedeletion/TTL.java
```diff
@@ -20,8 +25,14 @@
     public static final String YES = "Yes";
     public static final String NO = "No";
 
+    @JsonDeserialize(using = LocalDateDeserializer.class)
```
sullivangareth, line 28: I don't think you need these annotations, if you use the correctly configured ObjectMapper.

See

`ok, we've this setting to handle that https://github.com/hmcts/ccd-data-store-api/blob/master/src/main/java/uk/gov/hmcts/ccd/config/JacksonObjectMapperConfig.java#L29`

on following PR

`https://github.com/hmcts/ccd-data-store-api/pull/1679`

smathangi (approved):

src/main/java/uk/gov/hmcts/ccd/domain/model/casedeletion/TTL.java
KleoG: Fixed in [7473a67](https://github.com/hmcts/ccd-data-store-api/pull/1712/commits/7473a672e3d99d8d3efc91e65db194e1785a959f)

sullivangareth (approved):

PR #1721, opened 2021-11-09
CCD-2153: Strategic Fix for CCD-2129 & CCD-2130

rajaram-nagarajan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CCD-2153

### Change description ###

Removed the fields in UI if the user don't have read access

```
[ ] Yes
[x] No
```


MSancaktutar (approved):
LGTM subject to test-only improvements that may follow later.

We will need some automated FTAs around the potential regression area, but it's not required in presence of SRT.
A few more unit tests could be good, which may follow later on. No need to suspend the process for those.

rajaram-nagarajan:
The changes has already been added as part of CCD-2240. So, closing this.

PR #1726, opened 2021-11-12
RDM-12954: [AC] Can't Search (ES) for casetype higher than PUBLIC

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-12954

### Change description ###

```
[ ] Yes
[ X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/security/DefaultAuthorisedCaseDefinitionDataService.java
```diff
@@ -53,7 +45,7 @@ public DefaultAuthorisedCaseDefinitionDataService(CaseTypeService caseTypeServic
         CaseTypeDefinition caseTypeDefinition = caseTypeService.getCaseType(caseTypeId);
 
         if (applicationParams.getEnableAttributeBasedAccessControl()
-            && !userHasClassificationForJurisdiction(caseTypeDefinition.getJurisdictionDefinition().getId())) {
+            && !userHasClassificationForJurisdiction(caseTypeDefinition)) {
```
danlysiak, line 48: We should be able to delete this whole `if` statement as part of this PR now - the method is no longer checking jurisdiction SC but instead against the case type, which already happens a few lines down.

kiran-yenigala-hmcts: agree, removed this

danlysiak (approved):
Thanks 🙏

PR #1789, opened 2022-01-18
CCD-2400: Reschedule nightly build

paulridout:
CCD-2400 (https://tools.hmcts.net/jira/browse/CCD-2400)

### Change description ###
Updated Jenkinsfile_nightly.  Changed pipeline trigger hour to 04 so as not to conflict with ccd-definition-store-api nightly build.  Believe that conflict with test data used by ccd-definition-store-api tests is causing ccd-data-store-api tests to fail.

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -2,7 +2,7 @@
 
 properties([
     // H allow predefined but random minute see https://en.wikipedia.org/wiki/Cron#Non-standard_characters
-    pipelineTriggers([cron('H 04 * * *')])
+    pipelineTriggers([cron('H 07 * * *')])
 ])
 
 @Library("Infrastructure")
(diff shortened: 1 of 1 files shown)


MSancaktutar (approved):

Jenkinsfile_nightly
```diff
@@ -2,7 +2,7 @@
 
 properties([
     // H allow predefined but random minute see https://en.wikipedia.org/wiki/Cron#Non-standard_characters
-    pipelineTriggers([cron('H 05 * * *')])
+    pipelineTriggers([cron('H 04 * * *')])
```
MSancaktutar, line 5: May I change my advice to change this to 07, please?
This is to make data store after Def Store, and keep 2 hours difference against random factor in minutes.

paulridout: Hour changed to 07 as suggested.

PR #1799, opened 2022-01-20
RDM-13032 Start Event endpoint to check for Default Value in CaseEventToFields tab

Abi-Laver:
https://tools.hmcts.net/jira/browse/RDM-13032

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventFieldDefinition.java
```diff
@@ -27,6 +27,7 @@
     private Boolean publish;
     private String publishAs;
     private List<CaseEventFieldComplexDefinition> caseEventFieldComplexDefinitions = new ArrayList<>();
+    private String defaultValue = null;
```
danlysiak, line 30:
```suggestion
    private String defaultValue;
```

src/main/java/uk/gov/hmcts/ccd/domain/model/definition/CaseEventFieldDefinition.java
```diff
@@ -161,4 +162,13 @@ public void setPublishAs(String publishAs) {
         this.publishAs = publishAs;
     }
 
+    @ApiModelProperty(value = "")
```
danlysiak, line 165: Let's add a basic description here for Swagger (example suggestion below, feel free to word better!)
```suggestion
    @ApiModelProperty(value = "Default value for the case field, if no existing value")
```

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseService.java
```diff
@@ -25,6 +26,7 @@
 import static uk.gov.hmcts.ccd.config.JacksonUtils.MAPPER;
 
 // TODO CaseService and CaseDataService could probably be merged together.
+@Slf4j
```
danlysiak, line 29:
```suggestion
```
Can be removed now 😄

src/test/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperationTest.java
```diff
@@ -100,6 +100,9 @@
                                                                  + "  \"OrgPolicyCaseAssignedRole\": \"[Claimant]\""
                                                                  + "}");
             data.put("OrganisationPolicyField", orgPolicyDefaultValue);
+
+            JsonNode textFieldDefaultValue = MAPPER.readTree("{ \"TextField0\": \"Default text\" }");
+            data.put("TextField0", textFieldDefaultValue);
```
danlysiak, line 105: This is currently creating an `ObjectNode` with key `TextField0` which contains a subfield also called `TextField0` - for a top level field can use the similar approach as in `CaseService`:
```suggestion
            data.put("TextField0", MAPPER.getNodeFactory().textNode("Default text"));
```
(Same comment applies to other two places in this file)

danlysiak (commented):
Couple of minor comments, implementation looks good though

src/test/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperationTest.java
danlysiak: Also worth adding another top level field in here for another field, but don't add it into the existing case data - to make sure that fields which don't exist also work

src/test/java/uk/gov/hmcts/ccd/domain/service/startevent/DefaultStartEventOperationTest.java
```diff
@@ -128,6 +131,9 @@
                                                      + "}");
             data.put("OrganisationPolicyField", orgPolicy);
 
+            JsonNode textFieldDefaultValue = MAPPER.readTree("{ \"TextField0\": null }");
+            data.put("TextField0", textFieldDefaultValue);
```
danlysiak, line 135: May also be worth adding another top level field in the `existingCaseData` with a non-null value, with a non-null value also in the default, to verify that it doesn't default when it shouldn't.

danlysiak (approved):

PR #1837, opened 2022-02-04
CCD-2520 WA builds failing because of 500s from CCD initial commit

skandala-cgi:
https://tools.hmcts.net/jira/browse/CCD-2520

### Change description ###
WA builds failing because of 500s from CCD

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/matcher/RegionMatcher.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/matcher/RegionMatcher.java
@@ -18,6 +18,7 @@ public class RegionMatcher implements RoleAttributeMatcher {
 
     @Override
     public boolean matchAttribute(RoleAssignment roleAssignment, CaseDetails caseDetails) {
+
         String caseRegion = getRegion(caseDetails).orElse("");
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/matcher/RegionMatcher.java
```diff
@@ -33,9 +34,11 @@ public boolean matchAttribute(RoleAssignment roleAssignment, CaseDetails caseDet
     }
 
     private Optional<String> getRegion(CaseDetails caseDetails) {
-        JsonNode caseManagementLocation = caseDetails.getData().get(CASE_MANAGEMENT__LOCATION);
-        if (caseManagementLocation != null && caseManagementLocation.get(REGION) != null) {
-            return Optional.ofNullable(caseManagementLocation.get(REGION).asText());
+        if (caseDetails != null && caseDetails.getData() != null) {
```
MSancaktutar, line 37: This will avoid NPE, but It looks to me like...
If caseDetails is null, there should be an error handling for it in the first instance.
The method ideally shouldn't be garbage argument (null) and therefore it ideally shouldn't need NP check.

skandala-cgi:
closing this PR as opening another PR

PR #1846, opened 2022-02-09
CCD-2520 WA builds failing because of 500s from CCD

skandala-cgi:
https://tools.hmcts.net/jira/browse/CCD-2520

### Change description ###
WA builds failing because of 500s from CCD

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/matcher/RegionMatcher.java
```diff
@@ -33,9 +33,11 @@ public boolean matchAttribute(RoleAssignment roleAssignment, CaseDetails caseDet
     }
 
     private Optional<String> getRegion(CaseDetails caseDetails) {
-        JsonNode caseManagementLocation = caseDetails.getData().get(CASE_MANAGEMENT__LOCATION);
-        if (caseManagementLocation != null && caseManagementLocation.get(REGION) != null) {
-            return Optional.ofNullable(caseManagementLocation.get(REGION).asText());
+        if (caseDetails != null && caseDetails.getData() != null) {
```
MSancaktutar, line 36: The null check should be applied well before the code comes to this method, to make sure the treatment of a null object is done correctly. Furthermore, we need to generate a warning log to flag the situations where we unexpectedly have encountered a null case object.

src/main/java/uk/gov/hmcts/ccd/domain/model/casedataaccesscontrol/matcher/LocationMatcher.java
```diff
@@ -33,9 +33,11 @@ public boolean matchAttribute(RoleAssignment roleAssignment, CaseDetails caseDet
     }
 
     private Optional<String> getLocation(CaseDetails caseDetails) {
-        JsonNode caseManagementLocation = caseDetails.getData().get(CASE_MANAGEMENT__LOCATION);
-        if (caseManagementLocation != null && caseManagementLocation.get(BASE_LOCATION) != null) {
-            return Optional.ofNullable(caseManagementLocation.get(BASE_LOCATION).asText());
+        if (caseDetails != null && caseDetails.getData() != null) {
```
MSancaktutar, line 36: The null check should be applied well before the code comes to this method, to make sure the treatment of a null object is done correctly. Furthermore, we need to generate a warning log to flag the situations where we unexpectedly have encountered a null case object.

PR #1857, opened 2022-02-15
null check for SecurityClassification

kaanaktas:

https://tools.hmcts.net/jira/browse/CCD-2558

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/casedataaccesscontrol/DefaultCaseDataAccessControl.java
```diff
@@ -307,7 +308,10 @@ public boolean shouldRemoveCaseDefinition(Set<AccessProfile> accessProfiles,
 
     private Set<SecurityClassification> getSecurityClassifications(Set<AccessProfile> accessProfiles) {
         return accessProfiles.stream()
-            .map(accessProfile -> SecurityClassification.valueOf(accessProfile.getSecurityClassification()))
+            .map(accessProfile -> SecurityClassification.valueOf(Optional.ofNullable(accessProfile
```
MSancaktutar, line 311: I think we should return a http-502 if null security classifications are not expectable in any way. This will mean there is an issue in the way data is accepted into and then served to us from the very source of the AccessProfile objects, which is a downstream implementation error.

I understand not expectable, but worth consulting @danlysiak whether null SC's are actually expectable or not.

kaanaktas: This is done, return with 502 now

MSancaktutar (approved):
LGTM

PR #1892, opened 2022-03-02
RDM-13033: Fix Transactional issues

kiran-yenigala-hmcts:
### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperation.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperation.java
@@ -18,6 +18,7 @@ import uk.gov.hmcts.ccd.domain.model.std.Event;
 import uk.gov.hmcts.ccd.domain.service.common.AccessControlService;
 import uk.gov.hmcts.ccd.domain.service.common.CaseAccessCategoriesService;
 import uk.gov.hmcts.ccd.domain.service.common.CaseAccessService;
+import uk.gov.hmcts.ccd.domain.service.getcase.CaseNotFoundException;
 import uk.gov.hmcts.ccd.domain.service.getcase.GetCaseOperation;
 import uk.gov.hmcts.ccd.endpoint.exceptions.ResourceNotFoundException;
 import uk.gov.hmcts.ccd.endpoint.exceptions.ValidationException;
@@ -90,7 +91,7 @@ public class AuthorisedCreateEventOperation implements CreateEventOperation {
         Optional<CaseDetails> filteredCaseDetails = Optional.of(existingCaseDetails)
             .filter(caseAccessCategoriesService.caseHasMatchingCaseAccessCategories(accessProfiles, false));
         if (filteredCaseDetails.isEmpty()) {
-            throw new ResourceNotFoundException(NO_CASE_TYPE_FOUND);
+            throw new CaseNotFoundException(existingCaseDetails.getReferenceAsString());
         }
     }
 
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperation.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperation.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.ccd.domain.service.createevent;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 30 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperation.java
```diff
@@ -80,6 +86,14 @@ public CaseDetails createCaseEvent(String caseReference,
         return verifyReadAccess(caseTypeDefinition, accessProfiles, caseDetails);
     }
 
+    private void verifyCaseAccessCategories(Set<AccessProfile> accessProfiles, CaseDetails existingCaseDetails) {
+        Optional<CaseDetails> filteredCaseDetails = Optional.of(existingCaseDetails)
+            .filter(caseAccessCategoriesService.caseHasMatchingCaseAccessCategories(accessProfiles, false));
+        if (filteredCaseDetails.isEmpty()) {
+            throw new ResourceNotFoundException(NO_CASE_TYPE_FOUND);
```
danlysiak, line 93: Maybe for this scenario we should throw something else as it's more that the user doesn't have access to the case - may make more sense to stay consistent with elsewhere e.g. [here](https://github.com/hmcts/ccd-data-store-api/blob/63b846d297153af1a9fa22b3251b15c9654d52fc/src/main/java/uk/gov/hmcts/ccd/v2/external/controller/CaseController.java#L115)
```suggestion
            throw new CaseNotFoundException(caseDetails.getReferenceAsString());
```

src/test/java/uk/gov/hmcts/ccd/domain/service/common/CaseAccessCategoriesServiceTest.java
```diff
@@ -0,0 +1,81 @@
...
+    void shouldReturnTrueWhenCreateCase() {
+        AccessProfile accessProfile = mock(AccessProfile.class);
+        Predicate<CaseDetails> caseDetailsPredicate = caseAccessCategoriesService
+            .caseHasMatchingCaseAccessCategories(Sets.newHashSet(accessProfile), true);
+        CaseDetails caseDetails = new CaseDetails();
+        assertTrue(caseDetailsPredicate.test(caseDetails));
+    }
+
+    @Test
+    void shouldReturnTrueWhenAccessProfilesAreEmpty() {
```
danlysiak, line 39: If there are no access profiles, I think we should return `false`? Presumably access control elsewhere would always prevent access anyway, but better to be safe!

src/test/java/uk/gov/hmcts/ccd/domain/service/createevent/AuthorisedCreateEventOperationTest.java
```diff
@@ -344,4 +361,112 @@ void shouldFailIfNoCaseReadAccess() {
...
+        doReturn(classifiedCase).when(createEventOperation).createCaseEvent(CASE_REFERENCE,
+            newCaseDataContent);
+
+        assertThrows(ResourceNotFoundException.class, () ->
+            authorisedCreateEventOperation.createCaseEvent(CASE_REFERENCE, newCaseDataContent));
+    }
+
+    @Test
+    @DisplayName("should return true if case data match with case access categories")
+    void shouldReturnTrueIfCaseAccessCategoriesMatchCaseData() {
```
danlysiak, line 388:
```suggestion
    @DisplayName("should return case details if case data match with case access categories")
    void shouldReturnCaseDetailsIfCaseAccessCategoriesMatchCaseData() {
```

danlysiak (commented):
Couple of minor comments, looks good overall though 👍

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/ClassifiedCreateEventOperation.java
```diff
@@ -25,14 +23,13 @@ public ClassifiedCreateEventOperation(@Qualifier("default") CreateEventOperation
...
     @Override
     public CaseDetails createCaseEvent(String caseReference,
                                        CaseDataContent content) {
         final CaseDetails caseDetails = createEventOperation.createCaseEvent(caseReference,
                                                                            content);
         return Optional.ofNullable(caseDetails)
                        .flatMap(classificationService::applyClassification)
-                        .orElseThrow(() -> new CaseNotFoundException(caseReference));
+                       .orElse(null);
     }
```
Thor-tech-of-metal, line 34: Can we avoid returning null ?  It seems that a comuser of  createCaseEvent() would need to use the result to make a decision . would it be ok returning an  Optional .  Not mandatory .  Suggestion

Thor-tech-of-metal (approved):

src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/ClassifiedCreateEventOperation.java
kiran-yenigala-hmcts: This was the existing logic. better to keep this logic for now

PR #1895, opened 2022-03-04
Revert "Revert "Rdm 13033""

kiran-yenigala-hmcts:
### Change description ###

```
[ ] Yes
[ ] No
```


danlysiak (approved):

build.gradle
```diff
@@ -264,8 +264,8 @@ dependencies {
     compile 'org.jooq:jool-java-8:0.9.14'
     compile 'com.github.hmcts:ccd-case-document-am-client:1.7.1'
 
-    testCompile group: 'com.github.hmcts', name: 'ccd-test-definitions', version: '7.14.0'
-    testCompile group: 'com.github.hmcts', name: 'befta-fw', version: '8.7.2-Categories'
+    testCompile group: 'com.github.hmcts', name: 'ccd-test-definitions', version: '7.13.5-prerelease'
```
MSancaktutar, line 267: Please let's use a latest stable release.

build.gradle
```diff
@@ -264,8 +264,8 @@ dependencies {
...
+    testCompile group: 'com.github.hmcts', name: 'befta-fw', version: '8.7.1'
```
MSancaktutar, line 268: Would this now lose the specific delta in 8.7.2-Categories? I'd suggest releasing '8.7.2-Categories' into a next stable release and use that one, if I'm not getting it wrong.

dependency-check-suppressions.xml
```diff
@@ -49,4 +49,17 @@
...
+		<cve>CVE-2021-44832</cve>
+		<cve>CVE-2021-42550</cve>
+		<cve>CVE-2021-22060</cve>
+		<cve>CVE-2022-23181</cve>
+	</suppress>
+
+	<suppress>
+		<packageUrl regex="true">^pkg:maven/org\.springframework\.security/spring\-security\-core@.*$</packageUrl>
+		<vulnerabilityName>CWE-862: Missing Authorization</vulnerabilityName>
+	</suppress>
```
MSancaktutar, line 63: Well, I wouldn't think new suppressions would be needed, as the master build was very recently green.

MSancaktutar (approved):
Would be good to action 2 comments about befta and test-definitions dependencies, otherwise good. As discussed, would be good to make sure the fix in the contract isn't breaking any client code that was meant to tolerate the broken aspect of the contract.

PR #1957, opened 2022-04-22
CCD-2950: Multiparty fix for child case view field

rajaram-nagarajan:
### JIRA link  ###

https://tools.hmcts.net/jira/browse/CCD-2950

### Change description ###

Multiparty fix for child case view field

```
[ ] Yes
[x] No
```


dependency-check-suppressions.xml
```diff
@@ -74,6 +74,7 @@
 
 	<suppress until="2022-05-25">
 		<cve>CVE-2022-23708</cve>
+		<cve>CVE-2022-22968</cve>
```
MSancaktutar, line 77: i don't think needed.

rajaram-nagarajan: yes, removed.

MSancaktutar (approved):
LGTM - thank you.

PR #1959, opened 2022-04-29
suppression for CVE-2022-22968

duranmf:
This is suppression for:
In Spring Framework versions 5.3.0 - 5.3.18, 5.2.0 - 5.2.20, and older unsupported versions, the patterns for disallowedFields on a DataBinder are case sensitive which means a field is not effectively protected unless it is listed with both upper and lower case for the first character of the field, including upper and lower case for the first character of all nested fields within the property path.

```
[ ] Yes
[x ] No
```


MSancaktutar (changes requested):
Is this indeed needed? Nightly pipeline is already passing the dependency checks.

PR #1979, opened 2022-05-23
Ccd 216 malformed case type id error response

paulridout:
CCD-216 (https://tools.hmcts.net/jira/browse/CCD-216)

### Change description ###
Created DataStoreHttpStatusRequestRejectedHandler class.  This handles RequestRejectedExceptions raised by the Spring StrictHttpFirewall that occur when the URL contains forbidden characters.  The handler converts the status code of the response from 500 to 403.  Created FirewallConfiguration class to register a bean for the new handler and make it available to Spring.  Created a new functional test - S-1032.2 - to test this scenario.

Created CaseTypeIdValidator class.  This is used to validate the case type id in requests that are accepted by StrictHttpFirewall.  Created ValidCaseTypeId annotation for CaseTypeIdValidator and updated CaseController to use it to validate case type ids.

Updated RestExceptionHandler to add a new handler for ConstraintViolationExceptions.  These exceptions are raised when the case type id fails validation.  The handler returns a response with the status code 400.  Updated HttpError class to add a new constructor which is used by the new handler and reverted change to Exception & HttpServiceRequest constructor - new one provides the same functionality.  Changed existing handleException() method to also use this constructor.

Renamed S-1032 test to S-1032.1 and re-enabled it.  Changed expected response to match that returned by the new case type id validation.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/aat/resources/features/F-1003 - Case Creation With Documents/F-1003.feature
+++ b/src/aat/resources/features/F-1003 - Case Creation With Documents/F-1003.feature
@@ -36,7 +36,7 @@ Feature: F-1003: Submit Case Creation
     And   the response has all other details as expected,
     And   a call [to retrieve case details by case id] will get the expected response as in [Default_Get_Case_Data_Base_01].
 
-  @S-1032 @Ignore # Defect AM-707 / RDM-8403
+  @S-1032.1
   Scenario: must get an error response for a malformed case type ID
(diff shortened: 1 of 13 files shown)


marvincudjoe (approved):
LGTM

skandala-cgi (commented):
LGTM

skandala-cgi (approved):

PR #1985, opened 2022-05-25
CCD-3152 -Enable EventEnablingCondition for INJECTED_DATA

sullivangareth:
https://tools.hmcts.net/jira/browse/CCD-3152

### Change description ###

Update datastore to NOT filter out actionable events when EventEnablingConditions for InjectedData are valid and match

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
```diff
@@ -24,4 +33,24 @@ public Boolean isEventEnabled(String enablingCondition,
...
+                        caseViewField -> TextNode.valueOf((String) caseViewField.getValue())));
+
+            return !caseViewFieldData.isEmpty()
+                && this.enablingConditionParser.evaluate(replaceDotDelimiter(enablingCondition), caseViewFieldData);
+        }
+        return Boolean.TRUE;
+    }
+
+    private String replaceDotDelimiter(String delimitedString) {
+        return delimitedString.replace(DOT_DELIMITER, DOT_DELIMITER_REPLACEMENT);
```
danlysiak, line 54: Another observation to do with chaining e.g. `ComplexField.EmailField = "email.value@has.a.dot.com" AND [INJECTED_DATA.OtherField] = "B"` - we should probably replace `INJECTED_DATA.` with `INJECTED_DATA__` (or similar), otherwise any complex fields being referenced, or dots in the values, will also be replaced.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
```diff
@@ -24,4 +33,24 @@ public Boolean isEventEnabled(String enablingCondition,
         }
         return Boolean.TRUE;
     }
+
+    public Boolean isEventEnabled(String enablingCondition,
```
danlysiak, line 37: I'm not sure a separate method is an appropriate way for this, bearing in mind that the general "ShowCondition" feature is pretty flexible. If we had even a slightly more complex scenario such as `SomeCaseDataField = "A" AND [INJECTED_DATA.otherField] = "B"`, then looks like this isn't going to have access to the case data too to be able to evaluate `SomeCaseDataField`?

It might be easier (and work for any kind of chaining) to modify the existing method in this class to take an argument such as just the injection metdata/an initial map containing just the injection metadata, and merge with `caseDetails.getCaseDataAndMetadata()`?

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
```diff
@@ -24,4 +33,24 @@ public Boolean isEventEnabled(String enablingCondition,
...
+                                  List<CaseViewField> caseViewFields) {
+        if (StringUtils.isNotEmpty(enablingCondition)) {
+            Map<String, Object> caseViewFieldData = caseViewFields.stream()
+                .filter(injectedDataCaseViewField -> injectedDataCaseViewField.getId().startsWith("[INJECTED_DATA"))
+                .collect(
+                    Collectors.toMap(
+                        caseViewField -> replaceDotDelimiter(caseViewField.getId()),
+                        caseViewField -> TextNode.valueOf((String) caseViewField.getValue())));
```
danlysiak, line 45:
```suggestion
                        CaseViewField::getValue));
```
Wondering whether we need to bother with casting and creating a `TextNode`? It looks like later logic within `enablingConditionParser.evaluate()` already converts objects to appropriate `JsonNode` implementations.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
sullivangareth: Fixed in https://github.com/hmcts/ccd-data-store-api/pull/1985/commits/e9775ac2efe226ec855e75b4e136f32f16967b80

src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseEventEnablingService.java
sullivangareth: Fixed in https://github.com/hmcts/ccd-data-store-api/pull/1985/commits/e9775ac2efe226ec855e75b4e136f32f16967b80

danlysiak (approved):
Quick look over following changes, LGTM ✅

mattnayler (approved):
LGTM 👍

hemantt (approved):

PR #2002, opened 2022-06-14
fix idam exception handling to sync with ServiceAuthFilter

kaanaktas:

https://tools.hmcts.net/jira/browse/CCD-3321

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/ccd/security/idam/IdamRepository.java
```diff
@@ -66,19 +65,13 @@ public String getDataStoreSystemUserAccessToken() {
...
-        if (exception instanceof FeignException.FeignClientException) {
-            if (httpStatus != HttpStatus.UNAUTHORIZED) {
-                httpStatus = HttpStatus.INTERNAL_SERVER_ERROR;
-            }
-        } else if (exception instanceof FeignException.FeignServerException) {
-            if (httpStatus == HttpStatus.INTERNAL_SERVER_ERROR) {
-                httpStatus = HttpStatus.BAD_GATEWAY;
-            }
+        if (isClientError) {
+            return new InvalidTokenException(exception.getMessage(), exception);
```
rajaram-nagarajan, line 72: Can you throw the exceptions from here itself?

kaanaktas: Throwing exception directly from here would change the caller method's return structure as well. I removed **parseFeignException** method and moved its flow to the catch block.

hemantt (approved):

PR #2071, opened 2022-08-11
CCD-3533 auto create new suppression for cve

marvincudjoe:
https://tools.hmcts.net/jira/browse/CCD-3533

### Change description ###

```
[ ] Yes
[ ] No
```


.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Commit new Suppressions
+        run: ./gradlew suppressCves
+      - name: Clean suppressions
+        run: ./gradlew cleanSuppressions
+      - name: Commit
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          skip_dirty_check: false
+          commit_message: "Suppress CVEs"
+          branch: 'develop'
```
danlysiak, line 43:
```suggestion
          branch: 'master'
```

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,43 @@
+name: Suppressions
+
+on:
+  schedule:
+    - cron: "05 8-16/3 * * 1-5" # On the 5th minute, every 3 hours, from 8am to 4pm
```
danlysiak, line 5: I think we should do this throughout the night too, to reduce the risk of nightly pipelines failing as well - every 3 hours sounds reasonable though!

danlysiak, line 5: I think we should do this in a wider time period, to reduce the risk of nightly pipelines failing as well - every 3 hours sounds reasonable though!
```suggestion
    - cron: "05 6-18/3 * * 1-5" # On the 5th minute, every 3 hours, from 8am to 4pm
```

paulridout (commented):
Am I right in thinking that this creates a new <suppress> section for each new CVE rather than appending to a single existing <suppress> section?  The Jira ticket mentions that the proof of concept didn't quite match the proposal, so I was just wondering if this uses the proof of concept approach.

danlysiak (approved):

kaanaktas (commented):
Do we duplicate suppressions, even if they already exist? This is what I understood from the comment on the ticket.

kaanaktas (approved):
lgtm

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,42 @@
+name: Suppressions
+
+on:
+  schedule:
+    # // H allow predefined but random minute see https://en.wikipedia.org/wiki/Cron#Non-standard_characters
+    # every 3rd hour from 6 am to 6pm on every day-of-week from Monday through Friday
+    - cron: "H 6-18/3 * * 1-5"
```
rajaram-nagarajan, line 7: @marvincudjoe - we can run this over the weekends too

rajaram-nagarajan, line 7: @marvincudjoe Let's run from 10AM to 12AM

rajaram-nagarajan (changes requested):

rajaram-nagarajan (approved):
LGTM

PR #2023, opened 2022-07-05
dependencyCheck WIP

danlysiak:
**Before creating a pull request make sure that:**

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

Please remove this line and everything above and fill the following sections:

### Change description ###

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ ] No
```

PR #2074, opened 2022-08-15
CCD-3533 auto amend temp suppression section

marvincudjoe:
https://tools.hmcts.net/jira/browse/CCD-3533

### Change description ###
Builds on work from #2023 and #2071 

Add code to amend Temporary Suppression section
- new CVEs are now appended to the existing suppress section
- notes section is appended with 'cve name refer Ticket' per new CVE

```
[ ] Yes
[ x] No
```

### DIFF
--- a/gradle/suppress.gradle
+++ b/gradle/suppress.gradle
@@ -5,13 +5,12 @@ import org.w3c.dom.Document
 import org.w3c.dom.Element
 import org.w3c.dom.Node
 import org.w3c.dom.NodeList
-
 import javax.xml.xpath.XPathConstants
 import javax.xml.xpath.XPathExpression
 import javax.xml.xpath.XPathFactory
 
-void suppressCves(reportName) {
-    reportName  = reportName == null || reportName as String == "" ? "dependency-check-report.xml" : reportName as String
+void suppressCves(String reportName='dependency-check-report.xml') {
+    // Default output directory is 'build/reports'
     def reportFile = new File(project.file(dependencyCheck.getOutputDirectory()), reportName)
     if (!reportFile.isFile()) {
         throw new FileNotFoundException("Report file not found.\n" +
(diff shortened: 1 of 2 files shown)


gradle/suppress.gradle
```diff
@@ -24,40 +23,53 @@ void suppressCves(reportName) {
...
+            unSuppressedCves.each {
+                String whiteSpaceForIndent = "\n" + "\u0020".repeat(8) // white space for indentation
+                Document document = suppressionDoc.getOwnerDocument()
+                noteNode.appendChild(
+                        document.createTextNode(whiteSpaceForIndent + it + " refer [Ticket]"))
+                noteNode.getParentNode()
+                        .appendChild(document.createElement("cve")).appendChild(document.createTextNode(it))
+            }
+        }
     }
```
paulridout, line 52: 1.) Might want to change the for loop to a while loop with an exit condition.  There's only going to be one notes section you want to update, so you can exit the loop after updating it.
2.) The value of the whiteSpaceForIndent variable never changes after it has been declared so move the declaration outside of the outer for loop.  That'll save it from be redeclared in each iteration of the inner loop.  Might also be an idea to make it a constant.
3.) The value of the document variable never changes after it has been declared so move the declaration outside of the outer for loop.
4.) Could create a variable called suppressNode as first line of if statement and set it to value of noteNode.getParentNode().  Use this in place of noteNode.getParentNode() on line 48.  This would avoid the need to call noteNode.getParentNode() on each iteration of the inner loop.

marvincudjoe: 1. done; uses find{} in replace of the for loop. Exists when true is returned
2. done; moved whiteSpaceForIndent outside the loop
3. since it's only needed once, it's left inside the if statement
4. done

marvincudjoe:
This pr can be merged into #2071. Addressed this as a comment in the ticket.

paulridout (approved):
Changes look OK.

gradle/suppress.gradle
```diff
@@ -5,59 +5,78 @@ import org.w3c.dom.Document
...
-static def cleanXml(Element suppressions) {
+static def cleanXml(Element suppression) {
     // Strip out all empty nodes
     XPathExpression xpathExp = XPathFactory.newInstance().newXPath().compile(
             "//text()[normalize-space(.) = '']")
 
-    NodeList emptyTextNodes = xpathExp.evaluate(suppressions, XPathConstants.NODESET) as NodeList
+    NodeList emptyTextNodes = xpathExp.evaluate(suppression, XPathConstants.NODESET) as NodeList
     if (emptyTextNodes.getLength()) {
+        println("found")
```
danlysiak, line 68: Looks like debug that can now be deleted

danlysiak (approved):

gradle/suppress.gradle
marvincudjoe: I'll remove this once merged

PR #2116, opened 2022-10-05
LAU-496 Fix null pointer exception for lau case refs.

MartinYSpasov:
https://tools.hmcts.net/jira/browse/LAU-496

### Change description ###
Fix null pointer exception for lau case refs.

```
[ ] Yes
[x] No
```


davejones74 (approved):

paulridout (commented):
Might be an idea to add a test class for SearchLog that explicitly tests the behaviour of setCaseRefs().  The existing CaseSearchPostRequestTest.jsonContructionTestWithEmptyCaseRefs() covers the scenario where the list of case refs is null, but it uses the constructor to set up the class rather than the setCaseRefs() method.  Adding a test class would improve coverage and prove that the change prevents a null pointer exception.

paulridout (approved):
Changes look OK.

src/test/java/uk/gov/hmcts/ccd/domain/model/lau/SearchLogTest.java
```diff
@@ -0,0 +1,71 @@
...
+        assertThat(searchLog.getUserId()).isEqualTo(userId);
+        assertThat(searchLog.getTimestamp()).isEqualTo(timestamp.format(ISO_INSTANT));
+        assertThat(searchLog.getCaseRefs().size()).isEqualTo(3);
+        assertThat(searchLog.getCaseRefs().get(0)).isEqualTo("123");
+        assertThat(searchLog.getCaseRefs().get(1)).isEqualTo("567");
+        assertThat(searchLog.getCaseRefs().get(2)).isEqualTo("789");
+    }
+
+    @Test
+    void shouldNotThrowNpeWhenCaseRefIsNull() {
```
rajaram-nagarajan, line 52: We need to add another test case to pass null value for caseRefs through SearchLog constructor. (as highlighted by Paul too)
Ex:  final SearchLog searchLog = new SearchLog(userId, null, timestamp);

src/test/java/uk/gov/hmcts/ccd/domain/model/lau/SearchLogTest.java
```diff
@@ -0,0 +1,71 @@
...
+        final SearchLog searchLog = new SearchLog();
+        final String caseRefs = null;
+        searchLog.setCaseRefs(caseRefs);
+
+        assertThat(searchLog.getCaseRefs()).isNull();
+    }
+
+    @Test
+    void shouldTrimCaseRefs() {
+        final SearchLog searchLog = new SearchLog();
```
rajaram-nagarajan, line 62: We can add another test case to check the trim through SearchLog constructor.
Ex: caseRefs = asList(" 123 "," 567 "," 789  ");

src/main/java/uk/gov/hmcts/ccd/domain/model/lau/SearchLog.java
```diff
@@ -19,8 +20,6 @@ public class SearchLog implements Serializable {
 
     public static final long serialVersionUID = 432973322;
 
-    private static final String ANY_SPACE_MATCH = "\\s*";
-
     private String userId;
     private List<String> caseRefs;
```
rajaram-nagarajan, line 24: Should we need to initiate with empty list to avoid NPE for the caller?

rajaram-nagarajan (changes requested):

src/test/java/uk/gov/hmcts/ccd/domain/model/lau/SearchLogTest.java
MartinYSpasov: SearchLog constructor was removed since it was used for testing purposes only.

src/test/java/uk/gov/hmcts/ccd/domain/model/lau/SearchLogTest.java
MartinYSpasov: SearchLog constructor was removed since it was used for testing purposes only.

src/main/java/uk/gov/hmcts/ccd/domain/model/lau/SearchLog.java
MartinYSpasov: No need, caseRefs can be null.

rajaram-nagarajan (approved):
LGTM

PR #2123, opened 2022-10-12
CCD-3798 Access Granted & mapped access profiles 

mattnayler:

   [CCD-3798](https://tools.hmcts.net/jira/browse/CCD-3798) _"ACCESS_GRANTED metadata needs to consider role assignments giving access to cases"_

### Change description ###

Access Granted metadata should only consider role assignments giving access to the case.

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/casedataaccesscontrol/DefaultCaseDataAccessControl.java
```diff
@@ -420,9 +432,25 @@ private static boolean isRoleAssignmentsOfInterestToAccessProcessCheck(RoleAssig
...
+            || roleAssignment.getGrantType().equals(GrantType.BASIC.name());
+    }
+
+    private List<GrantType> generatePostFilteringAccessGrants(CaseDetails caseDetails,
+                                                              List<RoleAssignment> roleAssignments) {
+        String caseState = caseDetails.getState();
+        CaseTypeDefinition caseType = caseDefinitionRepository.getCaseType(caseDetails.getCaseTypeId());
+
         return roleAssignments.stream()
+            .filter(DefaultCaseDataAccessControl::isRoleAssignmentsOfInterestToAccessGrantCheck)
```
danlysiak, line 446:
```suggestion
            .filter(this::isRoleAssignmentsOfInterestToAccessGrantCheck)
```
If we swap to `this` then we don't need to make the methods `static`. (I think I missed this in the original PR when we introduced `isRoleAssignmentsOfInterestToAccessGrantCheck()` that it was `static`)

src/main/java/uk/gov/hmcts/ccd/domain/service/casedataaccesscontrol/DefaultCaseDataAccessControl.java
```diff
@@ -420,9 +432,25 @@ private static boolean isRoleAssignmentsOfInterestToAccessProcessCheck(RoleAssig
...
+            .filter(roleAssignment -> userHasAccessToCaseUsingRoleAssignments(caseState,
+                                                                              caseType,
+                                                                              List.of(roleAssignment))
+            )
             .map(roleAssignment -> GrantType.valueOf(roleAssignment.getGrantType()))
+            .distinct()
+            .sorted()
             .collect(Collectors.toList());
```
danlysiak, line 454: Optional suggestion, as I appreciate it requires a little change to `CaseAccessMetadata` and any associated tests too:
```suggestion
            .collect(Collectors.toSet());
```
We always end up needing the distinct values and there's no requirement for ordering so a `Set` is a bit more appropriate, and we could also remove the (repeated) `.distinct().sorted()` from the getter in `CaseAccessMetadata`.

mattnayler: Updated in f5f3be70b4f9421e46f4f2a96440eb99942b936c after discussion.  Calls to `distinct()` and `sorted()` removed as they are also performed in `CaseAccessMetadata.getAccessGrantsString(...)`, i.e. priror to output generation.  However left as list as Sort will make it easier to verify the CSV output in FTAs and volume of list is always low.

danlysiak (approved):

PR #2231, opened 2023-03-01
CCD-4371 Do not use TextValidator when validating a Document BaseType

marvincudjoe:
https://tools.hmcts.net/jira/browse/CCD-4371

### Change description ###
Remove TextValidator on category_id
- Using TextValidator uses the regex pattern supplied for the document.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/types/DocumentValidator.java
```diff
@@ -149,10 +146,10 @@ private List<ValidationResult> validateCategoryId(final String dataFieldId,
             return Collections.emptyList();
         }
 
-        List<ValidationResult> validationResults =
```
padmapriyanalam, line 152: Removing the textValidator.validate() removes Max and Min validation along with Regex. Is this expected?

rajaram-nagarajan (approved):
LGTM

src/main/java/uk/gov/hmcts/ccd/domain/types/DocumentValidator.java
marvincudjoe: Those checks work better for fields in the CaseFields tab. You can specify the min, max and regex for a field. For 
 Complex Types it's for for the complex type itself than the items inside it.

marvincudjoe: So for an item like category_id in a Document{} a Min, Max isn't needed. Unless it's made a requirement.

JamesCollettCGI (approved):

PR #2277, opened 2023-06-26
Further changes to reduce unnecessary logging

danlysiak:
### Change description ###
Further changes to reduce unnecessary logging - see PR comments for rationale behind each change.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/ccd-data-store-api/values.yaml
+++ b/charts/ccd-data-store-api/values.yaml
@@ -88,4 +88,4 @@ java:
 
     DATA_STORE_CROSS_JURISDICTIONAL_ROLES: caseworker-caa,caseworker-approver,next-hearing-date-admin
     DATA_STORE_CITIZEN_ROLES: citizen,letter-holder
-    DATA_STORE_SECURITY_LOGGING_FILTER_PATH_REGEX: '.*'
+    DATA_STORE_SECURITY_LOGGING_FILTER_PATH_REGEX: 'DISABLED'
--- a/src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
@@ -136,7 +136,7 @@ public class DefaultUserRepository implements UserRepository {
                 .getBody();
 
             if (userDefaultCollection == null || isEmpty(userDefaultCollection.getUserDefaults())) {
-                LOG.error("User Profile not exists for userId '{}'", userId);
+                LOG.debug("User Profile not exists for userId '{}'", userId);
                 return null;
             }
 
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceImpl.java
@@ -60,7 +60,7 @@ public class AccessControlServiceImpl implements AccessControlService {
 
         boolean hasAccess = hasAccessControlList(accessProfiles, criteria, caseType.getAccessControlLists());
         if (!hasAccess) {
-            LOG.info(NO_ROLE_FOUND, "caseType",
+            LOG.debug(NO_ROLE_FOUND, "caseType",
                     caseType.getId(),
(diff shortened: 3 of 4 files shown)


charts/ccd-data-store-api/values.yaml
```diff
@@ -88,4 +88,4 @@ java:
 
     DATA_STORE_CROSS_JURISDICTIONAL_ROLES: caseworker-caa,caseworker-approver,next-hearing-date-admin
     DATA_STORE_CITIZEN_ROLES: citizen,letter-holder
-    DATA_STORE_SECURITY_LOGGING_FILTER_PATH_REGEX: '.*'
+    DATA_STORE_SECURITY_LOGGING_FILTER_PATH_REGEX: 'DISABLED'
```
danlysiak, line 91: The same information that this filter logs is also logged elsewhere - adds no extra benefit.
Could add a proper regex that never matches anything like `$^` but suggesting just the word `DISABLED` for readability :)

src/main/java/uk/gov/hmcts/ccd/data/user/DefaultUserRepository.java
```diff
@@ -136,7 +136,7 @@ public UserDefault getUserDefaultSettings(final String userId) {
                 .getBody();
 
             if (userDefaultCollection == null || isEmpty(userDefaultCollection.getUserDefaults())) {
-                LOG.error("User Profile not exists for userId '{}'", userId);
+                LOG.debug("User Profile not exists for userId '{}'", userId);
```
danlysiak, line 139: For the last couple of years the majority of users do NOT have a user profile. It's not an error and implementation handles properly when none is provided, so this log shouldn't be required.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceImpl.java
```diff
@@ -60,7 +60,7 @@ public boolean canAccessCaseTypeWithCriteria(final CaseTypeDefinition caseType,
 
         boolean hasAccess = hasAccessControlList(accessProfiles, criteria, caseType.getAccessControlLists());
         if (!hasAccess) {
-            LOG.info(NO_ROLE_FOUND, "caseType",
+            LOG.debug(NO_ROLE_FOUND, "caseType",
```
danlysiak, line 63: In last 7 days this is logged ~8.5m times, it's part of the standard filtering so we shouldn't need to log everything.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceImpl.java
```diff
@@ -86,7 +86,7 @@ public boolean canAccessCaseStateWithCriteria(final String caseState,
         boolean hasAccess = hasAccessControlList(accessProfiles, criteria, stateACLs);
 
         if (!hasAccess) {
-            LOG.error(NO_ROLE_FOUND, "caseState",
+            LOG.debug(NO_ROLE_FOUND, "caseState",
```
danlysiak, line 89: In last 7 days this is logged ~6m times, it's part of the standard filtering so we shouldn't need to log everything.

src/main/java/uk/gov/hmcts/ccd/domain/service/common/AccessControlServiceImpl.java
```diff
@@ -86,7 +86,7 @@ public boolean canAccessCaseStateWithCriteria(final String caseState,
...
                     caseState,
                     AccessControlService.extractAccessProfileNames(accessProfiles),
                     "caseStateACL",
```
danlysiak, line 92: General note for other similar logs to the ones changed in this class:
- `CaseEvent` filtering logs only ~300 times in the last 7 days.
- `CaseField` filtering logs only ~20 times in the last 7 days. 

Both of these are messages that might be useful for debugging scenarios, and given the very low occurrence compared to the millions of others, suggest we keep them.

rajaram-nagarajan (approved):
LGTM

kerry-anne-mclaughlin (approved):

PR #2369, opened 2024-02-19
[Case Update] Populate Document's "upload_timestamp" Value [DO NOT MERGE Use CSFD-89]

markdathornehmcts:

https://tools.hmcts.net/jira/browse/CSFD-6

### Change description ###

To populate the date and the time at which a new document was added to a case during case update,
So that it is possible for other components (i.e. Case File View) to use this date for document sorting functionality.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ccd/domain/service/getcasedocument/CaseDocumentTimestampService.java
```diff
@@ -28,7 +30,10 @@ public void addUploadTimestamps(CaseDetails caseDetailsModified, CaseDetails cas
     protected List<String> findUrlsNotInOriginal(CaseDetails caseDetailsModified, CaseDetails caseDetailsInDb) {
         List<String> documentUrlsFromDb = getDocumentUrls(caseDetailsInDb);
 
-        List<JsonNode> jsonNodes = findNodes(caseDetailsModified.getData().values());
+        List<JsonNode> jsonNodes = new ArrayList<>();
+        if (null != caseDetailsModified && null != caseDetailsModified.getData()) {
```
sonny-hmcts, line 34: feel that the if statement would be more readable the other way around for example caseDetailsModified != null is more readable than null != caseDetailsModified but this is just preference

sonny-hmcts (approved):

src/main/java/uk/gov/hmcts/ccd/domain/service/getcasedocument/CaseDocumentTimestampService.java
markdathornehmcts: I agree but, particularly for equality/assignage, null can never be assigned an alternative value. Probably a hang over from C coding standards.

RebeccaBaker (approved):

PR #2386, opened 2024-03-15
CSFD-97 [Case Create/Update] Populate Document's "upload_timestamp" Value - Feature Toggle

markdathornehmcts:

https://tools.hmcts.net/jira/browse/CSFD-97

### Change description ###

Feature toggle the add upload timestamp logic

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
+++ b/src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
@@ -182,6 +182,12 @@ public class ApplicationParams {
     @Value("${enable-case-users-db-sync}")
     private boolean enableCaseUsersDbSync;
 
+    @Value("${ccd.jurisdiction-upload-timestamp-enabled}")
+    private boolean jurisdictionUploadTimestampEnabled;
+
+    @Value("${ccd.upload-timestamp-featured-jurisdiction}")
+    private String uploadTimestampFeaturedJurisdiction;
+
     @Value("${audit.log.enabled:true}")
     private boolean auditLogEnabled;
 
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/ccd/ApplicationParams.java
```diff
@@ -182,6 +182,12 @@ public class ApplicationParams {
     @Value("${enable-case-users-db-sync}")
     private boolean enableCaseUsersDbSync;
 
+    @Value("${ccd.jurisdiction-upload-timestamp-enabled}")
+    private boolean jurisdictionUploadTimestampEnabled;
+
+    @Value("${ccd.upload-timestamp-featured-jurisdiction}")
+    private String uploadTimestampFeaturedJurisdiction;
+
```
danlysiak, line 190: The first one can be made redundant when we have a per-jurisdiction flag too. If we want it off for all then we just provide none for the list of jurisdictions. We'll also need to support multiple jurisdictions so something like the following is probably more suitable:
```suggestion
    @Value("#{'${ccd.upload-timestamp.enabled-jurisdictions}'.split(',')}")
    private List<String> uploadTimestampEnabledJurisdictions;
```

src/main/java/uk/gov/hmcts/ccd/domain/service/getcasedocument/CaseDocumentTimestampService.java
```diff
@@ -111,4 +121,11 @@ protected void insertUploadTimestamp(JsonNode node, String uploadTimestamp) {
         }
     }
 
+    protected boolean isJurisdictionUploadTimestampFeatureEnabled(String caseJurisdiction) {
```
danlysiak, line 124:
```suggestion
    private boolean isJurisdictionUploadTimestampFeatureEnabled(String caseJurisdiction) {
```

RebeccaBaker (approved):

PR #2484, opened 2024-10-15
CCD-5919 Enable Data Store multi-party fix by default

patelila:
… is currently in Flux. values.yaml updated to what was in the properties file.

- [ ] the `keep-helm` label has been added, if the helm release should be persisted after a successful build

### Change description ###

```
[ ] Yes
[x] No
```


src/main/resources/application.properties
```diff
@@ -77,10 +77,6 @@ ccd.callback.retries=1,5,10
 
 ccd.case.search.wildcards.allowed=false
 
-ccd.multiparty.fix.enabled=${CCD_MULTIPARTY_FIX_ENABLED:false}
-ccd.multiparty.case-types=${CCD_MULTIPARTY_CASE_TYPES:CIVIL}
-ccd.multiparty.events=${CCD_MULTIPARTY_EVENTS:DEFENDANT_RESPONSE,ACKNOWLEDGE_CLAIM}
```
RebeccaBaker, line 82: do we not need to set line 80 to true 🤔  isMultipartyFixEnabled is still referenced throughout the code

patelila: I am not sure about this. However I made changes thinking that as the ticket is about the enviornments having issues that the cop-flux-config setting is being overwritten with app setting values. So I have removed this from app settings. And for purposes of testing moved the setting into test app settings. So the environments should now have the correct settings being used (as in from cnp-flux-config). Let me know if this is not the case.

RebeccaBaker: mu understanding is we just need to update the default values to True, *, *,  to match what is in flux config https://github.com/hmcts/cnp-flux-config/blob/fd9ca13d3ad668497691a344851a741b320bccb5/apps/ccd/ccd-data-store-api/ccd-data-store-api.yaml#L13-L15

danlysiak: Yep as Rebecca says, we should make the properties defaults match what is in Flux.

RebeccaBaker (approved):

amina-khatunCGI (approved):

PR #2555, opened 2025-04-08
Fix -> fortify issue to use try with resources instead

dinesh1patel:

NA

### Change description ###

Fix fortify issue to use try with resources instead

```
[ ] Yes
[X] No
```


dinesh1patel:
Closing as already implemented in master

PR #2579, opened 2025-05-22
Add wa_task_management_api to s2s whitelist by default

linusnorton:
This change is already enabled in all other environments (e.g. prod: https://github.com/hmcts/cnp-flux-config/blob/master/apps/ccd/ccd-data-store-api/prod.yaml#L22) but not having it in the base helm chart breaks the integration of chart-ccd and wa-chart.


kiran-yenigala-hmcts (approved):

markdathornehmcts (approved):
LGTM

RebeccaBaker (approved):

PR #2691, opened 2025-12-04
LAU-1325 Add Interceptor to override token

priyankaVerma21:
https://tools.hmcts.net/jira/browse/LAU-1325

### Change description ###
Add Interceptor 

```
[ ] Yes
[x] No
```


rapolaskaseliscgi:
> This should be perf tested

Actually I had a look in the existing implementation and it looks like we call s2s before each call anyway, so putting that call in the interceptor doesn't change much. We just need to make sure that the feign interceptor is not used in any other feign clients.

PR #2728, opened 2025-12-17
currently timeout is 10 sec not 60sec and IA team has issue in calling stitching for big documents EM-6942

mdayican:

currently timeout is 10 sec not 60sec and IA team has issue in calling stitching for big documents EM-6942

fix socket timeout only for callback rest template, there are other rest templates that needs to be fixed.

Screenshot 2025-12-17 at 14 21 17

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG JAVA_OPTS="-Djava.security.egd=file:/dev/./urandom"
 # renovate: datasource=github-releases depName=microsoft/ApplicationInsights-Java
-ARG APP_INSIGHTS_AGENT_VERSION=3.7.6
+ARG APP_INSIGHTS_AGENT_VERSION=3.7.7
 ARG PLATFORM=""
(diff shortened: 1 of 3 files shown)


patelila:
You don’t need to change the chart code to adjust the per-attempt read timeout:

  - The timeout is driven by the Spring property http.client.read.timeout, which already maps to the env var HTTP_CLIENT_READ_TIMEOUT (default 30000 in application.properties).
  - At deploy time, just override that env var (e.g., via Helm --set env.HTTP_CLIENT_READ_TIMEOUT=28000 or a values override file) or set HTTP_CLIENT_READ_TIMEOUT in your pod/deployment manifest. No
    code/chart edits are required.
  - Locally, you can also pass -Dhttp.client.read.timeout=28000 when running tests/apps.

  So pick the value you want (e.g., 28000) and set HTTP_CLIENT_READ_TIMEOUT in your deployment configuration.

PR #2840, opened 2026-03-26
CCD-6626 CCD is retaining value even after user has manually removed it from UI

markdathornehmcts:

https://tools.hmcts.net/jira/browse/CCD-6626

### Change description ###

CCD is retaining values on Check Your Answers which were removed by the user on a page.

```
[ ] Yes
[*] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseService.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseService.java
@@ -23,7 +23,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
-import static java.util.stream.Collectors.toList;
 import static uk.gov.hmcts.ccd.config.JacksonUtils.MAPPER;
 
 // TODO CaseService and CaseDataService could probably be merged together.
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/common/CaseService.java
```diff
@@ -76,12 +75,24 @@ public CaseDetails createNewCaseDetails(String caseTypeId, String jurisdictionId
      * @return <code>Optional&lt;CaseDetails&gt;</code> - CaseDetails wrapped in Optional
      */
     public CaseDetails populateCurrentCaseDetailsWithEventFields(CaseDataContent content, CaseDetails caseDetails) {
```
paulridout, line 77: Changes look OK, but I'm not sure how SSCS invokes this method.  I think it might be via the CaseDataValidatorController or CaseDetailsEndpoint endpoints, but I can't identify route.  It might be something from XUI.

I would suggest renaming the method it's now populating with whatever is returned by getData() as well as getEventData(), but only do this if the change fixes the issue.  Try it out as a proof of concept first.

paulridout (approved):
Changes look OK, but would recommend trying it out as a proof of concept first.

padmapriyanalam (approved):
change looks ok.  I agree with Paul. Could this be tested in a lower environment first to ensure that the change fixes the issue?

PR #2843, opened 2026-04-07
CCD-7242 Sponsor contact details do not show up on the Check Your Answers page

markdathornehmcts:

https://tools.hmcts.net/jira/browse/CCD-7242

### Change description ###

Test harness built to determine whether given field(s) are returned for given event(s) and given role(s), given the definition spreadsheet.

```
[ ] Yes
[x] No
```


src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
paulridout, line 1: This class isn't a test, it's a utility for checking a definition spreadsheet.  JUnit just appears to be used as a convenient mechanism for running it.  It probably needs to go in a separate support-tools source directory underneath the root directory of this project and reworked as a runnable class with a main() method.

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,490 @@
...
+ *       {@code RoleToAccessProfiles} sheets.</li>
+ *   <li>Translate input roles into access profiles via {@code RoleToAccessProfiles}.</li>
+ *   <li>For each target field, find matching event/field rows for the fixed event ID.</li>
+ *   <li>For each matching case type, derive CaseViewField visibility from {@code AuthorisationCaseField} CRUD.</li>
+ *   <li>Report decisions and fail the test if any field is not returned.</li>
+ * </ol>
+ */
+class DefinitionSpreadsheetHarnessTest {
+
+    private static final Logger LOG = LogManager.getLogger(DefinitionSpreadsheetHarnessTest.class);
```
paulridout: Instead of explicitly declaring the logger, use the Lombok `@Slf4j` tag on the class.

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,490 @@
...
+        assertThat("Blank -D" + TARGET_FIELDS_PROPERTY + " value",
+            fieldsCsv.isBlank(), is(false));
+
+        Set<String> roles = parseRoles(rolesCsv);
+        List<String> targetFields = parseFields(fieldsCsv);
+
+        List<FieldDecision> decisions = evaluateSpreadsheet(
+            Path.of(spreadsheetPath),
+            roles,
+            EVENT_ID,
```
paulridout: To make this generic the event id should be set from a parameter like the roles and target fields.  The EVENT_ID constant can then be removed.

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,490 @@
...
+                                                           List<String> fieldIds) throws Exception {
+        LOG.info("Evaluating spreadsheet: {}", spreadsheetPath);
+        LOG.info("Event ID: {}", eventId);
+        LOG.info("Roles: {}", String.join(", ", roles));
+        LOG.info("Target fields: {}", String.join(", ", fieldIds));
+        try (InputStream inputStream = Files.newInputStream(spreadsheetPath);
+             Workbook workbook = WorkbookFactory.create(inputStream)) {
+
+            List<Map<String, String>> eventToFields = readSheet(workbook, "CaseEventToFields");
+            List<Map<String, String>> authCaseFields = readSheet(workbook, "AuthorisationCaseField");
```
paulridout: Define constants for the CaseEventToFields and AuthorisationCaseField tab names, just like one has been defined for the RoleToAccessProfile tab name.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,490 @@
...
+        LOG.info("Initial header row index for {}: {}", sheetName, headerRow.getRowNum());
+        if (ROLES_TO_ACCESS_PROFILE_SHEET.equals(sheetName)
+            && !isRoleToAccessProfilesHeader(headersByIndex)) {
+            headerRow = findHeaderRow(sheet, headerRow.getRowNum() + 1, formatter, thisRowHeaders ->
+                isRoleToAccessProfilesHeader(thisRowHeaders));
+            if (headerRow != null) {
+                headersByIndex = readHeaders(headerRow, formatter);
+                LOG.info("Adjusted RoleToAccessProfiles header row to index {}", headerRow.getRowNum());
+            }
+        }
```
paulridout, line 384: This seems unnecessarily complex.  Rather than calling readHeaders() then findHeaderRow() if the first row is not the header row, it might as well just call findHeaderRow() starting at the first row.  This should allow `&& !isRoleToAccessProfilesHeader()` to be removed from the first if statement.

The same comment applies to the CaseEventToFields and AuthorisationCaseField checks below.

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,490 @@
...
+     */
+    private static FieldDecision evaluateField(String fieldId,
+                                               Set<String> accessProfiles,
+                                               String eventId,
+                                               List<Map<String, String>> eventToFields,
+                                               List<Map<String, String>> authCaseFields) {
+        LOG.info("Evaluating field: {} (eventId={})", fieldId, eventId);
+        long totalEventToFields = eventToFields.size();
+        List<Map<String, String>> eventIdMatches = eventToFields.stream()
+            .filter(row -> equalsIgnoreCase(row.get("caseeventid"), eventId))
```
paulridout: Would be best to declare constants for the column names caseeventid, casefieldid and casetypeid

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,490 @@
...
+
+        if (caseTypesForEventField.isEmpty()) {
+            LOG.info("No CaseEventToFields row found for event/field");
+            return FieldDecision.notReturned(fieldId, "No CaseEventToFields row found for event/field");
+        }
+
+        for (String caseTypeId : caseTypesForEventField) {
+            LOG.info("Checking case type: {}", caseTypeId);
+            List<Map<String, String>> caseTypeFieldRows = authCaseFields.stream()
+                .filter(row -> equalsIgnoreCase(rowValue(row, "casetypeid", "case type id", "case type"), caseTypeId))
```
paulridout: It looks like "case type id" and "case type" are alternative names for the "casetypeid" column.  I can't any columns named like this in a definition spreadsheet I've got.  If they aren't used then this code can be simplified to just look for the a column named "casetypeid".

The same comment applies to "casefieldid", "case field id" and "case field" below.

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,490 @@
...
+        boolean hasAccess = AccessControlService.hasAccessControlList(profileSet,
+            accessControlLists,
+            AccessControlService.CAN_READ);
+        LOG.info("ACL read access result: {}", hasAccess);
+        return hasAccess;
+    }
+
+    private static AccessControlList toAccessControlList(Map<String, String> row) {
+        AccessControlList accessControlList = new AccessControlList();
+        accessControlList.setAccessProfile(nullSafeTrim(rowValue(row, "accessprofile", "access profile")));
```
paulridout: Is "access profile" used as an alternative column name?  If not consider removing it.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,490 @@
...
+                isAuthorisationCaseFieldHeader(thisRowHeaders));
+            if (headerRow != null) {
+                headersByIndex = readHeaders(headerRow, formatter);
+                LOG.info("Adjusted AuthorisationCaseField header row to index {}", headerRow.getRowNum());
+            }
+        }
+        LOG.info("Using header row index for {}: {}", sheetName, headerRow.getRowNum());
+
+        List<Map<String, String>> rows = new ArrayList<>();
+        for (int i = headerRow.getRowNum() + 1; i <= sheet.getLastRowNum(); i++) {
```
paulridout, line 423: I think this will fail if headerRow is null, i.e. a tab has rows in it but no rows that can be identified as the header row.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,490 @@
...
+     * @param roles input role names (normalized)
+     * @param rolesToAccessProfiles sheet rows with role/profile mappings
+     */
+    private static Set<String> translateRolesToAccessProfiles(Set<String> roles,
+                                                              List<Map<String, String>> rolesToAccessProfiles) {
+        Set<String> accessProfiles = new HashSet<>();
+        LOG.info("Translating roles to access profiles from sheet: {}",
+            ROLES_TO_ACCESS_PROFILE_SHEET);
+        for (Map<String, String> row : rolesToAccessProfiles) {
+            String rawRoleName = rowValue(row, "rolename", "role name", "role");
```
paulridout: Are these alternative column names present in a definitions spreadsheet?  If not, consider removing.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,490 @@
...
+            }
+
+            boolean viewable = canAccessCaseViewFieldWithCriteria(caseTypeFieldRows, accessProfiles);
+
+            LOG.info("Derived CaseViewField visibility from CRUD: {}", viewable);
+            if (viewable) {
+                LOG.info("Viewable for accessProfiles {} in case type {}",
+                    String.join(", ", accessProfiles), caseTypeId);
+                return FieldDecision.returned(fieldId, caseTypeId);
+            }
```
paulridout: This if statement will cause the method to exit as soon as viewable access is confirmed for the specified field on a case type.  If there are multiple case types then access will not be confirmed for the others.  Is this what is desired?

A slightly wider question is would there ever be more than one case type in a definition file?  If not then it may be an idea to hold the extracted case type as a single object rather than a list.

src/main/resources/logback.xml
paulridout, line 1: This file is almost identical to the file of the same name that's included in the java-logging dependency (com.github.hmcts.java-logging).  Unsure why this has been added.

src/main/resources/reform-base.xml
paulridout, line 1: This file is almost identical to the file of the same name that's included in the java-logging dependency (com.github.hmcts.java-logging).  Unsure why this has been added.

src/test/resources/logback-test.xml
paulridout, line 1: Unsure why this has been added.  Is this needed?

src/main/resources/logback.xml
markdathornehmcts: Gone

src/main/resources/reform-base.xml
markdathornehmcts: Gone

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
markdathornehmcts: Done

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
markdathornehmcts: Done

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
markdathornehmcts: Done

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
markdathornehmcts: Done

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
markdathornehmcts: done

src/test/java/uk/gov/hmcts/ccd/definition/DefinitionSpreadsheetHarnessTest.java
markdathornehmcts: done

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
markdathornehmcts: donw

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
markdathornehmcts: done

src/test/resources/logback-test.xml
markdathornehmcts: Was needed at some stage otherwise it dodn't work but no longer. Gone

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
markdathornehmcts: Moved to standalone

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
markdathornehmcts: Refactored.

scripts/run-definition-spreadsheet-harness.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+# Runs the definition spreadsheet harness with the provided arguments.
+#
```
paulridout, line 3: Add a comment line showing the usage here (as is output on line 20).  Also mention that roles and target-fields can be comma separated lists.  Add the heading "Example:" before the example on line 4.

```
# Usage:
# run-definition-spreadsheet-harness.sh <definition-xlsx> <roles> <target-fields> <event-id>
#
# <roles> and <target-fields> can be comma separated lists.
#
# Example:
# ./scripts/run-definition-spreadsheet-harness.sh ...
```

scripts/run-definition-spreadsheet-harness.sh
```diff
@@ -0,0 +1,35 @@
...
+# ./scripts/run-definition-spreadsheet-harness.sh   /Users/<user>/Downloads/ccd-appeal-config-preview-pr3017.xlsx   caseworker-ia-admofficer   isFeePaymentEnabled,sponsorEmailAdminJ,sponsorMobileNumberAdminJ,sponsorAddress   editAppealAfterSubmit;
```
paulridout: Just use single spaces between parameters

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import uk.gov.hmcts.ccd.domain.model.casedataaccesscontrol.AccessProfile;
+import uk.gov.hmcts.ccd.domain.model.definition.AccessControlList;
+import uk.gov.hmcts.ccd.domain.service.common.AccessControlService;
+
+/**
+ * Harness-style test that inspects a definition spreadsheet and reports whether
```
paulridout: Reword this line as it's not a test any more.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+ * specific fields are readable for the supplied roles on a fixed event
```
paulridout: As the event is now a parameter, should this be "supplied event" now rather "fixed event"?

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+ * and are therefore returned if so, or not returned.
```
paulridout: This line reads a bit strangely.  I think the "or not returned" part needs rewriting to say why the fields aren't returned.  Might be better as a separate sentence.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+ * <p>Required system properties (parameters):</p>
+ * <ul>
+ *   <li>{@code -Ddefinition.file}: absolute or relative path to the definition XLSX.</li>
+ *   <li>{@code -Droles}: comma-separated list of roles (idam: prefix allowed).</li>
```
paulridout: Just to make it clearer, change the part in brackets to say "(the "idam:" prefix for a role is allowed)"

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+ *   <li>{@code -Dtarget.fields}: comma-separated list of case field IDs to check.</li>
+ *   <li>{@code -Devent.id}: page event id.</li>
+ * </ul>
+ * <p/>
+ * Example of env variables to run  from IntelliJ:
```
paulridout: The -D arguments are system properties rather than environment variables.  Might be better to reword this to "Example of command line arguments:"

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+        }
+        if (fieldIds == null || fieldIds.isEmpty()) {
+            throw new IllegalArgumentException("No valid target fields provided in -D" + TARGET_FIELDS_PROPERTY);
+        }
+        return evaluateSpreadsheet(spreadsheetPath, roles, eventId, fieldIds);
+    }
+
+    private static List<FieldDecision> runFromSystemProperties() throws Exception {
+        String spreadsheetPath = requireProperty(SPREADSHEET_PATH_PROPERTY,
+            "Missing -D" + SPREADSHEET_PATH_PROPERTY + "=<path-to-xlsx>");
```
paulridout: This line has has angle brackets around the example value whereas the others don't.  Either remove them from this line or add them to the others to be consistent.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+        return decisions.stream()
+            .filter(decision -> !decision.isReturned())
+            .collect(Collectors.toList());
+    }
+
+    private static String requireProperty(String key, String missingMessage) {
+        String value = System.getProperty(key);
+        if (value == null) {
+            throw new IllegalArgumentException(missingMessage);
+        }
```
paulridout: As all the missing messages follow the same pattern, the missingMessage parameter could be changed to something like missingPropertyValue and the `throw new IllegalArgumentException` line changed to `throw new IllegalArgumentException("Missing -D" + key + "=" + missingPropertyValue)`.  This would make the calls to the method neater, e.g. `requireProperty(ROLES_PROPERTY, "role1,role2");`

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+            "Missing -D" + EVENT_ID_PROPERTY + "=eventId");
+
+        Set<String> roles = parseRoles(rolesCsv);
+        List<String> targetFields = parseFields(fieldsCsv);
+        if (roles.isEmpty()) {
+            throw new IllegalArgumentException("No valid roles provided in -D" + ROLES_PROPERTY);
+        }
+        if (targetFields.isEmpty()) {
+            throw new IllegalArgumentException("No valid target fields provided in -D" + TARGET_FIELDS_PROPERTY);
+        }
```
paulridout, line 133: This logic is basically duplicated in the `run()` method which is called immediately after.  Remove the logic from one of these places.

If the logic is removed from `run()` then that method becomes just a pass through to `evaluateSpreadsheet()` and is effectively redundant.  In this situation it would be better to remove `run()` completely and replace it with a direct call to `evaluateSpreadsheet()`.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+            } else {
+                Row foundHeaderRow = findHeaderRow(sheet, headerRow.getRowNum() + 1, formatter,
+                    DefinitionSpreadsheetHarness::isRoleToAccessProfilesHeader);
+                if (foundHeaderRow != null) {
+                    headerRow = foundHeaderRow;
+                    headersByIndex = readHeaders(headerRow, formatter);
+                    log.info("Adjusted RoleToAccessProfiles header row to index {}", headerRow.getRowNum());
+                }
+            }
+        }
```
paulridout, line 384: I think this can be simplified to something like this:

```
DataFormatter formatter = new DataFormatter();
if (ROLES_TO_ACCESS_PROFILE_SHEET.equals(sheetName)) {
    Row headerRow = findHeaderRow(sheet, sheet.getFirstRowNum(), formatter, DefinitionSpreadsheetHarness::isRoleToAccessProfilesHeader);
    if (headerRow != null) {
        log.info("Using RoleToAccessProfiles header row at index {}", headerRow.getRowNum());
        ...
    } else {
        throw new IllegalArgumentException("Missing header row in sheet: " + sheetName);
    }
}
```

The same can be applied to the if statements for the other sheets.

The as findHeaderRow() will now always be using sheet.getFirstRowNum() then the method can be changed to remove the startRow parameter and just get it from Sheet object inside the method.

This change would negate the need for the if statements containing the `throw new IllegalArgumentException("Header row not found for sheet: " + sheetName);` statements later on in the method.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+        if ("CaseEventToFields".equals(sheetName)) {
```
paulridout: Replace string with CASE_EVENT_TO_FIELDS_SHEET constant

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+                Row foundHeaderRow = findHeaderRow(sheet, headerRow.getRowNum() + 1, formatter,
+                    DefinitionSpreadsheetHarness::isCaseEventToFieldsHeader);
+                if (foundHeaderRow != null) {
+                    headerRow = foundHeaderRow;
+                    headersByIndex = readHeaders(headerRow, formatter);
+                    log.info("Adjusted CaseEventToFields header row to index {}", headerRow.getRowNum());
+                }
+            }
+        }
+        if ("AuthorisationCaseField".equals(sheetName)) {
```
paulridout: Replace string with AUTHORISATION_CASE_FIELD_SHEET constant

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+            if (!header.isEmpty()) {
+                headersByIndex.put(cell.getColumnIndex(), header.toLowerCase(Locale.ROOT));
+            }
+        });
+        return headersByIndex;
+    }
+
+    private static boolean isRoleToAccessProfilesHeader(Map<Integer, String> headersByIndex) {
+        return headersByIndex.containsValue("rolename")
+            && headersByIndex.containsValue("accessprofiles");
```
paulridout: These strings can be constants.  Would also make the approach consistent with the other is...Header methods.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
```diff
@@ -0,0 +1,561 @@
...
+     * @param rolesToAccessProfiles sheet rows with role/profile mappings
+     */
+    private static Set<String> translateRolesToAccessProfiles(Set<String> roles,
+                                                              List<Map<String, String>> rolesToAccessProfiles) {
+        Set<String> accessProfiles = new HashSet<>(roles);
+        log.info("Translating roles to access profiles from sheet: {}",
+            ROLES_TO_ACCESS_PROFILE_SHEET);
+        log.info("Seeded direct access profiles from supplied roles: {}", String.join(", ", accessProfiles));
+        for (Map<String, String> row : rolesToAccessProfiles) {
+            String rawRoleName = rowValue(row, "rolename", "role name", "role");
```
paulridout: The alternative column names are still present.  The previous comment is just marked as done, so does that mean they are used?

src/test/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarnessTest.java
paulridout, line 1: Should this test be under support-tools/test rather than src/test?

src/test/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,171 @@
...
+    private static final List<String> TARGET_FIELDS = List.of(
+        "isFeePaymentEnabled",
+        "sponsorEmailAdminJ",
+        "sponsorMobileNumberAdminJ",
+        "sponsorAddress"
+    );
+
+    private static final Path KNOWN_SPREADSHEET =
+        Path.of(System.getProperty("user.dir"))
+            .resolve("ccd-appeal-config-preview-pr3017.xlsx")
```
paulridout: The spreadsheet used by the test needs to be included as a resource file.  A different mechanism will be needed to get it rather than the user.dir system property.

src/test/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test
+    void shouldFailWhenCaseEventToFieldsHeaderRowCannotBeFound(@TempDir Path tempDir) throws Exception {
+        Path invalidSpreadsheet = tempDir.resolve("invalid-definition.xlsx");
+        try (Workbook workbook = new XSSFWorkbook();
+             OutputStream outputStream = Files.newOutputStream(invalidSpreadsheet)) {
+            workbook.createSheet("CaseEventToFields").createRow(0).createCell(0).setCellValue("not-a-header");
+            workbook.createSheet("AuthorisationCaseField").createRow(0).createCell(0).setCellValue("not-a-header");
+            workbook.createSheet("RoleToAccessProfiles").createRow(0).createCell(0).setCellValue("not-a-header");
+            workbook.write(outputStream);
+        }
```
paulridout: The file this creates will be the same each time, so it could be included as a test resource file rather than creating it programatically.  Doing this means it also wouldn't show as an uncommitted new file in git.

src/test/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarnessTest.java
```diff
@@ -0,0 +1,171 @@
...
+            EVENT,
+            TARGET_FIELDS
+        ))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessage("Header row not found for sheet: CaseEventToFields");
+    }
+
+    @Test
+    void shouldReadHeadersWhenTheyAreOnFirstRow(@TempDir Path tempDir) throws Exception {
+        Path spreadsheet = tempDir.resolve("first-row-headers.xlsx");
```
paulridout: As with the invalid-definition.xlsx file, the contents of the file are the same each time so could be included as a test resource file.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
markdathornehmcts: Initial simplification broke. Simplified slightly and sticking with what works.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
markdathornehmcts: Gone now

src/test/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarnessTest.java
markdathornehmcts: No. leaving as is.

src/support-tools/java/uk/gov/hmcts/ccd/tools/DefinitionSpreadsheetHarness.java
markdathornehmcts: That would mean exposing the evaluateSpreadsheet method which preference is to remain private.
Removing the checks also broke tests. Done some tidying up but leaving as is.

paulridout (approved):
Changes look OK.

PR #2893, opened 2026-05-21
CCD-6129:

shahir-ali:
*### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/CCD-6129

### Change description ###
Fixes a PSQLException thrown when users with large numbers of specific case-level role assignments search for cases. The JDBC driver rejected queries where the IN (:p1, :p2, ...) clause exceeded PostgreSQL's 65,535 parameter limit.

Replaced the IN clause with = ANY(ARRAY(SELECT unnest(CAST(:references_param AS bigint[])))) in GrantTypeSqlQueryBuilder and ExcludedGrantTypeQueryBuilder, passing case references as a single bigint[] array parameter instead of individual JDBC parameters.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/ExcludedGrantTypeQueryBuilder.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/ExcludedGrantTypeQueryBuilder.java
@@ -4,9 +4,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
@@ -47,18 +45,20 @@ public class ExcludedGrantTypeQueryBuilder extends GrantTypeSqlQueryBuilder {
     @SuppressWarnings("java:S2789")
     private String getCaseReferences(Supplier<Stream<RoleAssignment>> streamSupplier,
                                      Map<String, Object> params) {
-        Set<String> caseReferences = streamSupplier.get()
+        Long[] caseReferences = streamSupplier.get()
             .filter(roleAssignment -> roleAssignment.getAttributes() != null)
             .map(roleAssignment -> roleAssignment.getAttributes().getCaseId())
             .filter(Objects::nonNull)
             .map(Optional::get)
             .filter(StringUtils::isNotBlank)
-            .collect(Collectors.toSet());
+            .filter(ref -> ref.matches("\\d+"))
+            .map(Long::parseLong)
+            .toArray(Long[]::new);
 
-        if (!caseReferences.isEmpty()) {
+        if (caseReferences.length > 0) {
             String paramName = "case_ids_excluded";
             params.put(paramName, caseReferences);
-            return String.format(QUERY, REFERENCE, paramName);
+            return String.format(QUERY_ANY, REFERENCE, paramName);
         }
         return EMPTY;
     }
--- a/src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/GrantTypeSqlQueryBuilder.java
+++ b/src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/GrantTypeSqlQueryBuilder.java
@@ -24,6 +24,10 @@ public abstract class GrantTypeSqlQueryBuilder extends GrantTypeQueryBuilder {
 
     public static final String QUERY = "%s in (:%s)";
 
+    // Used for reference lists to avoid PostgreSQL's 65,535 parameter limit
+    // = ANY() treats the entire array as a single parameter unlike IN which expands each element
+    public static final String QUERY_ANY = "%s = ANY(ARRAY(SELECT unnest(CAST(:%s AS bigint[]))))";
+
     public static final String EMPTY = "";
 
     public static final String SECURITY_CLASSIFICATION = "security_classification";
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/GrantTypeSqlQueryBuilder.java
```diff
@@ -123,11 +127,14 @@ private String addInQueryForReference(Map<String, Object> params,
                                           int index) {
         if (allRoleAssignmentsHaveCaseReference(searchRoleAssignments)) {
             String referencesParam = String.format(REFERENCES_PARAM, index, paramName);
-            params.put(referencesParam, searchRoleAssignments.stream()
+            Long[] caseReferences = searchRoleAssignments.stream()
                 .map(SearchRoleAssignment::getCaseReference)
-                .collect(Collectors.toList()));
+                .filter(ref -> ref != null && ref.matches("\\d+"))
```
aktaskaan, line 132: While .matches("\\d+") ensures the string is numeric, if an invalid digit string sneaks into the role assignments, Long.parseLong will throw a NumberFormatException and crash the entire search transaction (HTTP 500).
You can use **CaseReferenceValidator::validateUID.** It is strictly enforces a 16-digit length, it mathematically guarantees Long.parseLong() will never throw an exception.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/ExcludedGrantTypeQueryBuilder.java
```diff
@@ -47,18 +45,20 @@ public String createQuery(List<RoleAssignment> roleAssignments,
...
-        Set<String> caseReferences = streamSupplier.get()
+        Long[] caseReferences = streamSupplier.get()
             .filter(roleAssignment -> roleAssignment.getAttributes() != null)
             .map(roleAssignment -> roleAssignment.getAttributes().getCaseId())
             .filter(Objects::nonNull)
             .map(Optional::get)
             .filter(StringUtils::isNotBlank)
-            .collect(Collectors.toSet());
+            .filter(ref -> ref.matches("\\d+"))
+            .map(Long::parseLong)
```
aktaskaan, line 55: If we use a filter to silently drop invalid/corrupted case IDs, we create two different security outcomes:

- In Grant builders: Silently dropping a bad ID is safe. The user simply fails to get access to the malformed case (Fail-Secure).

- In the Excluded builder: Silently dropping a bad ID is a security risk. The AND NOT SQL clause is never generated, meaning the user retains access to a case they were explicitly forbidden from seeing (Fail-Open).

In **ExcludedGrantTypeQueryBuilder**, do not silently filter out invalid IDs. If an exclusion ID fails validateUID(), you must explicitly handle it. either by throwing an exception (failing the search entirely so the corrupt assignment can be fixed) or by logging a severe warning.

src/main/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/GrantTypeSqlQueryBuilder.java
```diff
@@ -24,6 +24,10 @@ public abstract class GrantTypeSqlQueryBuilder extends GrantTypeQueryBuilder {
 
     public static final String QUERY = "%s in (:%s)";
 
+    // Used for reference lists to avoid PostgreSQL's 65,535 parameter limit
+    // = ANY() treats the entire array as a single parameter unlike IN which expands each element
+    public static final String QUERY_ANY = "%s = ANY(ARRAY(SELECT unnest(CAST(:%s AS bigint[]))))";
```
aktaskaan, line 29: The new template is slightly redundant: = ANY(ARRAY(SELECT unnest(CAST(:%s AS bigint[]))))
This round-trips the parameter from an array → set → array for no functional effect.
Please check this out again to find a simpler way

src/test/java/uk/gov/hmcts/ccd/data/casedetails/search/builder/AccessControlGrantTypeQueryBuilderTest.java
```diff
@@ -233,16 +246,47 @@ void shouldReturnChallengedOrganisationalQueryWhenRoleAssignmentsGrantTypeExists
...
+
+    @Test
+    void shouldNotExceedPostgresMaxParametersWithLargeNumberOfRoleAssignments() {
+        // Regression test for CCD-6129
+        // When a user has a large number of specific case role assignments (e.g. from a bulk grant),
+        // the generated SQL query must not exceed PostgreSQL's hard limit of 65,535 parameters.
+        // Previously the IN clause expanded each case reference into an individual parameter,
+        // causing PSQLException: PreparedStatement can have at most 65,535 parameters.
+        List<RoleAssignment> roleAssignments = new ArrayList<>();
+        for (int i = 0; i < 70000; i++) {
```
aktaskaan, line 270: you can safely drop the 70k loop down to ~3 items. Since there is no database connected in the unit test, it can't reproduce the Postgres exception anyway. Testing a few items is enough to prove the = ANY() syntax is generated.
Note: If you switch to using validateUID(), make sure the mock data generates valid 16-digit case references instead of String.valueOf(i)

src/test/java/uk/gov/hmcts/ccd/data/casedetails/DefaultCaseDetailsRepositoryTest.java
```diff
@@ -861,4 +864,57 @@ private void assertCaseDetails(CaseDetails caseDetails, String id, String jurisd
...
+        CaseTypeDefinition caseTypeDefinition = loadCaseTypeDefinition("mappings/bookcase-definition.json");
+        caseTypeDefinition.setRoleToAccessProfiles(asList(roleToAccessProfileDefinition("[CREATOR]")));
+
+        stubFor(WireMock.get(urlMatching("/api/data/case-type/TestAddressBookCase"))
+            .willReturn(okJson(defaultObjectMapper.writeValueAsString(caseTypeDefinition))
+                .withStatus(200)));
+
+        // Create 70,000 role assignments to simulate a user with a large number of specific case assignments
+        List<RoleAssignmentResource> roleAssignmentRecords = new ArrayList<>();
+        for (int i = 0; i < 70000; i++) {
```
aktaskaan, line 887: Just keep an eye on CI memory, as serializing 70k objects in WireMock is heavy

aktaskaan (changes requested):

PR #2917, opened 2026-07-03
CCD-7841: combines 4262-4769, ccd 6395,  ccd 4311

patelila:

- [ ] the `keep-helm` label has been added, if the helm release should be persisted after a successful build

### Change description ###

```
[ ] Yes
[ ] No
```


aktaskaan (commented):
1. The delete happens in the wrong place

CCD-4262 suggestion 3, as well as the ticket comments, describe deleting the queue row as the “mark as ingested” step. This means it should happen after the document has been successfully written to Elasticsearch.

At the moment, all 17 Flux agent configs delete the row in the JDBC input, before the ES write. There is also no claim column or reaper mechanism.

This also creates a recovery problem. The existing re-indexing runbook (logstash_re_indexing_query.sql) resets marked_by_logstash, but nothing maintains that state after V20260702. So the existing recovery path no longer works.

2. Supplementary-data updates can be lost

I tested this against real Postgres and the actual write path. setSupplementaryData uses native SQL and bypasses Hibernate’s @Version. As a result, case_data.version does not change, so the external version sent to Elasticsearch remains unchanged and the subsequent write gets a 409.

This affects all jurisdictions.

This is also new with this release, since production currently does not use external versioning. One concrete example is HMCTSServiceId, which feeds global_search.

I see three possible approaches, but all of them require a Flux change:

* Use the queue ID as the Elasticsearch version (which would require changing it to bigserial first).
* Increment case_data.version when supplementary data is updated.
* Accept last-write-wins semantics.

3. Why was 8faaf6268 reverted?

I couldn’t find a reason for the revert. Also, the commit message for e67701dd3 says that its test was never actually run.

The current LogstashPipelineConfigurationTest has an assertion that explicitly prevents the queue-ID approach. Given that this is one of the possible solutions, I think we should relax that assertion before the discussion so that the option can be properly considered.

My latest commit also contains test cases covering this scenario, so the behaviour described in item 3 can be verified rather than relying only on the existing assertion.

PR #2950, opened 2026-08-05
POFCC-367 - allow AM roles for solicitor and pui-case-manager to be recognised as PROFESSIONAL role categories

jamescatto-cgi:

https://tools.hmcts.net/jira/browse/POFCC-367

### Change description ###

This just adds the new roles to the regex that is used to parse user IDAM roles and match them to categories,

Additionally, this moves the integration tests into a new source root to prevent shadowing of the unit tests that can occur when importing the project into Intellij. 

> I don't think these integration tests are run anywhere, is that correct?

Finally, this adds the spotless plugin to allow automatically fixing checkstyle violations:

```bash
./gradlew spotlessApply
```

```
[ ] Yes
[ X] No
```


anthonydummer (approved):
LGTM

PR #3009, opened 2026-09-09
CCD-8236 globalsearch next hearing changes

markdathornehmcts:

https://tools.hmcts.net/jira/browse/CCD-8236

### Change description ###

Add next hearing details to globalSearch

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/ccd-data-store-api/values.preview.template.yaml
+++ b/charts/ccd-data-store-api/values.preview.template.yaml
@@ -275,6 +275,11 @@ logstash:
                       rename => {"[data][caseManagementCategory]" => "[data_new][caseManagementCategory]" }
                     }
                 }
+                if [data][nextHearingDetails] {
+                    mutate {
+                      rename => {"[data][nextHearingDetails]" => "[data_new][nextHearingDetails]" }
+                    }
+                }
                 if [supplementary_data][HMCTSServiceId] {
                     mutate {
                       rename => {"[supplementary_data][HMCTSServiceId]" => "[supplementary_data_new][HMCTSServiceId]" }
@@ -303,6 +308,11 @@ logstash:
                 if [data_classification][caseManagementCategory] {
                     mutate {
                       rename => {"[data_classification][caseManagementCategory]" => "[data_classification_new][caseManagementCategory]" }
+                    }
+                }
+                if [data_classification][nextHearingDetails] {
+                    mutate {
+                      rename => {"[data_classification][nextHearingDetails]" => "[data_classification_new][nextHearingDetails]" }
                     }
                 }
                 mutate { remove_field =>[ "data" ,"supplementary_data", "data_classification", "last_state_modified_date", "type","last_modified", "created_date" ] }
--- a/src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/GlobalSearchQueryBuilder.java
+++ b/src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/GlobalSearchQueryBuilder.java
@@ -2,6 +2,7 @@ package uk.gov.hmcts.ccd.domain.service.search.global;
 
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.google.common.collect.Lists;
+import jakarta.inject.Named;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.lucene.search.join.ScoreMode;
@@ -19,7 +20,6 @@ import uk.gov.hmcts.ccd.domain.model.search.global.Party;
 import uk.gov.hmcts.ccd.domain.model.search.global.SearchCriteria;
 import uk.gov.hmcts.ccd.domain.model.search.global.SortCriteria;
 
-import jakarta.inject.Named;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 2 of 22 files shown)


src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/GlobalSearchQueryBuilder.java
```diff
@@ -111,6 +112,7 @@ public ArrayNode globalSearchSourceFields() {
             .add(CASE_MANAGEMENT_CATEGORY_NAME)
```
patelila, line 112: High

  - The response is still dependent on source-field availability and reindexing.

    The response now maps the value at ccd-data-store-api/src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/
    GlobalSearchResponseTransformer.java:112.

    The query requests the object at ccd-data-store-api/src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/
    GlobalSearchQueryBuilder.java:112.

    This will work only when:
      1. the global-search document contains data.nextHearingDetails;
      2. the Logstash clone preserved it;
      3. the Elasticsearch mapping accepts it;
      4. post-filtering allows the field for the user’s role.

    The implementation covers points 1–4 for new data, but no migration/backfill is included for existing data. Without that, the MC (Manage Case)
    migration can still lose dates for older cases.

  - Potential security regression must be explicitly protected.

    nextHearingDetails is included in the global-search source fields. The field classification is also copied through Logstash.

    This is correct only if child-level ACL filtering works for hearingDateTime. The existing tests assert that an unauthorised role
    receives null, but add a dedicated test covering:
      - authorised hearingDateTime;
      - unauthorised hearingDateTime;
      - authorised hearingID but unauthorised date, if mixed child ACLs are supported.

  - The issue requires returning a date, but the implementation also adds a new sort option.

    nextHearingDate sorting is introduced at ccd-data-store-api/src/main/java/uk/gov/hmcts/ccd/domain/model/search/global/
    GlobalSearchSortByCategory.java:19.

    Sorting may be useful for MC, but it expands the API surface beyond the stated requirement and introduces additional malformed/
    missing-date behaviour. Confirm that sorting is required by the consuming application before retaining it.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/GlobalSearchResponseTransformer.java
```diff
@@ -129,10 +132,11 @@ private String findValue(@NonNull final Map<String, JsonNode> jsonNodeMap,
 
         final Optional<JsonNode> optionalJsonNode = Optional.ofNullable(jsonNodeMap.get(parentKey));
         return optionalJsonNode.map(node -> {
-            if (node.isContainerNode()) {
-                return Optional.ofNullable(node.at(childPath)).map(JsonNode::asText).orElse(null);
+            if (childPath != null && node.isContainerNode()) {
```
patelila, line 135: Low

  - Malformed date can be returned as an invalid API value.

    Elasticsearch ignores malformed dates for indexing, but the transformer returns the raw _source value at ccd-data-store-api/src/
    main/java/uk/gov/hmcts/ccd/domain/service/search/global/GlobalSearchResponseTransformer.java:135.

    A malformed value could therefore:
      - sort as missing;
      - still be returned as "nextHearingDate": "not-a-date".

    Add an explicit policy and test. Prefer returning null for invalid values if the API contract says this is a date.

  - No contract/API test was identified.

    Add a contract assertion that /globalSearch returns:

    {
      "nextHearingDate": "2026-10-12T09:30:00.000"
    }

    Also verify that the field is omitted or null when unavailable, according to the API contract.

  - Date/time format and timezone behaviour are unspecified.

    The response returns the original string rather than a normalised value. Add tests for:
      - milliseconds;
      - no milliseconds;
      - Z timezone;
      - positive/negative offsets;
      - null;
      - malformed input.

charts/ccd-data-store-api/values.preview.template.yaml
```diff
@@ -305,6 +310,11 @@ logstash:
                       rename => {"[data_classification][caseManagementCategory]" => "[data_classification_new][caseManagementCategory]" }
                     }
                 }
+                if [data_classification][nextHearingDetails] {
```
patelila, line 313: - Environment configuration needs confirmation.

    The Logstash block was added to ccd-data-store-api/charts/ccd-data-store-api/values.preview.template.yaml:313. Equivalent
    configuration was not found in the AAT/default values files.

    Confirm whether those environments receive Logstash configuration from another deployment source. If not, the feature will work in
    preview but not elsewhere.

src/test/java/uk/gov/hmcts/ccd/ElasticsearchIT.java
```diff
@@ -187,22 +187,8 @@ public class ElasticsearchIT extends ElasticsearchBaseTest {
     private static final String REFERENCE_GLOBAL_SEARCH_04 = "4444111122223333";
     private static final String REFERENCE_GLOBAL_SEARCH_05 = "1999866820969999";
     private static final String REFERENCE_GLOBAL_SEARCH_06 = "1999866820970009";
+    private static final String REFERENCE_GLOBAL_SEARCH_07 = "5555666677778888";
```
patelila, line 190: Low 
 - Large unrelated test churn increases merge risk.

    ElasticsearchIT.java has substantial formatting/assertion-only changes. Separate that cleanup from the feature where possible.

src/test/java/uk/gov/hmcts/ccd/ElasticsearchIT.java
```diff
@@ -1997,6 +2268,7 @@ void shouldReturnCaseWithLookupAndRefDataPopulated() throws Exception {
                 () -> assertThat(result1.getCaseManagementCategoryId(), is("987")),
                 () -> assertThat(result1.getCaseManagementCategoryName(), is("Category label Order-02")),
                 () -> assertThat(result1.getCaseNameHmctsInternal(), is("Name Internal 01")),
+                () -> assertThat(result1.getNextHearingDate(), is("2026-10-12T09:30:00.000")),
```
patelila, line 2271: High
Missing functional coverage:
  This verifies the response using direct Elasticsearch test fixtures, but the issue requires proving the complete flow: case
  creation/update → Logstash projection → global-search index → /globalSearch.
  Please add/update the AAT functional test under src/aat/resources/features/F-1005 - GlobalSearch - Search cases:
  - include nextHearingDetails.hearingDateTime in the prerequisite case data;
  - call /globalSearch after the Logstash wait;
  - assert nextHearingDate in the expected response;
   - verify the authorised/unauthorised ACL behaviour.
The CCD-8236 diff currently contains no src/aat changes, so the end-to-end path is not covered.

src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/GlobalSearchQueryBuilder.java
markdathornehmcts: done

charts/ccd-data-store-api/values.preview.template.yaml
markdathornehmcts: done

src/main/java/uk/gov/hmcts/ccd/domain/service/search/global/GlobalSearchResponseTransformer.java
markdathornehmcts: done

src/test/java/uk/gov/hmcts/ccd/ElasticsearchIT.java
markdathornehmcts: done

src/test/java/uk/gov/hmcts/ccd/ElasticsearchIT.java
markdathornehmcts: done

